PR #15, opened 2026-08-15
docs: open Rust-crate positioning + move web portal to enterprise

bjmeetsfo:
…l to enterprise

The open repos (TemporalStore, MatrixCache, MatrixRaft) are Apache-2.0 Rust crates anyone can use — this makes that the headline and keeps the enterprise surface to just the hosted web console.

- README: new "Use it as a Rust library" section (Cargo git deps for all three
  crates + a real typed-client example: TemporalStoreClient -> open_table ->
  set/get) and a module map. Reframe the Cloud API section: the /v1 HTTP API, proxy, and every engine surface are open/self-hostable; only managed hosting and the web management console are enterprise. README validator tokens intact.
- crate README: add a "Using the crate" section (Cargo dep + client example + public module map); fix two C++-scrub leftovers ("so and Rust" / "the and Rust surfaces").
- remove website/matrixark-site/ (the web portal: HTML/CSS/JS) from the open repo — it is the enterprise console, a separate product. tools/ is unchanged.

open_source_ready: true; scrub clean; py_compile clean.

NOTE: the private->public sync will re-add website/ unless it is excluded at the source (bjmeetsfo) or in sync_oss_mirror.sh — follow-up.


matrixarkai (commented):
APPROVED

PR #26, opened 2026-08-16
feat(storage): true single-WAL-barrier durability with base-only recovery (TS_WAL_SINGLE_BARRIER)

bjmeetsfo:
## Summary

Completes the write-path durability barrier reduction to a **true single WAL barrier**, gated behind `TS_WAL_SINGLE_BARRIER` (default **OFF**; when off the write and load paths are byte-for-byte unchanged). Under the flag only the WAL takes a synchronous `fdatasync` per write (**1.00 fdatasync/write**); the data-page `fdatasync`, the served-index delta-log `fdatasync`, the band-manifest persist, and the per-write base-index rewrite are all deferred.

Correctness rests on the WAL + the durable dump checkpoint being a complete source of truth:

- **WAL-ordered config-log** makes config-driven eviction (the `feature_max_size` trim) reconstructable. Previously the trim lived only in the served-index delta and the shard config was neither logged nor persisted, so a pure WAL-tail replay re-executed feature appends with the default config and resurrected the evicted points. `set_config` now appends `{after_seq, config}` to a durable, WAL-sequence-ordered per-shard config-log and fsyncs it (config changes are rare admin ops, off the per-write hot path). Recovery restores the config effective at each record's WAL frontier before re-executing it, so replay re-derives the exact historical trim.
- **Expiry** resolves against the leader timestamp captured per WAL record and applies lazily; **compaction** is background and non-destructive to logical membership — both already WAL-re-derivable.
- **Watermark-bounded base-only recovery** makes deferring the data-page `fdatasync` safe. The durable dump/flush path fsyncs every data page *before* it materializes the base index and advances its watermark, so every page at/below the base watermark is on disk. On reload under the flag, recovery trusts **only** that durable checkpoint (the base snapshot, or the latest durable dump manifest if newer) — never the un-synced per-write base rewrite or the served-index delta, which may reference pages that were never fsync'd. It then replays the WAL tail from the durable watermark, re-deriving every post-dump page from its WAL command **exactly once** (the delta is not folded, so no double-apply of non-idempotent commands). A page written but never fsync'd is rebuilt, never left as a dangling reference.

## Mirror-specific data-safety fix

This mirror advanced the on-disk base index (its watermark **and** its page references) on **every write** via the default served-index-persist path. Combined with a deferred data-page `fdatasync`, that would leave the durable base pointing past pages that were never synced, so a crash would silently drop them (a naive port loses data here: a dump-based crash recovers a counter as `None` and drops post-dump keys). The fix keeps the invariant that base-only recovery never trusts a watermark or page reference beyond the durable page frontier: **under the flag the per-write base rewrite is deferred** (both in the ack path and in the post-replay reconstruct), so the base index is materialized only by the durable dump/flush path. The base watermark and its page references therefore never exceed the durable frontier. This is proven stable across a *recursive* crash-before-dump (a second recovery still restores every key).

## Verification (flag on)

Crash harness models real power loss by dropping every non-fsync'd byte; each phase runs in its own subprocess so the mode env never leaks.

- `single_barrier_data_page_loss_after_dump_rebuilds_from_wal` — dump at 150, write 151..300 (pages never fsync'd), truncate each slab to its recorded post-dump durable length, recover → `{"recovered":300,"missing":[],"mismatched":[],"ok":true}`.
- `single_barrier_non_idempotent_counter_applies_exactly_once` — a hash counter +1 ×200, dump at 100, lose the un-synced tail → `{"counter":200,"ok":true}` (not `None`, not doubled).
- `single_barrier_full_page_loss_no_dump_rebuilds_from_wal` — no dump, wipe all non-WAL state → base-only replay from 0 rebuilds all 300 keys.
- `single_barrier_evict_then_crash_before_dump_does_not_resurrect` — config-log re-derives the trim, no resurrection.
- `strace`: **1.00 fdatasync/write** under the flag (WAL only), vs 2.05 default.
- Full library suite unchanged with the flag off (default).

## Tradeoff (disclosed, not a data-loss bug)

Base-only recovery re-derives every post-dump page by replaying the WAL, so on reload the page store transiently holds re-written pages (GC reclaims the extra stale pages). A page store cannot be concurrently shared and re-written across two live engine instances under the flag; a correct restore/failover (old engine stopped, new engine loads) has a single active writer and is unaffected. Single-engine crash recovery loses nothing.

---

**Please do not merge without maintainer approval.** This changes durability-critical recovery behavior (behind a default-off flag) and should get a maintainer review before landing.

### DIFF
--- a/crates/temporalstore-rust/src/bin/wal_single_barrier_crash_harness.rs
+++ b/crates/temporalstore-rust/src/bin/wal_single_barrier_crash_harness.rs
@@ -27,7 +27,10 @@ use std::fs;
 use std::path::{Path, PathBuf};
 
 use serde::Serialize;
-use temporalstore_rust::{Command, CommandResponse, ExecuteRequest, TemporalEngine};
+use temporalstore_rust::{
+    Command, CommandResponse, Config, ExecuteRequest, FeaturePoint, SetConfigRequest,
+    TemporalEngine,
+};
 
 fn key_of(i: u64) -> String {
     format!("sbk-{i:010}")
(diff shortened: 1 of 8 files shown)


matrixarkai (commented):
APPROVED

PR #56, opened 2026-08-19
Storage: block and stream descriptors, and a reclaim guard for blocks in the WAL

bjmeetsfo:
Two pieces of the block-in-WAL work — the descriptors it needs, and the reclaim guard that has to exist before any block is addressed into the WAL — plus a vocabulary fix on one config knob.

## Storage descriptors

`storage_descriptor.rs` completes the on-disk type set alongside `wal_record` and `index_log_record`:

- **`BlockHeader`** — the per-block descriptor, including `block_in_wal`, which records that a block's bytes are still in the WAL rather than in a band.
- **`SlabInfo` / `SlabHeader` / `BlockFooter`** — the segmented-stream layer. A stream is a chain of slabs; each slab header names the whole chain, so a reopen can rebuild from the last slab alone rather than scanning all of them, and the fixed-size blocks inside carry footers that make a torn tail detectable.
- **`StreamInfo` / `WalInfo`** — the runtime views a follower uses to adopt a leader's stream state without re-reading the stream.

These are inert: nothing constructs them yet, so no on-disk format changes here.

## Reclaim guard

A block written into the WAL has **no second copy** until it is dumped into a band. Reclaiming the record that carries it destroys the block, and the failure does not surface at reclaim time — it surfaces later, as an unrelated read error.

`set_block_retention_floor(shard, sequence)` holds reclaim at the dump watermark. `gc_before_sequence` clamps its retain floor to it, and the report now carries `effective_retain_from_sequence` and `clamped_by_block_retention` so the clamp is observable rather than silent.

The floor is opt-in. Until a shard registers one, reclaim behaves exactly as before, so this is inert for callers that put no blocks in the WAL. It also composes with the existing rule that the highest-sequence record is never removed, rather than overriding it.

## Why nothing addresses a block by its WAL offset yet

Worth flagging, because it changes the shape of the remaining work.

This GC **rewrites** the WAL file. Removing a prefix shifts every surviving record down, so a byte offset is not a stable address across a reclaim — and a shifted offset still parses, as some *other* record. That failure is silent, which is the bad kind.

The retention floor does not fix this. It stops live records being deleted; it does not stop the survivors from moving.

Offset-stable reclaim needs the slab chain, where whole slabs leave the head and the remaining slabs keep their absolute positions. That is what `SlabInfo::truncated_offset` is for, and it is the next step. `reclaim_moves_the_surviving_records_so_byte_offsets_are_not_stable` pins the current behaviour so this cannot be forgotten.

## Config knob rename

The index-dump gap knob measures undumped WAL length, so it is now named for it: `index_dump_wal_gap_bytes`, read from `TS_INDEX_DUMP_WAL_GAP_BYTES`.

**This is not a configuration change.** The previous environment variable name is still honoured as a fallback, so a deployment that sets it is unaffected, and precedence when both are set is defined rather than incidental.

## Testing

27 passing in the touched areas, including every pre-existing WAL test. New coverage:

- reclaim is unconstrained when no floor is registered — the no-regression case
- reclaim will not pass the floor, and reports being held back
- advancing the floor releases the held-back records
- the floor composes with the tail-retention rule rather than overriding it
- reclaim shifts surviving offsets — the blocker above, pinned
- the previous env name still configures the knob; the current name wins when both are set; neither falls back to the default

### Full-suite attribution

Full suite on this branch: **747 passed / 10 failed**. Pristine `main` at the same base: **736 passed / 11 failed**. Diffing the failing-test *name* sets rather than the counts:

- **8 fail on both** — pre-existing.
- **3 fail only on `main`** and pass here.
- **2 fail only here** — `raft::tests::part2::http_raft_transport_sends_append_vote_and_snapshot_over_tcp` and `raft::tests::part3::distributed_raft_chunks_large_sequence_add_under_default_entry_limit`. Both pass 3/3 run isolated single-threaded, so they are parallel/port-contention flakes.

The raft set being unstable in *both* directions is the signature of a flaky cluster rather than a regression. Test-count delta 759 − 749 = exactly the 10 tests added here.

### DIFF
--- a/crates/temporalstore-rust/src/engine/persistence.rs
+++ b/crates/temporalstore-rust/src/engine/persistence.rs
@@ -357,14 +357,14 @@ impl TemporalEngine {
     }
 
     /// MANIFEST-PARITY FOLD threshold check: dump the index catalog when the undumped index-log
-    /// gap has crossed `index_dump_oplog_gap_bytes` (the index-meta dump gap
+    /// gap has crossed `index_dump_wal_gap_bytes` (the index-meta dump gap
     /// cadence). No-op with the `TS_INDEX_CATALOG_FOLD` gate off, so the background cycle is
     /// byte-identical when the fold is not enabled. Returns whether a dump fired.
     pub fn maybe_dump_index_catalog(&self, shard_id: ShardId) -> bool {
         if !crate::index_log::index_catalog_fold_enabled() {
             return false;
         }
-        let gap = crate::storage_config::index_dump_oplog_gap_bytes();
+        let gap = crate::storage_config::index_dump_wal_gap_bytes();
         let undumped = self.index_log_store.undumped_len_since_dump(shard_id);
         if !crate::index_log::should_dump_index_catalog(undumped, gap) {
             return false;
--- a/crates/temporalstore-rust/src/engine/storage_manager_cycle.rs
+++ b/crates/temporalstore-rust/src/engine/storage_manager_cycle.rs
@@ -948,7 +948,7 @@ impl TemporalEngine {
         );
 
         // MANIFEST-PARITY FOLD threshold dump (gate on only, never on a dry run): if the undumped
-        // index-log gap has crossed `index_dump_oplog_gap_bytes`, materialize the base index +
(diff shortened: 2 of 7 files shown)


superhaiou (commented):
lgtm

PR #102, opened 2026-08-21
test: remove a duplicated wal_record test that breaks the lib test build

bjmeetsfo:
## The problem

`wal_record.rs` defines `record_encoding_cost_per_byte_written` **twice**. The two copies are byte-for-byte identical, so this is a duplicated paste rather than two tests that drifted apart.

Rust rejects the redefinition, so **`cargo test --lib` does not compile on `main` at all** — not just for this module. Every lib test is unrunnable until one copy goes, which also means no branch based on `main` can verify itself against the lib suite.

## The fix

Remove the second copy. The surviving test is untouched.

Verification:

- `cargo test -p temporalstore-rust --lib wal_record` — 12 passed, 0 failed. Before this change the target does not compile.


bjmeetsfo:
Superseded by #101, which removed the same duplicate. Closing rather than leaving a conflicting duplicate open.

PR #132, opened 2026-08-22
Wait for a leader the group agrees on, instead of dictating one

bjmeetsfo:
`raft_secondary_replication_harness` fails on `main`: **0/5**. It is not flaky-by-nature — it was **3/3** at the elect-through-the-real-path commit, and **1/3** immediately after the failover work landed. Bisected to that merge.

Nothing is wrong with the failover work. It made leadership *real* — elections run between the processes, and a leader that loses contact with a majority for a few beats stands down. Two things in this test still assumed leadership was something it could pin.

## It insisted on a particular leader

The test asked a chosen node to lead and then waited for every other node to observe *exactly that node*. With real elections the group is free to choose someone else, and a leader can stand down on its own — so this asserts something the system does not promise.

It now nudges its preference, waits until every node names the **same** leader and that leader agrees it is one, and continues with whichever node the group settled on.

## It applied a membership change assuming a leader existed

A membership change needs a leader, and there can legitimately be no leader at a given instant: an election in flight, or a leader that just stood down. Every other admin call here already retries the conditions that clear on their own; this adds the one that appeared when leadership became real, bounded by a deadline and reporting every node's view if it expires.

## Result

Same machine, five runs each:

| | result |
|---|---|
| `main` | **0/5** |
| membership wait only | 3/5 |
| both changes | **5/5** |

The remaining two failures at the intermediate step were all `node N never observed node 1 as leader` — the test dictating a leader — which is what the first change addresses.

`cargo check --all-targets` exits 0. Test-only: nothing outside the harness binary changes.

### DIFF
--- a/crates/temporalstore-rust/src/bin/raft_secondary_replication_harness.rs
+++ b/crates/temporalstore-rust/src/bin/raft_secondary_replication_harness.rs
@@ -1155,6 +1155,18 @@ fn propose(node: &ProductionRaftNode, key: &str, value: &str) -> WriteSummary {
     }
 }
 
+/// Conditions that clear once the group settles, rather than a failed operation.
+///
+/// Leadership is decided by real elections now, so "no leader right now" is one of them: an
+/// election is in flight, or a leader that briefly lost contact with a majority stood down and
+/// another is about to be chosen.
+fn is_transient_admin_error(status: &Status) -> bool {
+    status.code == "raft_error"
+        && (status.message.contains("leader is not available")
+            || status.message.contains("not leader")
+            || status.message.contains("not enough live replicas"))
+}
+
 fn is_transient_proposal_error(status: &Status) -> bool {
     status.code == "raft_error"
         && (status.message.contains("not enough live replicas")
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Correction to the numbers in this description.

The "5/5 after" was real when measured, but it does not reproduce, and I should not have presented it as a settled result. Re-running the **same commit** later gave 0/5. The difference was the machine: another workload had taken the box to load average 21–28 on 16 cores, and this harness is timing-sensitive enough that its pass rate tracks that more than it tracks the code.

What still holds:
- The bisect. 3/3 at the elect-through-the-real-path commit, 1/3 immediately after the failover work — that comparison was made back to back and is what identified where the behaviour changed.
- The two changes here are right regardless. Leadership is decided by real elections now, so a test may not dictate who leads, and a membership change has to allow for there being no leader at that instant.

What does not hold: any claim that this makes the harness reliably green. It is currently red on `main` again, at a later point — a membership apply on a follower whose own view has no live leader while the real leader is healthy. Tracked in the pre-vote change (#134), which does not fix it either (0/5 both ways, interleaved on one machine).

The broader issue, which I think is the real one: this harness drives per-process admin APIs against each node's own view, and those views are only as fresh as the last AppendEntries that reached them. Each layer fixed exposes the next.

PR #158, opened 2026-08-23
Read the expiry window instead of examining every deadline

bjmeetsfo:
An expiry round acts on a bounded window — sixteen keys, say. To choose it, the round asked *"does this record still exist?"* of **every** key carrying a deadline, twice, then copied them all into a vector and sorted it. So the cost of a round tracked the whole set rather than the window, and it is paid on every storage cycle.

| deadlines | before | after |
|---:|---:|---:|
| 1,000 | 11,337 us | **359 us** |
| 5,000 | 55,773 us | **685 us** |
| 20,000 | **258,943 us** | **276 us** |

Flat rather than linear, and 938x at twenty thousand deadlines.

## The sort was not the problem

My first attempt kept the set unordered and chose the window with a bounded heap — no mass copying, no sort. It moved the first number by **a tenth**, from 11,337 to 10,134 us.

That is what identified the real cost: not the copying, but asking the question of every key. Keeping the deadlines in key order lets the window be read from where the cursor left off, so the question is asked only of the keys in the window. I am recording the failed attempt because the obvious read of the old code — "it sorts everything, that must be it" — is wrong, and someone will have it again.

## Two things the window has to get right

**A scan budget bounds the walk.** Looking for the few keys in one category when the set is almost entirely the other would still walk the set. The round stops early instead.

**The cursor advances past everything examined, not everything taken.** Otherwise a run of keys the round rejected is re-examined every round and the sweep never reaches what lies beyond it. There is a test for exactly that: 1,000 deadlines of which only the last ten are wanted, which must complete in several bounded rounds.

## Tests

- paging by cursor reaches every deadline exactly once, at window sizes from 1 to larger than the set
- a long run of rejected keys does not stall the sweep
- no limit still means every match after the cursor, with nothing to resume from

## Testing

Library suite: **1034 passed, 8 failed**. One name differs from the `main` comparison run — `manifest_fold_reload_reconstructs_catalog_with_band_manifest_deleted` — which has isolated 5/5 repeatedly today.

The deadline map changed from unordered to ordered. Every other use is `get`, `insert` or `remove`, which behave the same; one generic helper that named the unordered type is now written against the two operations it uses.

### DIFF
--- a/crates/temporalstore-rust/src/engine.rs
+++ b/crates/temporalstore-rust/src/engine.rs
@@ -1555,19 +1555,47 @@ fn apply_key_states(shard: &mut ShardState, key_states: &[serde_json::Value]) {
     }
 }
 
-fn apply_key_state_field<V: serde::de::DeserializeOwned>(
-    map: &mut HashMap<String, V>,
-    key: &str,
-    value: Option<&serde_json::Value>,
-) {
+/// Set or clear one key's entry, in whichever map holds it.
+///
+/// Written against the operations it actually uses -- insert and remove -- so that a map can be
+/// kept in key order where that matters without this having to care.
+/// The two things [`apply_key_state_field`] does to a map, so it does not have to name the map.
+trait KeyedState<V> {
+    fn insert_entry(&mut self, key: String, value: V);
+    fn remove_entry(&mut self, key: &str);
+}
+
+impl<V> KeyedState<V> for std::collections::HashMap<String, V> {
+    fn insert_entry(&mut self, key: String, value: V) {
+        self.insert(key, value);
+    }
+    fn remove_entry(&mut self, key: &str) {
+        self.remove(key);
+    }
+}
+
+impl<V> KeyedState<V> for std::collections::BTreeMap<String, V> {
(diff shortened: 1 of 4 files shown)


bjmeetsfo:
Rebased onto `main`, which had moved.

Resolving that conflict turned up a mistake of mine, and this branch now also corrects it.

**#153 shipped two dump measurements, not one.** The plan-only version — `what_delaying_a_dump_would_save_and_cost` — never calls `create_bucket_dump_manifest`, so no dump is ever taken, the dumped watermark never advances, and its numbers (100 / 96 / 76 / 1 dumps with a constant backlog) describe nothing. I wrote it, said in that PR that it was misleading and that the corrected version replaced it, and then shipped both: the script that added the corrected test **appended** it beside the old one instead of replacing it.

It is removed here. The corrected `what_delaying_a_dump_saves_and_costs`, which takes the dumps, stays.

I would rather this were its own PR, but the conflict was in that exact file and leaving a knowingly-wrong measurement on `main` while I opened a second one seemed worse. The diff is one deleted function.

PR #175, opened 2026-08-23
Stop rewriting the whole band manifest on every slab install

bjmeetsfo:
Installing a slab rewrote the **entire** band manifest: serialize every descriptor, write a fresh file, fsync it, rename, fsync the directory. So installing n slabs wrote the manifest n times, and each install cost time proportional to how many slabs already existed.

| slabs in store | per install, before | after |
|---:|---:|---:|
| 200 | 111.7 ms | **5.42 ms** |
| 800 | 270.7 ms | **5.38 ms** |

Twenty times faster at two hundred and fifty times at eight hundred — but the number that matters is that **5.42 and 5.38 are the same**. The cost has stopped tracking the size of the store; it is a shape change, not a faster constant.

## Why deferring the write is safe

The manifest is a cache. A reopen already reconciles it against the slabs actually on disk — adopting ones it has never heard of, and marking as purged the ones whose files are gone.

**I nearly broke that.** My first attempt added a "verify and rebuild" step to the load, on the theory that a stale manifest would be trusted. Two tests failed immediately: `reopen_marks_manifest_band_without_stream_file_as_purged` and `reopen_reconciles_manifest_missing_existing_stream_band`. The reconciliation was already there, already handled both directions, and preserved the purged marking that my wholesale rebuild threw away.

So the load is untouched. Only the write is deferred, and the existing reconciliation is what makes that sound. The change is smaller than the one I set out to make, because the safety net I was about to build already existed.

## The guard counts, it does not time

`installing_slabs_does_not_write_a_manifest_each_time` asserts on the **number of manifest writes**, via a new `band_manifest_writes` stat. A duration would assert the right thing on an idle machine and something else on a busy one; one write per install is the shape itself, so that is what is checked — along with a reopen still seeing every slab.

## Testing

`block_store` tests: **47 passed, 0 failed.**

The full library suite is harder to report honestly right now. One run gave 1032 passed / 26 failed, but it took 979s against a typical 250s with the machine at load 19–20, and the failures were the usual load-sensitive block: sixteen proxy tests plus a raft transport test. Three representatives — including the one name that is genuinely mine, `appending_does_not_get_slower_as_the_log_grows` — isolate **5/5**. A confirming run is queued behind another workload.

Worth noting separately: that WAL guard is mine, from #150, and its *timing* assertion is what flaked. Its count assertion held. Same lesson as above, which is why this PR's guard counts.

### DIFF
--- a/crates/temporalstore-rust/src/block_store.rs
+++ b/crates/temporalstore-rust/src/block_store.rs
@@ -127,6 +127,13 @@ pub struct BlockStoreStats {
     pub compressed_records_read: u64,
     #[serde(default)]
     pub compression_bytes_saved: u64,
+    /// Times the whole band manifest was written out.
+    ///
+    /// Writing it costs the whole manifest, so one write per slab install made installing n slabs
+    /// cost n manifests -- and each install cost time proportional to how many slabs already
+    /// existed. Counted rather than timed, because a count says the same thing on a busy machine.
+    #[serde(default)]
+    pub band_manifest_writes: u64,
     /// Slabs fetched on-demand from a shared-storage read-through source (parity
     /// lazy recovery). Each shared slab is fetched at most once, only when a read
     /// misses it locally; a nonzero count proves recovery did not install every slab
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
The confirming run finished: **1051 passed, 7 failed**, with only `manifest_fold_reload_reconstructs_catalog_with_band_manifest_deleted` differing from the `main` comparison set — the same long-standing flake, which has isolated 5/5 repeatedly today.

That is the cleanest suite result of this series, and it settles the earlier 1032/26: that run took 979s at load 19–20, this one 960s at load 9–12, and the sixteen proxy failures are gone without any code change between them.

PR #183, opened 2026-08-23
Let a shard log be more than one file, so reclaim can unlink instead of copy

bjmeetsfo:
Reclaim keeps a suffix of the log by copying it into a new file, so its cost tracks what it
**keeps**, not what it frees: dropping a prefix of a large log rewrites nearly all of it. One
measured pass copied 19.6 MB to free 3.8 MB. The cost guard added earlier declines passes like
that, which bounds the damage without changing the shape that causes it.

A log in more than one file fixes the shape. A piece whose every record is below the retain floor
is unlinked whole, and nothing is copied:

| log | reclaim | copied | unlinked |
| --- | --- | --- | --- |
| one piece | 38,656 us | 25,048 B | nothing |
| 8 KiB pieces | **14,517 us** | **6,944 B** | 55 pieces holding 463,829 B |

Same 2,000 records, same floor. The freed space stopped being paid for in copying.

## Off by default

`TS_WAL_SEGMENT_BYTES` of zero never rolls, and then a shard's log is one file and every byte of
this is inert. Existing logs are unaffected: a single file is just a log with one piece.

## No new bookkeeping

Each piece already records, in its first bytes, the log id its contents start at — the base header
that has been there since addressing moved off raw file offsets. So a log id already says which
piece holds it and where inside it, with nothing extra to persist and nothing to keep consistent.

Sealed pieces are named `shard-{id}.wal.{start_log_id:020}.jsonl`, which sorts into log order;
`shard-{id}.wal.jsonl` is the piece being written. Every position in and out of the readers is a
log id rather than an offset into a particular file — the addressing the log already uses for the
blocks it carries, and the only one that means anything across pieces or survives reclaim.

## Three things the tests caught that I had wrong

**The end of the log is not the end of the file being written.** That file is empty right after a
roll, so reading its tail said the log was empty and the next record reused sequence 1. It is the
end of the newest piece that holds a record. Only that piece can have a torn tail — the sealed ones
were whole when they were renamed — so recovery still repairs exactly one file.

**A record's address has to be taken before the roll it triggers.** Taken after, the record that
filled a piece was addressed as though it were in the piece that comes next, and read back as
garbage.

**Sealing is a rename and creating the next piece is a separate step, so a crash lands between
them:** sealed pieces, and nothing to append to. An absent piece reads as starting at log id zero,
which would hand out addresses the sealed pieces already own and point old readers at new records.
The next piece now derives its start from what is on disk, so it is the same answer whether or not
the crash happened.

## Tests

`a_rolled_log_reads_whole_and_keeps_its_addresses`, `a_rolled_log_continues_after_a_reopen`,
`a_crash_between_sealing_and_starting_the_next_piece_does_not_reuse_addresses`,
`reclaim_drops_whole_pieces_without_copying_them`, and
`what_reclaim_costs_in_one_piece_against_many` (the table above).

53 WAL tests pass; `cargo check --all-targets` clean.

The threshold override is per thread rather than per process, so a test that rolls does not make
every test beside it roll too.

### DIFF
--- a/crates/temporalstore-rust/src/wal.rs
+++ b/crates/temporalstore-rust/src/wal.rs
@@ -395,6 +395,13 @@ pub struct WriteAheadLogGcReport {
     /// The pass was declined because the copy it required bought too little space. The records
     /// are untouched and a later pass, once the prefix has grown, will take them.
     pub skipped_not_worth_rewrite: bool,
+    /// Whole pieces of the log that went without being copied, because everything in them was
+    /// below the floor.
+    #[serde(default)]
+    pub dropped_segments: usize,
+    /// What those pieces held.
+    #[serde(default)]
+    pub dropped_segment_bytes: u64,
 }
 
 #[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
@@ -571,6 +578,7 @@ impl LocalWriteAheadLogStore {
             let mut inner = self.inner.lock().expect("write-ahead log lock poisoned");
             fs::create_dir_all(&inner.root)?;
             let _append_lock = WalAppendLock::acquire(&inner.root, shard_id)?;
+            ensure_active_wal_segment(&mut inner, shard_id)?;
             let (last_sequence, on_disk_len) =
                 resolve_last_sequence_for_append(&mut inner, shard_id)?;
             inner.last_sequence_by_shard.insert(shard_id, last_sequence);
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Full `--lib` suite on this branch: **1058 passed, 7 failed**, 970s. Same failure count as `main`,
but not the same seven, so the trade is worth stating rather than reporting the count and moving on.

Four that fail on `main` pass here:

- `raft::tests::part1::raft_rejects_electing_stale_replica_until_it_catches_up`
- `raft::tests::part1::request_vote_higher_term_resets_prior_vote_before_decision`
- `raft::tests::part4::wal_backed_apply_snapshot_fence_survives_snapshot_restart`
- `raft::tests::part4::wal_backed_installed_snapshot_survives_restart_without_old_log_entries`

One is new:

- `engine::tests::part2::manifest_fold_reload_reconstructs_catalog_with_band_manifest_deleted`

That one passes 3/3 in isolation (`--test-threads=1 --exact`). It sets `TS_INDEX_CATALOG_FOLD`
process-wide through `FoldEnvGuard`, so under the suite's 16-way parallelism a concurrent test
observes the flipped value — the same env-var race that moves the raft tests, which is why four of
them changed verdict in the other direction on the same run. Nothing here touches the band catalog
or the fold; the shared surface is the suite's own parallelism.

I am reporting this as *not attributable to this change* rather than as *passing*, because I have
not eliminated the race — I have only shown the test is green when it is the only thing running.

PR #197, opened 2026-08-24
meta: stop the hottest reads taking the exclusive lock to count themselves

bjmeetsfo:
## The two hottest reads took the exclusive lock, to count themselves

Resolving a shard's location and refreshing a table's topology are the requests every client and proxy makes most. Both opened like this:

```rust
let mut state = self.inner.write().expect("meta lock poisoned");
state.counters.topology_query_total += 1;
```

Everything after that line is reading. The exclusive lock was held for one reason: the counter lived inside the state it was protecting.

The cost is not the increment, it is the exclusion. Every topology refresh serialised against **every other topology refresh**, and against every registration, heartbeat, freeze and table change in the cluster — so a fleet of proxies polling their routes queued up behind each other one at a time, and behind the control plane, to add one to an integer none of them read.

## What this changes

The counters move out of `MetaState` and onto the meta itself as atomics, shared by clone like the metrics recorder already beside them. Both read paths take a **shared** lock. `Relaxed` ordering throughout — these are reported, never used to order anything.

Nothing else about the paths changes.

## Proving it, rather than asserting it

Two tests hold a shared lock and then require a reader to get through from another thread, failing rather than hanging if it blocks. I checked they actually catch the regression by putting the exclusive lock back on just those two lines:

```
test a_topology_read_does_not_need_the_exclusive_lock  ... FAILED
test resolving_a_shard_does_not_need_the_exclusive_lock ... FAILED
a topology read blocked
a shard lookup blocked
test result: FAILED. finished in 20.00s
```

Twenty seconds, both timing out. With the shared lock the whole suite finishes in 0.42s.

## The trade this could have been

Moving a counter out of a lock is how counters start drifting. A third test runs eight threads through fifty topology reads each and requires the total to land on exactly four hundred — atomicity, not approximately-atomicity.

A fourth covers the thing that would otherwise break quietly: counters live outside the state now, so installing a snapshot has to carry them across explicitly, or a peer taking over would report having served nothing.

## Tests

4 new. Verification:

- `cargo test -p temporalstore-rust --lib meta:: -- --test-threads=1` — **226 passed, 0 failed.**
- `cargo check -p temporalstore-rust --all-targets` — clean.

---

### These four are a stack, not four independent changes

I cut each branch from the previous one, so they nest: **#193 ⊂ #197 ⊂ #198 ⊂ #200**. Each PR's diff against `main` therefore contains its predecessors, and merging #200 alone takes all four.

They all touch the same read path and two of them touch the same function, so untangling them into independent branches would mean resolving conflicts between them for no reviewing benefit. Reviewing in order — #193, #197, #198, #200 — reads as four separate ideas, which is what they are.

| | what it removes from the topology read path |
|---|---|
| #193 | a `shard_count` factor from retention and placement planning |
| #197 | the exclusive lock from the two hottest reads |
| #198 | per-request re-aggregation of unchanged load data |
| #200 | per-shard re-parsing of unchanged locations, and two allocations per candidate |

### DIFF
--- a/crates/temporalstore-rust/src/meta.rs
+++ b/crates/temporalstore-rust/src/meta.rs
@@ -2,6 +2,7 @@
 // Copyright 2026 MatrixArkAI
 
 use std::collections::{BTreeMap, HashMap, VecDeque};
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::fs::{self, OpenOptions};
 use std::io::{self, BufRead, BufReader, Write};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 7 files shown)


bjmeetsfo:
Superseded — main already does this.

`get_table_topology` on main bumps an atomic counter and takes a **read** lock, with the same reasoning this PR gives:

```rust
self.counters.topology_query_total.fetch_add(1, Ordering::Relaxed);
// Resolving a topology only reads. It took the exclusive lock solely to
// count, which serialised every client and every proxy hottest call...
```

That landed in #161. This branch makes the same change in a different spelling (`MetaCounters::bump(...)` rather than `fetch_add`), so rebasing it onto main produces duplicate imports, a duplicate `counters` field, and a duplicate `counted_meta` test helper — 6 compile errors, all of the form "defined multiple times".

Closing rather than reworking it into a no-op diff. If `MetaCounters::bump` is wanted as the house spelling, that is a separate tidy-up across all the counters at once.

PR #229, opened 2026-08-24
Export the switch that silently parks the metaserver's background work

bjmeetsfo:
## The problem

Muting metadata change is the widest switch on the metaserver. Besides refusing
every mutation, it parks four background loops:

- the retention sweep
- freeze aging
- the server failure detector
- the proxy failure detector

So **while the mute is set, a dead server is never detected and never frozen**,
and the cluster keeps routing to it. Tombstones are never collected.

None of that was exported. Metrics keep flowing while muted — the exporter is
unaffected — and `temporalstore_meta_topology_version` simply stops advancing,
which is indistinguishable from a quiet cluster. An operator who muted during
maintenance and forgot had no signal, and no alert could be written.

A second problem found while fixing the first: **`MetaRaftCluster` had no read
path for the mute at all.** It could be set through that backend and never read
back — the write-only-state shape. The metric could not have been reported on a
raft-backed metaserver without adding one.

## The change

- `meta_change_muted()` on both backends, returning `Option<bool>` so
  `backend_call!` gets one shape.
- The raft backend answers from a quorum read, or `None` when no quorum can
  answer.
- `temporalstore_meta_change_muted` is exported as a gauge.

**The series is emitted only when the answer is known.** Reporting `0` from an
unreadable cluster would claim changes are flowing at exactly the moment nothing
can be known; an absent series can be caught with `absent()`, a false zero
cannot.

"Since when" needs no new state: setting the mute already records a timestamped
topology event, so the recorded history answers it.

## What this deliberately does not change

The mute parking the failure detectors may well be intended — that is plausibly
the point of a maintenance switch. This change makes the blast radius visible
and documents it at the call site. It does not alter the behaviour, which is a
judgement about operational intent rather than a defect, and is yours to make.

## Tests

Two, in the metaserver binary's tests:

- the switch is visible on the dashboard (gauge flips 0 → 1 through `/meta/mute`)
- a raft-backed metaserver can read its own mute back

Both fail if the gauge is emitted as a constant.

## Verification

Serial deliberately: these tests bind fixed ports and collide with themselves
under the default thread count.


bjmeetsfo:
**The resolution for the conflict between #229 and #269 is written and tested — branch `oss/the-mute-metric-and-the-backend-warnings`.**

Neither branch has been touched. That one is #229 with #269 merged into it.

Both PRs append a test to the same place in the metaserver binary's test module,
and nothing else in either diff collides. The diff3 base of the single hunk is
empty — neither side changed anything the other also changed — so the resolution
is to keep both tests, and I checked that emptiness rather than assuming it. The
two bind different ports (18131 and 18211), so they do not contend.

`cargo test --bin matrixark_rust_metaserver` is 29 green, including both of the
tests in question. Note these live in the binary, so `--lib` does not run them.

PR #240, opened 2026-08-24
Retire the ContextEmbedding record

bjmeetsfo:
Retires the separate embedding record entirely, before the release ships it to anyone. Three commits, reviewable in order; net **-797 lines**.

### 1. Batch summary-vector reads (`ContextQuerySummaryVectors`)

Retrieval scored `node_l1` through the rows because there was no batched way to ask for summary vectors. The new command returns exactly the scoring input -- the newest summary vector at a level per node -- and the L1 half of summary scoring reads it, chunked at the same cap as before.

### 2. No writer emits rows

Live ingest keeps only the inline assignments (vectors ride on the summaries, the event, the node); the drainer and the backfill tool write the node-addressed command alone. The resource/skill ingest verifier proves embeddings landed by fetching the owner nodes and requiring vectors on them -- the thing retrieval actually depends on.

### 3. The record is gone

Struct, wire codec, both commands, the shard map, the `ctx:embedding` keyspace, the catalog entry, every classification entry. Former readers:

| Reader | Now |
|---|---|
| tree traversal | `node.vector`, nothing else |
| retrieval summary scoring | node vectors + L1 summary vectors; a vectorless node goes to the hybrid lexical pass, like any un-embedded node |
| backfill tool | detects from the node record, writes node-addressed |
| persistence workflow bin | counts embeddings on their owners |
| `event_text` rows | had **no reader at all** -- write-only cost |

### Compatibility

A pre-retirement shard index decodes cleanly (its `context_embedding` entries fall through to the ignore arm); orphaned rows become unreachable bytes reclaimed by normal compaction; pre-fold nodes count as un-embedded until the backfill re-embeds them. Acceptable exactly because this lands before the release -- no deployed store to migrate.

### Verification

- full lib suite, serial: **1185 passed / 2 failed** -- the known pre-existing pair, unchanged
- workflow, engine, wal slices green along the way; lib + all bins compile
- the python tools never serialized these commands (checked), so no cross-surface breakage


bjmeetsfo:
Pushed a fourth commit: MATRIXARK_CONTEXT_SECONDARY_INDEX (default OFF) skips ctxidx building at both ingest paths, plus the query-back that would have reported the deliberate skip as a failure. Retrieval never read these refs -- they were write-only cost whose only reader was the verifier checking its own writes. Refs are still reported so callers see what would have been indexed. The two tests that exercise the index surface itself run with the gate on; the redesign is tracked in its own issue.

bjmeetsfo:
Two more commits landed on this branch while chasing its CI:

**Revive the conformance corpus** -- it had been failing at deserialization long enough that everything behind the first error was archaeology. Un-burying it surfaced two real engine defects, both fixed here: the event rekey had missed **ContextWriteExtractedEvent** (every extracted event wrote successfully and was invisible to time-ranged queries) and the **recovery validator** (it compared page timestamps against the now id-keyed primary map -- this was the long-standing "known pre-existing" full-suite failure, , which now passes). The corpus drift itself is brought current rather than deleted; all four corpus tests pass serially and via the client transport.

The other long-standing failure of the "known pair" () is a wait_until timeout -- the proven flake under load, unrelated.

PR #262, opened 2026-08-24
Give a shard a read and write rate limit

bjmeetsfo:
`readiness.rs:396` reports `rate_limit_policy_ready = true`. **Nothing implemented it.** The only
rate limiting anywhere bounds how many shards a *rebalance* moves per window, which is a different
thing — so a single shard would absorb as much read or write load as a caller cared to send, and the
readiness report said that was fine.

A token bucket per direction, per shard. **A rate of zero means that direction is not limited, and
that is the default**, so a deployment that sets nothing behaves exactly as it does today.

Limits are settable on a running engine (`set_shard_quota`). Replacing one rebuilds the bucket,
because credit earned at one rate does not mean anything at another.

## How the bucket works

Tokens are not counted. What is stored is the point in time consumption has been charged up to, so
the tokens available are however much time separates that point from now, capped at the burst.
Nothing accumulates while idle beyond the cap, and no timer has to run to refill anything.

Configured by `TS_SHARD_{READ,WRITE}_QPS` and `TS_SHARD_{READ,WRITE}_BURST`, or per shard at
runtime. A burst left at zero defaults to one second of credit — a burst of literally zero would
refuse a caller sending at exactly the configured rate on any jitter.

## The part that is not about rate limiting

**Replicated applies and recovery replay are never charged.** Refusing either is not shedding load:

- a follower that rejects an entry its leader committed has *diverged* from it, and that surfaces
  much later as a shard disagreeing with its peers about what it holds;
- a replay that rejects a record already in the log cannot rebuild the shard.

Both paths already carry a thread-local marker (`RAFT_APPLYING`, `REPLAYING_WAL`) and both are
checked before the charge. `a_replicated_apply_is_never_refused_by_the_limit` exhausts *both*
directions through the caller-driven path, asserts the shard is genuinely over its limit, then
drives 200 committed entries through `execute_raft_apply` and asserts every one lands.

## Two bugs the tests caught

**The bucket started empty rather than full.** At the origin there is no elapsed time to draw on, so
the very first request was refused — a shard given a limit would have rejected everything until a
burst-window had passed. The charge point now starts one full burst *behind* the origin, which needs
a signed timeline: an unsigned zero cannot represent a new bucket having credit.

**The first wiring put a real cost on the hot path.** It read four environment variables and took a
**write** lock on every command — including the ~100% of commands on shards with no limit at all,
which would have serialised the engine on a feature almost nobody turns on. The default now resolves
once per process, and an unlimited shard settles under a read lock.

## One deliberate choice

The charge happens *before* the read-only fast path. A read served without taking the shard lock
still costs the shard, and a limit that the cheapest reads slip past is not a limit.

17 tests pass (5 on the bucket itself, 5 through the engine, plus the existing proxy quota tests);
`cargo check --all-targets` clean.

### DIFF
--- a/crates/temporalstore-rust/src/engine.rs
+++ b/crates/temporalstore-rust/src/engine.rs
@@ -33,6 +33,7 @@ mod recovery_sweep_compact;
 mod persistence;
 mod bucket_dump_io;
 mod command_validation;
+pub mod quota;
 pub(crate) mod eviction_sampler;
 // Single source of truth for write-command classification (shared with the data_node layer,
 // which previously kept a drifted subset that mis-classified context/control-state writes
@@ -106,6 +107,9 @@ pub struct TemporalEngine {
     configs: Arc<RwLock<HashMap<ShardId, Config>>>,
     infos: Arc<RwLock<HashMap<ShardId, ShardInfo>>>,
     admissions: Arc<RwLock<HashMap<AdmissionScope, AdmissionState>>>,
+    /// Per-shard read and write rate limits. Empty and inert unless something sets a limit, or the
+    /// environment carries a default.
+    quotas: Arc<RwLock<quota::QuotaTable>>,
     /// Diagnostics: number of per-execute `promote_model_maps_to_bucket_index_authority` full
     /// O(store) reconcile scans this engine has run at the hot-path call site. Without
     /// TS_PHASE1_FLAT this fires once per command (O(writes)); with the gate on the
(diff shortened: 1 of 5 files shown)


bjmeetsfo:
Local engine suite on this branch: **224 passed, 0 failed**, 2 ignored (1251s), and
`cargo check --all-targets` clean.

That matters more than the 17 quota tests, because the charge sits on the path *every* command takes
— a mistake there would show up as unrelated engine tests failing, not as a quota test failing.

PR #269, opened 2026-08-24
Say what the raft backend does not do, instead of implying it does

bjmeetsfo:
> Independent of the other open work — `main` base, one file.

## Six startup warnings told an operator the raft backend had work covered

It does not. Each names a setting the raft backend ignores, and each explained
that away.

### Two claimed placement

```
TS_META_AUTO_REBALANCE ignored: raft backend manages placement itself
TS_META_SHARD_DIVERGENCE_CHECK ignored: raft backend manages placement itself
```

Audited three ways: the raft timer loop does **peer failure detection only**;
there is **no rebalance or divergence machinery in the raft module at all**; and
`start_auto_rebalance_loop` / `start_shard_divergence_loop` are called from
exactly two places, both `MetaBackend::Single` arms. `add_node` / `remove_node`
change raft **membership**, not shard placement.

### Three claimed ownership

`TS_META_FREEZE_AGING`, `TS_META_RETENTION_GC` and `TS_META_PROXY_CALIBRATION`
said the backend "owns its own meta state". True, and leaving out what the
operator needs: nothing ages frozen resources, collects tombstones or calibrates
proxy groups there. Tombstones accumulate for the life of the cluster.

### And one conflated two different failovers

```
TS_RAFT_AUTO_FAILOVER ignored: raft backend fails over itself
```

The raft backend does fail over **its own** leader. This setting drives failover
for **datanode raft shard groups**, through `meta/raft_failover.rs`. Different
things:

- `plan_raft_failover` exists only on `SingleNodeMeta`
- `MetaRaftCluster` never calls it
- the `failover_primary()` in the runtime's heartbeat tick acts on the
  metaserver's own cluster

So a frozen datanode leaves its shard groups without a trigger, and that module's
own doc says what follows:

> the surviving replicas are never told their leader is gone… `tick_election`
> never observes a dead leader, and **writes to that group stall indefinitely**

That one is worse than the other five: not a capability quietly absent, but writes
stalling while the operator is told it is handled.

## One is untouched because it is accurate

`TS_META_ADAPTIVE_FAILURE_DETECTOR ignored: raft backend runs its own timer loop`
— it genuinely does run failure detection there.

## What this does not do

It corrects what the metaserver *says*. Whether a raft-backed metaserver should
*drive* datanode failover, rather than only stop claiming it does, is a design
decision and a larger one than the wording.

## Test

A message change has no behaviour to test, so the test pins the **fact** the new
messages assert: the raft backend reports no background subsystem activity. A
future change that gives raft one of these capabilities has to update the text
too, or it fails.

Mutation-checked — making the raft backend report any subsystem activity fails
it. It also exercises the single-node backend, so an empty report there could not
make the check pass vacuously.

## Verification

- `cargo test --bin metaserver -- --test-threads=1` — 27/27

### DIFF
--- a/crates/temporalstore-rust/src/bin/metaserver.rs
+++ b/crates/temporalstore-rust/src/bin/metaserver.rs
@@ -131,7 +131,10 @@ fn main() {
                 ))
             }
             MetaBackend::Raft(_) => {
-                warn!("TS_META_AUTO_REBALANCE ignored: raft backend manages placement itself");
+                warn!(
+                    "TS_META_AUTO_REBALANCE ignored: shard rebalancing is not available on the \
+                     raft backend, and nothing else performs it there"
+                );
                 None
             }
         }
@@ -176,7 +179,10 @@ fn main() {
                 Some(start_shard_divergence_loop(meta.clone(), options, interval_ms))
             }
             MetaBackend::Raft(_) => {
-                warn!("TS_META_SHARD_DIVERGENCE_CHECK ignored: raft backend manages placement itself");
+                warn!(
+                    "TS_META_SHARD_DIVERGENCE_CHECK ignored: divergence checking is not available \
+                     on the raft backend, and nothing else performs it there"
+                );
                 None
             }
         }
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
**The resolution for the conflict between #229 and #269 is written and tested — branch `oss/the-mute-metric-and-the-backend-warnings`.**

Neither branch has been touched. That one is #229 with #269 merged into it.

Both PRs append a test to the same place in the metaserver binary's test module,
and nothing else in either diff collides. The diff3 base of the single hunk is
empty — neither side changed anything the other also changed — so the resolution
is to keep both tests, and I checked that emptiness rather than assuming it. The
two bind different ports (18131 and 18211), so they do not contend.

`cargo test --bin matrixark_rust_metaserver` is 29 green, including both of the
tests in question. Note these live in the binary, so `--lib` does not run them.

PR #270, opened 2026-08-24
Make silence mean "commit the buckets before truncating the index log"

bjmeetsfo:
Index-log records may only be discarded once the buckets they describe have been dumped.
Truncating first throws away the record of state that is not durable anywhere else — exactly what
is needed to rebuild those buckets after a crash.

The guard enforcing that order is a field on the cycle request, and **that request is parsed from a
request body** (`bin/server.rs`), so what an *absent* field decodes to is the behaviour every caller
gets who does not name it.

The field carried a bare `#[serde(default)]`. On a `bool` that decodes an absent field to
**`false`** — the unsafe order — no matter that the type's own `Default` says `true`:

```rust
#[serde(default)]                                     // -> false when omitted
#[serde(rename = "index_gc_commit_dirty_slots_before_truncation")]
pub index_gc_commit_dirty_buckets_before_truncation: bool,
```

```rust
StorageManagerCycleRequest::default()
    .index_gc_commit_dirty_buckets_before_truncation   // true
```

So a caller sending a cycle request without mentioning the guard was silently choosing to truncate
first. Now defaulted explicitly, so an omitted field decodes to what the type already said it
should be. Naming it `false` still works — this is about what silence means, not about removing the
choice.

The new test takes a well-formed body and removes **only** that field, so it tests what silence
means rather than whether every other field happens to be optional. It fails without the change.

## Validation

`cargo test --lib -- --test-threads=1`: engine **216 passed, 0 failed**; data_node **46 passed, 1
failed** — that one is long-standing on unmodified `main` and unrelated.

### DIFF
--- a/crates/temporalstore-rust/src/engine/reports.rs
+++ b/crates/temporalstore-rust/src/engine/reports.rs
@@ -2871,6 +2871,11 @@ pub struct StorageLifecycleRequest {
     pub warm_cache: bool,
 }
 
+/// Index-log truncation waits for dirty buckets by default; see the field this serves.
+fn commit_dirty_buckets_before_truncation_default() -> bool {
+    true
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct StorageManagerCycleRequest {
     pub shard_id: ShardId,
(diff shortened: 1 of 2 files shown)


superhaiou:
Best regards,
haiou jiang(Vincent)

PR #271, opened 2026-08-24
Report what proxy calibration could not do

bjmeetsfo:
> Independent of the other open work — `main` base.

## A proxy group running short of capacity was computed and thrown away

Proxy calibration was the one background subsystem reporting nothing. The other
five each have a recorder — `record_conviction`, `record_divergence`,
`record_freeze_aging`, `record_reassignment`, `record_retention` — and
calibration had none. No HTTP route serves its plan either, and the loop ends in
`let _ = meta.calibrate_proxy_groups(options);`.

Its attach and detach are not invisible: both go through `set_proxy_group`, which
records a mutation and a topology event, so they land in the change history.

**The shortfall does not.** `ProxyGroupShortfall` says of itself:

> Idle proxies eligible for this group at the time of planning. Zero here with a
> shortfall means the cluster has no spare capacity in that location, **which is
> an operator problem rather than a metaserver one.**

The type names the operator as the audience, and nothing was telling them.

## The change

`record_calibration`, in the shape the other five use, and three gauges beside the
retention ones it mirrors:

```
temporalstore_meta_calibration_shortfall_groups    groups left short
temporalstore_meta_calibration_shortfall_proxies   proxies they are missing
temporalstore_meta_calibration_capped              changes the per-round cap held back
```

Attach and detach counts go into the existing `reassigned_total` family rather
than a new one, since they are the same kind of fact.

Recorded before the plan is applied, so a round that fails partway still reports
what it set out to do.

## The formula, which the test corrected

My first version counted `wanted - attached`. That overstates every shortfall by
exactly the proxies the round is about to attach — `attached` is the count
*before* the round. A group wanting three with one idle proxy available would
have reported three missing when the answer is one attach and **two** missing.

The test asserted two, failed, and the metric was wrong rather than the test. A
capacity gauge that overstates is the kind that gets muted.

## Tests

- **a group short of proxies is reported** — wants three, one available, expects
  one group and two missing proxies
- **a group at its target reports no shortfall** — the guard, so a healthy round
  cannot look like a capacity problem

Mutation-checked both ways: removing the recorder fails the first, and restoring
the overstating formula fails it too.

## Verification

- `cargo test --bin metaserver -- --test-threads=1` — 26/26
- `cargo test --lib -- --test-threads=1` — 1197 passed, 1 failed

That one failure, `data_node::…jitter_backoff…`, fails 3/3 in isolation on
unmodified `main` and is untouched here.


bjmeetsfo:
**Follow-up commit: two corrections to this branch, found while auditing the same shape elsewhere.**

The attach and detach counts were recorded from the plan, before any
`set_proxy_group` had run. Calibration returns on the first change that is
refused and leaves the rest of the plan standing, so a round could report
attaching proxies it never touched — muting metadata change mid-round is
enough to do it. It now counts what the round applied. The shortfall and the
cap still come from the plan, since what a round wanted and what it held back
are facts about planning it, and the round is counted either way.

They also landed on `reassigned_total`, which renders as
`temporalstore_meta_shards_reassigned_total` and is documented as "Shard
ownership changes, by cause". A proxy joining a group is not a shard moving,
and folding it in inflates the number an operator watches for rebalance churn.
It now has its own family, `temporalstore_meta_proxy_attachments_total`,
labelled by kind.

Two tests, both failing before the change: a refused round reports zero
attachments and leaves the shard counter untouched, and a round that runs to
the end still counts all of them — so the fix cannot have simply stopped
counting.

The same defect, independently, in freeze aging is #302.

bjmeetsfo:
**The resolution for the conflict between #271 and #276 is written and tested — branch `oss/calibration-and-check-metrics-together`.**

Neither PR's branch has been touched. That branch is #276 with #271 merged into
it and #302 stacked on top, so all three are there composed rather than
something to work out at merge time.

Both PRs add new fields to the same metrics struct and new blocks to the same
render function, at the same two places. The diff3 base of both conflict hunks
is empty — neither side changed anything the other also changed, they only both
added — so the resolution is to keep both, and I checked that emptiness rather
than assuming it. A hunk with a non-empty base would mean one side rewrote text
the other also rewrote, and keeping both there would quietly leave the old
version standing next to the new one.

`meta::` is 296 green with all three composed: 295 with #271 and #276 together,
296 once #302 goes on top.

Nothing else in either PR conflicts — the rest of both diffs merge on their own.

PR #276, opened 2026-08-25
Report what the divergence check did not examine

bjmeetsfo:
> Independent of the other open work — `main` base, two files.

## A divergence check that examined nothing reported the same as one that found nothing

`check_shard_divergence` skips servers outright — it `continue`s past a server
that has never reported shard states, and past one inside its reboot grace. Those
servers' shards are never examined. Both lists are on the report:

```rust
pub skipped_in_reboot_grace: Vec<String>,
pub skipped_without_shard_reports: Vec<String>,
```

Neither reached the metrics. So `temporalstore_meta_shard_divergence_total 0`
meant either "nothing diverged" or "the servers that mattered were never looked
at", with no way to tell them apart. The `diverged` field's own doc says the
report keeps "the full picture" — it does, and the exported view did not.

That is the failure mode `missed_events` exists to prevent elsewhere in this
metaserver: a gap that reads as quiet.

## And freeze aging never reported its cap

Retention records the work its per-round cap held back (`retention_capped`).
Freeze aging has the identical `capped` field on its plan and recorded nothing,
so a round doing less than it wanted looked like a round with less to do.

## The change

Both follow the recorders already there:

```
temporalstore_meta_divergence_skipped{reason="no_shard_reports"}
temporalstore_meta_divergence_skipped{reason="reboot_grace"}
temporalstore_meta_freeze_aging_capped
```

The skip reasons are labels on one gauge rather than two metrics, matching the
labelled families already exported.

## Test

A server registered but never reporting shard states: the check skips it, finds
no divergence, and the export now says so — `no_shard_reports 1` alongside
`reason="reboot_grace" 0`, so the two causes stay distinguishable. Mutation-checked:
dropping the recording fails it.

## Verification

- `cargo test --bin metaserver -- --test-threads=1` — 26/26
- `cargo test --lib -- --test-threads=1` — 1207 passed, 3 failed

Two of those three are environmental and were checked rather than assumed:
`proxy::…topology_check_stays_off_the_request_path` and
`raft::…http_raft_transport…` each pass **6/6 and 3/3 in isolation**, and the
proxy one passes 6/6 on unmodified `main` too; its body references nothing this
change touches. The third, `data_node::…jitter_backoff…`, fails 3/3 in isolation
on unmodified `main` and is genuinely pre-existing.

### DIFF
--- a/crates/temporalstore-rust/src/meta/shard_check.rs
+++ b/crates/temporalstore-rust/src/meta/shard_check.rs
@@ -425,6 +425,45 @@ impl SingleNodeMeta {
 
 #[cfg(test)]
 mod tests {
+
+    #[test]
+    fn a_server_the_check_skipped_is_not_reported_as_consistent() {
+        // The check skips a server that has never reported shard states -- it
+        // `continue`s, so that server's shards are never examined. Without this
+        // reported, `diverged 0` means either "nothing diverged" or "the servers
+        // that mattered were never looked at", and nothing tells them apart.
+        let meta = SingleNodeMeta::default();
+        assert!(meta
+            .register_server(RegisterServerRequest {
+                numa_nodes: Vec::new(),
+                server_addr: "quiet".to_string(),
+                node_id: 1,
+                location: "rack-1".to_string(),
+                binary_version: "v1".to_string(),
+            })
+            .status
+            .ok);
+
+        let mut checker = ShardChecker::new(ShardCheckOptions::default());
+        let (report, _) = meta.check_shard_divergence(&mut checker);
+        assert!(report.diverged.is_empty());
+        assert_eq!(report.skipped_without_shard_reports.len(), 1, "{report:?}");
+
+        let exported = meta.subsystem_metrics().prometheus();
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
**The resolution for the conflict between #271 and #276 is written and tested — branch `oss/calibration-and-check-metrics-together`.**

Neither PR's branch has been touched. That branch is #276 with #271 merged into
it and #302 stacked on top, so all three are there composed rather than
something to work out at merge time.

Both PRs add new fields to the same metrics struct and new blocks to the same
render function, at the same two places. The diff3 base of both conflict hunks
is empty — neither side changed anything the other also changed, they only both
added — so the resolution is to keep both, and I checked that emptiness rather
than assuming it. A hunk with a non-empty base would mean one side rewrote text
the other also rewrote, and keeping both there would quietly leave the old
version standing next to the new one.

`meta::` is 296 green with all three composed: 295 with #271 and #276 together,
296 once #302 goes on top.

Nothing else in either PR conflicts — the rest of both diffs merge on their own.

PR #337, opened 2026-08-25
Serve a loading or refused store honestly, and turn upsert index-log delta records on by default

bjmeetsfo:
## What this fixes

A ~45K-record store (456MB) built through batch-committing ingest with upsert index-log
delta records on served `{"memories": [], "count": 0}` with clean 200s on every reload,
while all of its records sat durably on disk. The fold of its 97MB upsert-record log was
the suspect, and the emission gate shipped dark pending a scale reload-equality test.

Loading the preserved damaged store directly through the engine (every recovery mode ×
every artifact combination: base present/sidelined, config-log present/sidelined, default
and TS_WAL_LEGACY_RECOVERY=1, all 8 store prefixes) reconstructs the FULL served view every
time — including a 9.4s fold of the whole 97MB upsert log with no base at all. The fold was
never the defect. The emptiness came from the serving layers, three defects deep:

1. **Lane response desync (the core mechanism).** The proxy lane protocol answers requests
   strictly in order on one stdout. A request abandoned by its own client timeout still
   gets its response line later, so the next caller on that lane read the abandoned
   request's answer as its own — and every later reply shifted one back. On a store whose
   first open outlives the client timeout this poisons the lane on every restart: reads
   flap between full, wrong-scope, and empty (one scope's scan was observed answered with
   another scope's records). Small stores load inside the timeout, which is why the
   small-store proofs never caught it.
2. **Vacuous-empty swallows.** `_get_count`/`_get_index` swallowed any backend failure into
   `0`/`[]`, and the by-count/by-index record loads silently dropped every record a loading
   shard could not serve — turning "the backend cannot answer yet" into an authoritative
   empty store (and letting a write path derive its append position from the same lie).
3. **Discarded load refusals.** The proxy cached an engine whose `load_shard` result was
   ignored, so a genuinely refused load (corrupt WAL, replay hole) became a healthy-looking
   server whose every op fails fast with `shard_not_loaded` — which the swallows above
   turned into an empty store, permanently.

## The changes

- Requests carry a `client_request_id` the proxy echoes on every response line; the lane
  readers discard responses tagged for a different request (untagged responses from older
  binaries are accepted unchanged).
- `_get_count` / `_get_index` / the record loads raise on retryable backend failures and
  answer empty only for an absent key on a reachable backend. The retry classifiers learn
  the load-window statuses ("not loaded", "recovering").
- `open_engine` refuses to open a root whose shard load was refused, names the cause, and
  caches nothing, so the next request retries the load.
- Upsert index-log delta records default ON (`TS_INDEXLOG_UPSERT_DELTAS=0` is the escape
  hatch), carrying the scale reload-equality proof the gate asked for:
  `tests/upsert_reload_equality.rs` + a harness bin drive 3,000 batch-committed upsert
  batches (each batch emits one upsert record), a durably-logged config, threshold dumps
  mid-stream, and SIGKILL restarts across two generations, then assert the served view
  equals the acked view under BOTH recovery modes (default base + WAL tail replay, and the
  delta-fold path the flat-log format relies on at scale). An ignored `evidence_store_probe`
  test loads any on-disk store prefix through the normal lifecycle and reports what it
  serves — the tool used for the damaged-store matrix above.

## Verification

- `tests/upsert_reload_equality.rs`: green under both recovery modes (607s, release).
- Full `--lib` suite with the new default: 1233 passed; 3 shared-box flakes pass when rerun
  serially; 1 failure (`storage_manager_runtime_supports_stop_pause_resume_...`) also fails
  on the unmodified parent build, i.e. pre-existing and unrelated.
- `--bin matrixark_rust_proxy`: 36/36 including the new refusal test (which fails without
  the open_engine fix).
- Recovery integration suites (`delta_served_index`, `wal_single_barrier_recovery`,
  `delta_index_log_gc`, `delta_index_cost`, `list_recovery`): all green.
- Python: `tools/test_backend_unavailable_reads.py` 9/9 — the swallow/classifier tests fail
  4-of-7 on the pre-fix tree; the lane-correlation tests pin stale-response discard.
- Live gateway proof on the REAL damaged store (fresh copy, bases sidelined so the first
  open replays the full 57K-record WAL ≈ 35s, default 20s client timeouts, config-log
  present): before the fix, 149 consecutive `{"count": 0}` 200s and wrong-scope flapping,
  indefinitely; after the fix, one honest `backend_timeout` during the load window, then
  the exact count (50) on every read, indefinitely — zero vacuous empties.
- Live multi-generation reload proof (gateway level, 3 generations incl. SIGKILL): exact
  counts, upserts on by default.
- Fresh store built by this branch's binary (upsert records + batch commit via the gateway;
  the host killed the ingest gateway partway, so the durable set is what reached the engine),
  then cold-restarted twice on default timeouts: one honest backend error during each load
  window, then exact per-user counts, identical across both restarts (fully-ingested users
  serve 50/50; users whose tail sends were only 202-accepted into the killed gateway's
  in-memory pipeline serve exactly what the engine committed, stably) — no vacuous empties
  anywhere.

### DIFF
--- a/crates/temporalstore-rust/src/bin/upsert_reload_harness.rs
+++ b/crates/temporalstore-rust/src/bin/upsert_reload_harness.rs
@@ -1,0 +1,253 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 MatrixArkAI
+
+//! Scale reload-equality harness for the upsert index-log delta records
+//! (TS_INDEXLOG_UPSERT_DELTAS). Builds a store the way a batch-committing ingest does --
+//! thousands of hash batches (HashMultiSet + StringSet only, so every batch emits ONE
+//! upsert delta record), a durably-logged shard config (config-log present), threshold
+//! dumps part-way through (anchored base + durable pages), and an abrupt abort in place
+//! of a clean shutdown -- then reloads and asserts the SERVED view equals the view every
+//! acked write implies. Each phase runs in its own process so the caller controls the
+//! recovery-mode environment per phase.
+//!
+//! Modes:
+//!   build         --root R --batches N
+//!                   generation 1: write batches 0..N (each ack'd), threshold-dump at
+//!                   N/3 and 2N/3, then abort.
+//!   extend        --root R --batches N --extend M
+//!                   generation 2: load (recovers gen 1), VERIFY the gen-1 view, write
+//!                   batches N..N+M on top, then abort.
+//!   verify        --root R --batches T
+//!                   load (recovers everything) and assert the served view equals the
+//!                   final view of batches 0..T; print a JSON report; exit nonzero on
+//!                   any missing/mismatched/extra entry.
+
+use std::collections::BTreeMap;
+use std::path::{Path, PathBuf};
+
+use temporalstore_rust::{
+    BatchExecuteRequest, Command, CommandResponse, Config, ExecuteRequest, SetConfigRequest,
+    TemporalEngine,
+};
+
+const USERS: u64 = 120;
+const KEYS_PER_USER: u64 = 3;
+const FIELDS_PER_KEY: u64 = 40;
+const SETS_PER_BATCH: u64 = 8;
+const ENTRIES_PER_SET: u64 = 3;
+
+fn hash_key(user: u64, slot: u64) -> String {
+    format!("harness:mem:{user}:{slot}")
+}
+
+fn placement_key(user: u64) -> String {
+    format!("harness:placement:{user}")
+}
+
+fn field_name(index: u64) -> String {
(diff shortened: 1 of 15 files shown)


bjmeetsfo:
Ratchet note: the two remaining NEW-vs-baseline failures (test_listing_costs_one_read_and_reports_each_subjects_count, test_the_log_is_read_once_regardless_of_how_many_subjects) reproduce identically on unmodified main — they belong to the recently merged one-pass subject-counts work, not to this branch. The seven scoped-serving TypeErrors that were also flagged came from test doubles missing the new scan keywords and are fixed in this branch. cargo check / build / test is green.

PR #348, opened 2026-08-26
read the fields a request named, not the whole shard they live in

bjmeetsfo:
Records live in shard hashes of up to 1 024 fields, and each field's value is a bundle of records.
The locator and the type index resolve an id to exact `shard:field` addresses -- and then both the
delete path and the id-scoped payload fetch called `hgetall` on the shard, pulling back and
decoding every bundle in it to keep two.

Both now ask the engine for the fields they named. The full-shard snapshot cache is still consulted
first, so a caller that already has a snapshot pays nothing, and the field read deliberately does
NOT populate that cache: a handful of fields is not a shard snapshot, and storing it as one would
serve a partial answer to the next full read.

Measured on one 1 000-memory store:

    get(id)   60.1 ms  ->   8.7 ms
    delete     1 812 ms -> 1 359 ms
    update    12 851 ms -> 10 665 ms

Strict mem0 conformance 22/22, ten mem0 unit suites green, `cargo check --all-targets` clean.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -1188,7 +1188,7 @@ fn fetch_indexed_payload_values(
     let shards_touched = fields_by_shard.len() as u64;
     let mut values = Vec::with_capacity(locations.len());
     for (shard, fields) in fields_by_shard {
-        let snapshot = hgetall_map(engine, format!("{record_hash_key}:{shard}"))?;
+        let snapshot = fetch_shard_fields(engine, format!("{record_hash_key}:{shard}"), &fields)?;
         for field in fields {
             match snapshot.get(&field) {
                 Some(value) if !value.is_empty() => values.push(value.clone()),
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Withdrawing this: the change is not supported by measurement once the trade-off is measured properly, and the code already documents the opposite.

fetch_indexed_payload_values reads whole shard hashes deliberately -- the comment there records that a named-field read costs ~0.5 ms per field and NEVER warms, while a snapshot is read once and then kept current by the write runtimes. My first A/B favoured named-field reads on a cold snapshot (get(id) 60.1 -> 8.7 ms), but on a warm one both land at 3-4 ms, and the sweeps got worse: get_all 43.3 -> 48.6 ms and search 33.4 -> 35.7 ms.

Adding a field-count threshold did not rescue it either. A subject scatters ~12 fields per shard, so a sweep sits under any sensible threshold and still forfeits the snapshot that would have stayed warm for the next sweep. The existing design is right for this path; the waste I thought I saw is paid back by the snapshot on the very next read.

The delete path may still be worth narrowing on its own -- it names few fields AND mutates the shard immediately after, so the snapshot it would populate is invalidated anyway -- but that needs its own controlled measurement, not this one.

PR #423, opened 2026-08-28
Seven flag comments state the opposite default from the code

bjmeetsfo:
Seven flag doc comments state the opposite default from the code they sit on.

Each calls `env_flag_default_on`, which returns true unless the variable is explicitly `0/false/no/off` — so the behaviour described is what production runs. Every one of these comments says **"Default OFF"**, and most follow it with a reassuring *"byte-identical"* or *"exactly as before"*:

| flag | comment says | code does |
|---|---|---|
| `engine_concurrent_commit` | OFF | **ON** |
| `phase1_flat_enabled` | OFF | **ON** |
| `raft_apply_coalesce` | OFF | **ON** |
| `raft_snapshot_state_image_on` | OFF | **ON** |
| `raft_wal_coalesce_on` | OFF | **ON** |
| `raft_propose_serialize_on` | OFF | **ON** |
| `wal_fast_append_seq` | OFF | **ON** |

So a reader asking the question that actually matters — *is this path live?* — gets told the optimization is off and the legacy path is running, when the opposite is true. For the durability-adjacent ones (`raft_apply_coalesce`, `raft_wal_coalesce_on`, `wal_fast_append_seq`) the comment also implies the more conservative per-entry barrier behaviour is in force. It is not.

These read like defaults that were flipped after validation without the comment following.

## What changes

Only the default clause of each comment. What the flag does and why is already accurate and is untouched. No code changes — `cargo check --all-targets` clean, and the diff is 14 inserted / 7 deleted lines across three files, all inside `///` blocks.

## The test

A comment cannot fail, which is why this survived. The test reads the sources: for every `fn name() -> bool` whose body picks a default helper, the doc block above it must not assert the other default. It checks 10 documented flags today and fails with the offender named:

```
flag docs contradict their code (1 of 10 checked):
  engine.rs::phase1_flat_enabled documents "default off" but calls env_flag_default_on
```

That output is from a deliberately broken build — restoring one of the seven stale claims fails the test, so the guard is real rather than decorative.

## Why I was looking

While timing the write path I needed to know whether the phase-1 reconcile fast-skip was engaged, because with it off every batch pays a scan that "walks and CLONES every live model-map entry in the shard". The comment said default off. The code says on. Getting that backwards would have sent me optimizing a path that is already skipped — and on this codebase "is this optimization actually engaged?" has been answered wrongly before.

### DIFF
--- a/crates/temporalstore-rust/src/engine.rs
+++ b/crates/temporalstore-rust/src/engine.rs
@@ -2546,7 +2546,8 @@ pub(super) fn wal_single_barrier() -> bool {
 /// UNDER the `shards` lock (preserving WAL-order == apply-order), then RELEASES the lock and
 /// awaits the durable barrier (`commit_barrier`). This lets concurrent same-shard writers reach
 /// the group-commit queue while a peer's fdatasync is in flight, so #45's fsync coalescing
-/// actually engages (fewer fsyncs than writes; QPS scales with concurrency). Default OFF ->
+/// actually engages (fewer fsyncs than writes; QPS scales with concurrency). Default ON; set
+/// the variable to 0 for
 /// byte-identical to the legacy in-lock `append_with_sync` barrier. The ack is always returned
 /// strictly AFTER the covering barrier succeeds, so durability is never weakened.
 fn engine_concurrent_commit() -> bool {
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
Closing: main already has all of this, and it is verified there rather than assumed.

Both halves landed while this was open.

The comment corrections are on main, in every place this branch touched -- the WAL fast-append gate, the repeat-scan gate, the snapshot state-image gate, the fsync change-check gate, the propose-order gate and the WAL tail-scan gate. Main's wording is fuller in one case: it says outright that a stale default in a comment is worse than no comment, because it gets read as fact while deciding whether another switch can be turned on.

The guard is on main too: `flag_default_doc_tests::flag_docs_state_the_default_the_code_actually_has`, in engine.rs, which walks every boolean flag reader in the source and checks the documented default against the one the code actually has. Run against main just now, it passes.

Checked rather than assumed: merging main into this branch produced a second copy of that test module beside main's, which is what said the work was already there.

Nothing here is lost by closing.

PR #437, opened 2026-08-29
The page-ref key was stored three times per page; share one allocation

bjmeetsfo:
The page-ref key — `kind:object_key:component:slab:offset:length:page_id:generation` — is stored **three times per page**: as the `page_index` map key, again in `PageLookupRef.page_ref_key`, and again in `ComponentPageLookupRef.page_ref_key`. Three separate allocations of byte-identical strings.

Measured on a 500-memory store with 28,093 page entries, the in-memory index holds **15.9 MB of identity strings**, and the three copies of this one key are **3,208 KB each — 9.6 MB of that total**.

Making all three `Arc<str>` lets the lookups hold the map key's allocation instead of copying it.

## Measured

Counting distinct allocations by pointer identity, because byte counts cannot show sharing — three `Arc`s pointing at one allocation still report three lengths:

| | page-ref key holders | distinct allocations |
|---|---:|---:|
| after | 83,466 | **27,822** |

Exactly 3.000 holders per allocation, collapsed to one each. Proxy RSS over the same 500-memory ingest went **425.9 → 412.5 MB**, about 27 KB per memory, and it scales with the store.

## Why this shape

The wire format does not change: serde serializes `Arc<str>` exactly as a string (it needs the `rc` feature, enabled here). And a type change is compiler-enforced — every holder of the key had to be updated or the build fails, which is the opposite of the usual risk with a shared-state optimization where a missed site goes silently wrong. The compiler named nine sites; all nine are in this diff.

`WalOutcomeItem.page_ref_key` deliberately stays a plain `String`. It is a wire type that gets serialized and shipped, so sharing an allocation across that boundary buys nothing.

On reload, serde's `rc` support does not restore sharing — each key deserializes into its own allocation. That is fine here: `rebuild_object_page_lookup` clones the map key's `Arc` into both lookups, and `Arc::clone` shares, so the sharing comes back on the standard post-load path.

## What this does not address

The same measurement shows two further duplications, left for separate changes:

- `PageIndex` stores `object_key`, `model_id` and `component` — all three already encoded in the map key it is filed under (89 B per entry, 2,442 KB).
- `model_id` has **10 distinct values across 28,093 entries** and `object_key` has **556**, so both are candidates for interning rather than a `String` per page.

## Tests

`cargo check --all-targets` clean. 42 bucket-index tests, 22/22 strict mem0 conformance with 0 failures, 98 tests across 8 mem0 suites.

The full `--lib` run shows 7 failures. None are from this change: run individually with `--test-threads=1`, six pass on both this branch and pristine main, and the seventh (`storage_manager_runtime_supports_stop_pause_resume_jitter_backoff_and_phase_flags`) fails on pristine main too. That set includes `a_numeric_component_travels_as_a_number_and_returns_intact`, which exercises the component path this change touches — it passes on both.

### DIFF
--- a/crates/temporalstore-rust/Cargo.toml
+++ b/crates/temporalstore-rust/Cargo.toml
@@ -16,7 +16,7 @@ bytes = "1"
 rmp-serde = "1.3"
 hex = "0.4"
 prost = "0.13"
-serde = { version = "1", features = ["derive"] }
+serde = { version = "1", features = ["derive", "rc"] }
 serde_json = { version = "1", features = ["raw_value"] }
 base64 = "0.22"
 sha2 = "0.10"
--- a/crates/temporalstore-rust/src/engine.rs
+++ b/crates/temporalstore-rust/src/engine.rs
@@ -2086,7 +2086,7 @@ fn collect_command_index_items(
             items.push(crate::index_log::IndexItem {
                 kind: crate::index_log::IndexItemKind::Page,
                 routing_bucket,
-                page_ref_key: page_ref_key.clone(),
+                page_ref_key: page_ref_key.to_string(),
                 object_key: page.object_key.clone(),
                 model_id: page.model_id.clone(),
                 component: page.component.clone(),
@@ -2276,7 +2276,7 @@ fn fold_delta_page_items(
             });
         bucket.object_index.insert(item.object_id);
         bucket.page_index.insert(
-            item.page_ref_key.clone(),
+            Arc::<str>::from(item.page_ref_key.as_str()),
             PageIndex {
                 object_key: item.object_key.clone(),
                 model_id: item.model_id.clone(),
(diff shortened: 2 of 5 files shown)


bjmeetsfo:
Closing: shipped in a better form as #480.

This PR and #480 address the same duplication — the page-ref key held once per page in the index
map and again in each of the two lookup structures, three copies of the same string. #480 does it
by making the key an `Arc<str>` shared across all three, which is the smaller change and leaves the
wire format alone (`Arc<str>` encodes as the string it holds).

The measurement that settled it, at 600 memories:

| part | MB | B/item |
|---|---:|---:|
| `object_component_lookup` ref keys | 3.90 | 117 |
| `object_page_lookup` ref keys | 3.90 | 117 |
| `object_page_lookup` map keys | 3.40 | 102 |

The two ref-key rows are the duplication both PRs target. #480 measured **-11.8 MB, -3.4% proxy
RSS**, with add latency measured and unchanged.

Worth recording for anyone reading this later: a census that sums field lengths predicted 14.8% and
was wrong, because sharing removes an *allocation*, not a length. Dedup changes have to be measured
by resident memory, not by summing.

PR #440, opened 2026-08-29
Intern the page index identity strings instead of one allocation per page

bjmeetsfo:
`PageIndex` held `object_key`, `model_id` and `component` as owned `String`s — one set per page entry — and all three are already encoded in the map key the entry is filed under. Measured on a 500-memory store with 28,093 page entries: **2,443 KB, 89 B per entry**.

They repeat heavily, which is what decides interning over removal (the fields are read on hot comparison paths, so they have to stay readable):

| field | distinct values | entries | repeat |
|---|---:|---:|---:|
| `model_id` | **10** | 28,093 | 2,809x |
| `object_key` | **556** | 28,093 | 50x |
| `component` | 9,288 | 28,093 | 3x |

`model_id` is one of about ten string literals passed at the call sites; an `object_key` is shared by every page of that object.

## Measured

Distinct identity allocations, counted by pointer, because bytes cannot show interning any more than they show sharing — N entries pointing at one allocation still report N lengths:

| | distinct identity allocations |
|---|---:|
| before | 83,471 |
| after | **9,802** |

**−88%.** Proxy RSS over the same 500-memory ingest, bracketed start-to-end and interleaved:

| arm | RSS growth |
|---|---:|
| before | +402.2 MB |
| **after** | **+386.3 MB** |
| before (repeat) | +395.8 MB |

About 12.7 MB below the before mean, ~25 KB per memory, and it scales with the store.

One caution on that RSS figure: an earlier unbracketed reading of absolute RSS on a proxy carrying prior state suggested this change made things *worse* by 76 MB. It did not — measuring the delta from process start, interleaved, reverses it. Absolute RSS on a long-lived proxy is not a usable measure here.

## Shape

The table is global and bounded by the distinct identities a store actually holds — 9,802 at 500 memories against 28,093 entries. `Arc<str>` serializes exactly as a string, so the wire format is unchanged, and the type change is compiler-enforced: it named every site rather than letting one be silently missed.

`LivePageEntry`, `IndexItem` and the WAL outcome item keep plain `String`s. They are wire and transfer types, so interning across those boundaries buys nothing and would only add a lookup.

## Tests

`cargo check --all-targets` clean, 42 bucket-index tests, 98 tests across 8 mem0 suites.

**Strict conformance: 22/22, twice, on a cleaned machine.** Worth stating plainly, because two earlier runs of this branch scored 21/22 on `forget: get_all == 0` and `reset: tenant wiped`. Those two checks flake on this machine under load and with the disk near full — on the same cleaned runs, the branch this builds on scored 21/22 twice, failing exactly those two checks, while this branch passed both times. Across every run of both arms the flake rate is indistinguishable, so it is environmental rather than anything this change introduces.

### DIFF
--- a/crates/temporalstore-rust/src/engine.rs
+++ b/crates/temporalstore-rust/src/engine.rs
@@ -2080,16 +2080,16 @@ fn collect_command_index_items(
             continue;
         };
         for (page_ref_key, page) in &bucket.page_index {
-            if !keys.contains(page.object_key.as_str()) {
+            if !keys.contains(&*page.object_key) {
                 continue;
             }
             items.push(crate::index_log::IndexItem {
                 kind: crate::index_log::IndexItemKind::Page,
                 routing_bucket,
                 page_ref_key: page_ref_key.to_string(),
-                object_key: page.object_key.clone(),
-                model_id: page.model_id.clone(),
-                component: page.component.clone(),
+                object_key: page.object_key.to_string(),
+                model_id: page.model_id.to_string(),
+                component: page.component.as_deref().map(str::to_string),
                 object_id: page.object_id,
                 page_id: page.address.page_id.unwrap_or(0),
                 address: Some(page.address.clone()),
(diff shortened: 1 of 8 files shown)


bjmeetsfo:
Closing: the premise did not survive measurement.

This PR interns the page-index identity strings on the theory that they repeat across pages. A
field-level census of index memory at 600 memories says they mostly do not:

| field | MB | B/item |
|---|---:|---:|
| `object_key` text | 1.68 | 50 |
| `model_id` text | 0.13 | 4 |

`model_id` is 0.13 MB — there is nothing there to recover. `object_key`'s values are nearly all
distinct, so interning them adds a table and a lookup to hold very nearly the same bytes.

The duplication that *was* real in this area — the page-ref key held three times per page — is
shipped separately in #480: **-11.8 MB, -3.4% proxy RSS**.

For the record, the whole index is 52.5 MB at 600 memories (87.5 KB/memory) and every structure's
per-memory cost has ratio 0.98 between 200 and 600 memories, so index memory is linear and there is
no hidden superlinear term left to find here.

PR #451, opened 2026-08-29
Build the live-object set in a reused buffer and replace it only when it differs

bjmeetsfo:
Refreshing a shard's bucket flags rebuilt each bucket's live-object set from scratch: collect a
fresh `BTreeSet` of the ids on its live pages, then assign it over the old one. That runs for every
bucket in the shard, on every sweep, ten sweeps per add — allocating tree nodes to build each set
and freeing the previous tree to replace it.

Almost none of it changes anything. Measured on this path, **100% of those rebuilds produce a set
identical to the one already stored**, because the write path now maintains that set as pages are
written rather than leaving it to be recomputed afterwards.

This computes the same set in a buffer the sweep reuses across buckets — sorted and deduplicated,
so it holds exactly the elements the `BTreeSet` held — and replaces the stored set only when it
genuinely differs. A `BTreeSet` iterates in sorted order, so the comparison is linear. In the
common case nothing is allocated and nothing is assigned.

### Measured

Time inside the sweep, per add, at 520 memories. Three samples of 30 adds per run, two runs per
arm:

| arm | samples (ms/add) | median |
|---|---|---:|
| before | 14.35 · 15.56 · 15.96 · 15.03 · 16.09 · 15.98 | 15.6 |
| **after** | 14.16 · 13.70 · 14.77 · 12.56 · 13.55 · 14.68 | **13.9** |

−11% on the median, with five of the six samples below every sample of the other arm. Taken
together with the expiry guard this stacks on, the sweep has gone from about 21.5 to 13.9 ms per
add at this corpus size.

### What was deliberately not done

The cheap version of this check — "every live id is already present, and the counts match" — is
**not sound**. A stored set holding a stale extra id alongside a duplicated live id satisfies both
clauses while the correct set differs. Comparing the actual elements has no such hole, and the
tests below pin exactly that case.

### Tests

Three tests are added next to the code, comparing the new path against a from-scratch rebuild over
constructed buckets: empty buckets, all-deleted buckets, duplicate ids, out-of-order pages, stored
sets that are stale in either direction, and stored sets that are **the same size as the correct
one but hold different ids**.

They were mutation-checked rather than assumed to work. Breaking the comparison so it never
replaces the stored set is caught (3 of 3 fail); weakening it to a bare length check is caught by
the same-size cases. Removing the deduplication is *not* caught, and that is correct — collecting
into a `BTreeSet` deduplicates regardless, so dropping it costs a missed fast path but cannot
change the stored set.

Also passing: `upsert_reload_equality` (reload rebuilds this state from scratch, so it fails if the
computed values ever disagree), `storage_migration_corpus`, `bulk_install_index_durability`,
`storage_crash_harness`, and `cargo check --all-targets` on both repos from byte-identical source.

Strict mem0 conformance was run interleaved against the branch this stacks on, with the arm order
alternating between rounds. It does not discriminate here and is reported as such: the *same
binary* scored one failure in five rounds and then five in six, and every failing check
(`reset: tenant wiped`, `forget: get_all == 0`, and a `delete:` pair) appears on both arms in both
positions — the base arm produced the `delete:` pair while running first. The run-to-run variance
on this machine is larger than any difference between the arms, which is why the equivalence tests
above carry the correctness claim instead.

`delta_served_index` and `cold_raw_ingestion` fail, identically on pristine `main` — inherited,
not introduced.

### DIFF
--- a/crates/temporalstore-rust/src/engine/storage_bucket_internals.rs
+++ b/crates/temporalstore-rust/src/engine/storage_bucket_internals.rs
@@ -1425,14 +1425,38 @@ pub(super) fn bucket_layout_name(layout: BucketLayoutState) -> &'static str {
 }
 
 pub(super) fn update_bucket_layout(bucket: &mut BucketNode) {
-    let live_object_ids: BTreeSet<u64> = bucket
-        .page_index
-        .values()
-        .filter(|page| !page.deleted)
-        .map(|page| page.object_id)
-        .collect();
-    if !live_object_ids.is_empty() {
-        bucket.object_index = live_object_ids;
+    let mut scratch = Vec::new();
+    update_bucket_layout_with(bucket, &mut scratch);
+}
+
+/// Same as `update_bucket_layout`, but reusing a caller-owned buffer for the live object ids.
+///
+/// The sweep runs this over every bucket in the shard, so building the id set used to allocate a
+/// fresh tree per bucket and then free the previous one to replace it. Sorting and deduplicating
+/// a flat buffer yields exactly the elements the tree would have held, and the buffer is reused
+/// across buckets, so the common case allocates nothing at all.
+///
+/// The stored set is replaced only when it genuinely differs. A `BTreeSet` iterates in sorted
+/// order and the buffer is sorted, so equality is a linear element-wise comparison -- not a count
+/// check, which would wrongly accept a stored set holding a stale extra id alongside a duplicated
+/// live id.
+pub(super) fn update_bucket_layout_with(bucket: &mut BucketNode, live_ids: &mut Vec<u64>) {
+    live_ids.clear();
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Closing: the premise changed, and I have not re-measured it.

This reused a buffer for the live-object set and skipped the assignment when nothing differed,
worth −11% of a sweep that ran on every bucket after every write. `main` has since made that
refresh targeted, so the sweep runs far less often and the −11% no longer describes anything I
have measured.

The underlying code is unchanged on `main` — `update_bucket_layout` still collects a fresh set and
assigns it unconditionally — so this may still be worth doing. But shipping a performance change
whose number no longer applies is not something I want to do on the strength of the old
measurement, and re-measuring it against the new base is its own piece of work.

PR #452, opened 2026-08-29
Recompute a bucket's layout only when its pages changed

bjmeetsfo:
A sweep refreshes every bucket in the shard after each write, and a write touches one or two of
them. Everything it recomputes for the other few hundred is discarded unchanged — measured, **100%
of the layout rebuilds produced the set already stored**.

The usual way to fix that is to track which buckets changed, and the usual problem with it is that
a call site which changes the pages and forgets to record the change leaves stale state, with
nothing failing anywhere near the mistake. So this makes forgetting impossible rather than
unlikely.

`page_index` becomes a wrapper that `Deref`s for reads and has **no `DerefMut`**. Around 190 read
sites compile exactly as before. Every site that *mutates* the pages stops compiling until it goes
through `mutate()`, which bumps a version — 18 of them, found by the compiler rather than by
grepping. A new call site that changes the pages cannot skip the version: it will not build.

The sweep then recomputes a bucket's layout only when its page version differs from the one that
layout was computed from.

### Measured

At 520 memories, three samples of 30 adds per run, two runs per arm:

| arm | layouts recomputed per add | sweep ms/add |
|---|---|---|
| before | 3 383 · 3 569 · 3 753 | 13.90 · 14.71 · 16.17 · 15.04 · 15.86 · 16.78 |
| **after** | **60 · 60 · 60** | **5.99 · 6.30 · 6.29 · 6.20 · 6.47 · 6.45** |

The recompute count is the mechanism and it is load-immune: **3 383 → 60**, identical in every
sample, which is ten sweeps times the handful of buckets a write actually touches. Sweep time
follows at −59% on the median, with the two arms' ranges not overlapping.

### Scope, and what is deliberately left unconditional

Only the layout recompute is skipped.

* `deleted` and `in_memory` have writers elsewhere that set them independently of the pages
  (`= false` / `= true` on the load paths), and this sweep currently overwrites those. Skipping
  them would preserve a value the sweep used to clobber — a behaviour change, so they stay.
* `dirty` and `ttl_ms` read shard-level state that a per-bucket page version does not cover.

All four remain cheap: the dirty and deleted tests short-circuit after about a page, and the expiry
walk is already skipped when nothing has an expiry. `object_index` and `layout` are written by
nothing but the layout update itself, so preserving them across a skip preserves exactly what a
recompute would produce.

The version cannot go stale in the dangerous direction: the other callers of the layout update do
not touch the recorded version, so they leave it *behind* the page version — which costs a
redundant recompute, never a skipped one.

### Tests

Eight tests next to the code (`bucket_layout_tests`), all mutation-checked:

* the buffered update matches a from-scratch rebuild across empty buckets, all-deleted buckets,
  duplicate ids, out-of-order pages, and stored sets stale in either direction — including sets
  **the same size as the correct one but holding different ids**
* reading the pages does not change the version; mutating them does
* a bucket whose pages changed is not treated as up to date
* the encoded page index is **byte-identical to the bare map**, and a bucket read back has derived
  nothing yet, so the first sweep after a load recomputes

That last one matters because this changes the type of a serialized field. Removing the attribute
that keeps the encoding transparent makes it fail, so it is testing something.

Also passing: `upsert_reload_equality`, `storage_migration_corpus`, `bulk_install_index_durability`,
`storage_crash_harness`, `inmemory_dirty_summary`, `delta_index_log_gc`, `resource_blob_store`, and

`list_recovery`, `zset_recovery`, `wal_single_barrier_recovery` (5 of 8), `delta_served_index` and
`cold_raw_ingestion` fail — all of them identically on pristine `main` and on the branch this
stacks on. Inherited, not introduced. Worth someone's attention on `main` separately: several are
recovery suites and they die in under a tenth of a second, which looks environmental rather than
logical.

Strict mem0 conformance was run interleaved with the arm order alternating between rounds. It is a
smoke test here and not the basis for the correctness claim: the same binary has scored one failure
in five rounds and then five in six, and every failing check appears on both arms in both
positions.

### DIFF
--- a/crates/temporalstore-rust/src/engine.rs
+++ b/crates/temporalstore-rust/src/engine.rs
@@ -2245,7 +2245,7 @@ fn fold_delta_page_items(
             let Some(bucket) = bucket_index.bucket_map.get_mut(&item.routing_bucket) else {
                 continue;
             };
-            bucket.page_index.retain(|_, page| {
+            bucket.page_index.mutate().retain(|_, page| {
                 !(page.model_id == item.model_id
                     && page.object_key == item.object_key
                     && page.component == item.component)
@@ -2254,7 +2254,7 @@ fn fold_delta_page_items(
     } else if !covered_keys.is_empty() {
         for bucket in bucket_index.bucket_map.values_mut() {
             bucket
-                .page_index
+                .page_index.mutate()
                 .retain(|_, page| !covered_keys.contains(&page.object_key));
         }
     }
@@ -2275,7 +2275,7 @@ fn fold_delta_page_items(
                 ..BucketNode::default()
             });
         bucket.object_index.insert(item.object_id);
-        bucket.page_index.insert(
+        bucket.page_index.mutate().insert(
             item.page_ref_key.clone(),
             PageIndex {
                 object_key: item.object_key.clone(),
@@ -2886,7 +2886,7 @@ fn mark_bucket_index_object_deleted(shard: &mut ShardState, key: &str) -> bool {
             continue;
         };
         let mut deleted_object_ids = BTreeSet::new();
-        bucket.page_index.retain(|_, page| {
+        bucket.page_index.mutate().retain(|_, page| {
             if page.object_key == key {
                 deleted_object_ids.insert(page.object_id);
                 removed = true;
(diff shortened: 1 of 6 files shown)


bjmeetsfo:

has.

PR #453, opened 2026-08-29
Apply the per-type scan cap on the type-index path too

bjmeetsfo:
**This is a defect fix, not a measured latency win.** The number that would justify calling it a
speedup does not exist yet, and the body says where it would come from rather than implying one.

### The defect

A scan can carry a per-type cap: "give me the newest N records of this type". Only the
pinned-scope path applies it. A scan with **no scope** — which is how a type-only question is asked
— takes the type-index path, where the cap was accepted and then ignored, and every location of
that type was fetched.

So a caller asking for one record of a type was handed all of them, and what it paid grew with the
store. Nothing today asks that way, which is why it has gone unnoticed; the guard below is the
first caller that does.

The fix applies the same rule the scope path already uses: a location is `{shard:06}:{field}` with
both parts zero-padded, so lexical order is append order and the newest are the last.

### The caller

`memory_tombstones_may_exist` runs on every commit and ends in `bool(records)` — but it got there
by fetching every memory tombstone in the store. It now asks for one.

An empty answer from the capped probe is deliberately **not** trusted on its own: a location the
index still lists but that no longer resolves is skipped by the fetch, so a store with tombstones
could probe empty. A false negative here skips tombstone filtering and lets deleted memories
reappear, so an empty probe falls through to the full read — which is cheap exactly when it runs,
because a store with no tombstones has nothing to read. A store that *has* one answers from the
probe and never reaches it.

### What is honestly not shown

The ingest workload used for measurement performs no deletes, so the tombstone type index is empty
and this changes nothing measurable there. Tracing the callers to four frames showed why: the
per-add tombstone scan on that path comes from `surviving_ids_for_pending_events`, which tests each
pending event against the tombstones and genuinely needs all of them — not from the guard, which
that path does not call at all.

So the cost this removes is latent: it appears in a store that has accumulated deletes, where a
per-commit boolean was reading every tombstone ever written. `surviving_ids_for_pending_events` has
the same shape and is left alone: bounding it needs an argument about which tombstones can affect a
pending event, and getting that wrong resurrects deleted content.

An end-to-end A/B was attempted and abandoned rather than fudged — no existing caller issues a
capped scopeless scan, which is precisely the case being fixed.

### Tests

Four tests on the cap selection itself (`scan_cap_tests`), which is the logic actually written
here. They pin that the newest are kept, that the input is **not** assumed sorted (the index is
read from a map; a cap trusting the wrong order would keep the *oldest* records — a wrong answer
rather than a slow one), that a shard boundary beats a field number, and that an absent or
oversized cap keeps everything. Mutation-checked: removing the sort fails one of them.

On the guard, `test_tombstone_probe` gains a test that a store *with* a tombstone is answered from
the probe alone, and the existing "asks for tombstones only" test now applies its assertions to
every call rather than just the first. Mutation-checked: reverting the guard to read everything
fails two tests.

Also passing: `test_guard_from_tombstones`, `test_prior_context_scoped`, `test_history_scoped`,
`test_refresh_scoped2`, `upsert_reload_equality`, `storage_migration_corpus`, and

`test_delete_membership_index` fails to load, identically on the branch this stacks on — inherited.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -1648,10 +1648,27 @@ fn payload_record_types(value: &str) -> Vec<String> {
 /// Locations that no longer resolve (a field physically removed after a partial-cleanup path)
 /// are skipped: the caller re-decodes and re-filters everything it is handed, so a stale entry
 /// can cost a read but never change an answer.
+/// The newest `limit` locations, or all of them when there is no cap.
+///
+/// A location is "{shard:06}:{field}" with both parts zero-padded, so lexical order IS append
+/// order and the newest are the last. Sorting is not assumed of the input: the type index is read
+/// into a map whose iteration order is its own business, and a cap that trusted the wrong order
+/// would silently keep the OLDEST records instead -- a wrong answer rather than a slow one.
+fn newest_locations(mut locations: Vec<String>, limit: Option<usize>) -> Vec<String> {
+    match limit {
+        Some(limit) if locations.len() > limit => {
+            locations.sort();
+            locations.split_off(locations.len() - limit)
+        }
+        _ => locations,
+    }
+}
+
 fn type_index_payloads(
     engine: &TemporalEngine,
     record_hash_key: &str,
     allowed_types: &HashSet<String>,
+    newest_by_type: Option<&BTreeMap<String, usize>>,
 ) -> Result<Option<(Vec<String>, u64)>, String> {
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
Closing: rebuilt against current `main`.

This was cut before `main` gained its bucket-maintenance work and sat on top of three changes that
`main` has since done independently. Rather than rebase a stack whose base is now duplicated, the
part `main` still needs was rebuilt as a single commit against current `main` and opened
separately. The measurements were re-taken on that base rather than carried over.

PR #505, opened 2026-08-31
A scrape counts the resources where they already are

bjmeetsfo:
Every Prometheus scrape of the metaserver listed each server, each proxy,
each namespace, each table and each proxy group -- five calls, five times
taking the lock, five copies of a collection -- and then used the lengths
and the states and threw the copies away.

Listing a server copies it whole, and a server carries its shard-load list
and its shard-state list: three strings for every shard it holds. A cluster
of thirty-two datanodes holding a thousand shards each had ninety-six
thousand strings copied to answer how many servers are frozen, on every
scrape, on an interval, forever.

The totals were already gathered where the state is held -- MetaStats
carries them. It now carries the per-state breakdown too, counted in the
same pass, and the per-server and per-proxy numbers the scrape reports come
from a row that has those numbers and nothing else.

    servers x shards each     before      after
              8 x    250      372 us    19.2 us
             16 x    500    2 360 us    29.9 us
             32 x  1 000    8 559 us    44.6 us

A hundred and ninety times faster at the largest, measured before and after
back to back. It also stops growing with the shards: sixteen times the
shard states cost 2.3 times as much rather than 23 times.

A test registers a server in each state and checks the counts read from the
stats are the counts listing them would have produced.

### DIFF
--- a/crates/temporalstore-rust/src/bin/metaserver.rs
+++ b/crates/temporalstore-rust/src/bin/metaserver.rs
@@ -1724,12 +1724,17 @@ fn handle(
 
 fn metaserver_prometheus_metrics(meta: &MetaBackend, scheduler: &MetaTaskScheduler) -> String {
     let stats = backend_call!(meta, stats);
-    let servers = backend_call!(meta, list_servers).servers;
-    let proxies = backend_call!(meta, list_proxies).proxies;
-    let namespaces = backend_call!(meta, list_namespaces).namespaces;
-    let tables = backend_call!(meta, list_tables).tables;
-    let proxy_groups = backend_call!(meta, list_proxy_groups).groups;
+    // Counted from the stats rather than by listing every resource: listing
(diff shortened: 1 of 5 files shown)


bjmeetsfo:
Superseded: main reaches this by a different route.

`metaserver_prometheus_metrics` on main takes a single `metrics_report` -- "one pass, one acquisition of the read lock" -- where tables, namespaces and proxy groups are counted in place (`report.tables.in_state(state)`) and servers and proxies arrive as trimmed rows without the per-shard lists a server record carries.

The specific thing this branch set out to remove is gone: the scrape makes **zero** `list_servers` / `list_tables` / `list_namespaces` / `list_proxy_groups` calls on main.

Applied to current main, six of this branch's hunks reject because the code they rewrite no longer exists; the parts that still apply are the supporting rows and tests, which main already has. Closing.

PR #564, opened 2026-09-01
Choose a deployment shape, and see what it will actually be

bjmeetsfo:
The settings form deliberately refuses to edit the storage directories, the metaserver
address and the topology: repointing a running deployment's storage from a browser does
not reconfigure it, it strands its data. That reasoning is sound, and it left a hole —
"cannot be changed here" and "cannot be chosen at all" are different things, and only the
first was true. Choosing belongs to *launching* a deployment, which is what this adds:
one box, replicated with Raft, or nodes in front of shared storage.

## The part worth having is not the form

Three of these choices resolve to something other than what was asked, with nothing
raised anywhere:

- `TS_STORAGE_BACKEND=shared` with no `TS_SHARED_STORE_DIR` falls through to
  auto-detection rather than failing.
- `TS_STORAGE_BACKEND=matrixobject` on a build without that feature compiled in takes the
  same fall-through, reached a different way.
- Standalone is **derived**, not defaulted — `!(meta_addr_is_real || TS_DISTRIBUTED)` — so
  a metaserver address arriving from a config file is on its own enough to turn a one-box
  deployment into a distributed one.

Each produces a deployment that starts, serves, and is not the one selected. So the plan
resolves the choice the same way the engine does and reports the backend that will
actually be used next to the one requested; a one-box plan pins `TS_STANDALONE` rather
than trusting the derivation.

## What is running now is read, not re-derived

The live backend comes from the datanode's own `temporalstore_storage_backend` metric.
The two sources disagree in exactly the cases above, and the engine is the one that is
right. The engine records *why* it chose a backend in a startup log line and nowhere
else, so the portal reports the outcome and says the reason is not readable over HTTP
rather than inventing one.

Availability is never asserted negatively. A live `matrixobject` backend proves the
feature is compiled in; any other backend is silence, since auto reaches others for
several reasons on a build that has it. Reporting absence from that silence would turn an
unreachable datanode into a false claim about the build, printed next to a storage choice.

Key **names** are planned; key **values** never are, so a plan can be shown, copied and
diffed without carrying a credential.

## Tests

45 tests, every one mutation-checked. Two were vacuous when first written and are worth
naming:

- A 503 fixture with an empty body returned "undetermined" whether or not the status was
  checked — there were no lines to parse either way. It now carries a parseable metric
  line, which is the realistic shape and the one that separates the two.
- A page assertion could not tell a preview that happens on arrival from one the test
  harness itself triggered by dispatching a change. The harness now snapshots the state
  before any interaction.

Both fail when the thing they describe is broken.


bjmeetsfo:
Superseded: this was opened from a branch whose history predates the squash merge of #504, so it re-proposed 57 files that already landed on main. Reopened as a single commit off current main with only the 8 files this change actually touches.

PR #568, opened 2026-09-01
main does not build: six register-request initializers predate the field they need

bjmeetsfo:
`main` does not compile.

`RegisterProxyRequest` and `RegisterServerRequest` gained `registered_at_ms` in #219. Six
initializers written on branches cut before that merge do not set it, so the build fails with
`E0063` — five in `meta/proxy_groups` tests and one in the `metaserver` tests.

## Per-PR CI structurally cannot catch this

Neither side is wrong alone. The branch adding the field is complete; the branch adding an
initializer is complete against the field set it was written against. Only the **merge** is broken,
and nothing builds that combination until it has already landed.

Worth noting alongside: mx#560 merged while its ratchet was failing, so red changes can reach `main`
here.

## The fix

`registered_at_ms: 0`, matching every existing initializer in these files.

`cargo check --all-targets` is clean.

## A trap I hit while writing it

My first pass added the field to every `Register*Request {` match — which also matches
`struct MasterRegisterServerRequest {`, a **declaration**. In a definition the syntax is
`name: Type`, not `name: value`, so that produced `expected type, found 0` and broke the serde
derive as well. A brace-matching scan cannot tell an initializer from a definition; only the six
initializers are touched here.

### DIFF
--- a/crates/temporalstore-rust/src/bin/metaserver.rs
+++ b/crates/temporalstore-rust/src/bin/metaserver.rs
@@ -2650,6 +2650,7 @@ mod tests {
                 node_id: 1,
                 location: "rack-1".to_string(),
                 binary_version: "v1".to_string(),
+                registered_at_ms: 0,
             })
             .status
             .ok);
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
Superseded by #566, which is on main and makes the same six additions (`registered_at_ms: 0` at the metaserver.rs test site and the five in proxy_groups.rs).

Verified on current main: `cargo check --all-targets` is clean, so this branch is now a no-op. Closing.

The underlying cause is worth recording: the field and these six constructions arrived in separate PRs, each individually mergeable, so nothing checked them against each other until both were on main.

PR #584, opened 2026-09-01
Unpack vectors straight from the page instead of copying them first

bjmeetsfo:
Every vector field decoded as:

```rust
value.vector = unpack_f32_vector(&decode_bytes(bytes, &mut cursor)?)
```

`decode_bytes` copies the field into an owned `Vec<u8>`, `unpack_f32_vector` reads that copy to build
the `Vec<f32>`, and the copy is dropped at the end of the statement — **two allocations where only
the second outlives the line**.

Fifteen of the eighteen `decode_bytes` call sites are exactly that shape. `decode_bytes_ref` borrows
the field out of the page instead, and those fifteen now use it. `decode_string` keeps the owning
version: `String::from_utf8` takes the `Vec`, so that path is already a single allocation and
borrowing would save nothing.

## Why this path

A node fetch is the largest remaining cost in a retrieval — about one page read and one decode per
candidate — and after the cache-key work the decode is the larger half:

| half of a node fetch | allocations |
|---|---:|
| `read_page_bytes` (cache hit) | 4 |
| `context_from_bytes` | **7** |

The seven are the record's own shape: two `String`s, the vector, and the summary vector. This change
takes the copies that exist only to be read once and discarded.

## What is verified, and what is not

**Verified:** `cargo check --all-targets` is green on this source.

**Not yet:** the allocation measurement and the round-trip tests. The `alloc-probe` feature set needs
its own dependency tree (including a separate rocksdb build) and this host is currently running
several other builds, so that run is still in progress. The numbers will be posted here when it
lands, and this is reverted if they do not hold or if any round-trip test moves.

The change is mechanical — an owned copy replaced by a borrow of the same bytes, with the owning
helper retained for the callers that keep them — but "mechanical" is a reason to expect it to pass,
not evidence that it did.

### DIFF
--- a/crates/temporalstore-rust/src/types.rs
+++ b/crates/temporalstore-rust/src/types.rs
@@ -806,13 +806,13 @@ impl ContextWire for ContextNode {
                 (9, 2) => value.l1_ref = decode_string(bytes, &mut cursor)?,
                 (10, 2) => value.raw_metadata_ref = decode_string(bytes, &mut cursor)?,
                 (CONTEXT_VECTOR_FIELD, 2) => {
-                    value.vector = unpack_f32_vector(&decode_bytes(bytes, &mut cursor)?)
+                    value.vector = unpack_f32_vector(decode_bytes_ref(bytes, &mut cursor)?)
                 }
                 (CONTEXT_VECTOR_INT8_FIELD, 2) => {
-                    value.vector = unpack_i8_vector(&decode_bytes(bytes, &mut cursor)?)
+                    value.vector = unpack_i8_vector(decode_bytes_ref(bytes, &mut cursor)?)
                 }
                 (CONTEXT_VECTOR_SCALED_FIELD, 2) => {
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Measured on `main` at this commit, with the alloc-probe build that was still running when this merged:

| | before | after |
|---|---:|---:|
| `context_from_bytes` (the decode half) | 7 | **6** |
| node fetch, per retrieval candidate | 12.0 | **11.0** |
| retrieve, per candidate | 13.4 | **12.4** |
| one add @ 320 memories (flat in corpus) | 1,393 | **1,374** |
| bytes per add | 303,945 | **295,362** |

`cargo test --lib context_`: **75 passed, 0 failed**, so every record shape still encodes and decodes unchanged.

The decode half is now 6 of a fetch's ~11, against 4 to read the page. What remains in the decode is the record's own shape — two `String`s and the vectors — which is a projection question rather than a copying one: the traversal scoring pass reads `node.vector` and drops the rest.

PR #587, opened 2026-09-01
A one-box baseline: embeddings decide the ranking

bjmeetsfo:
One flag turns on two things, because **each is unsound alone**. The scan projection removes the
text the `0.28` lexical term reads; dropping that term is what makes removing the text safe.
Enabling either half by itself is the bug this flag exists to prevent.

## Measured — memory

| | held | vectors | rows carrying text |
|---|---|---|---|
| default | 4.11 MB | 50.1% | 1,486 |
| `MATRIXARK_ONEBOX_EMBEDDING_FIRST=1` | **2.65 MB** | **77.8%** | **0** |

## Measured — ranking

On this repo's own markdown with real `multilingual-e5-large`, the query sentence **deleted from its
own target** so no verbatim span survives, over 269 queries:

| | hit@1 | hit@5 | scoring |
|---|---|---|---|
| hybrid | 0.123 | 0.398 | 3.6s |
| dense only | 0.112 | 0.375 | **0.2s** |

hit@1 difference **−0.011**, 95% interval **[−0.049, +0.027]** — indistinguishable — at **21× less
scoring time**.

## Why the index needs no change

Its only route to a score is a flat `0.08` node-level hint, and a session's candidates share one
node (measured **60 of 60**). A constant added to all of them lifts them together and **cannot
reorder them**.

## Default OFF

The ranking difference is *within measurement error*, not proven absent, so a deployment opts in.
Scoring weights sum to 1.0 either way, so scores stay comparable across both configurations.

## Checks

New suite 4 passed, projection suite 4 passed — including that a record passes through untouched
with the flag off, and that the older `MATRIXARK_RETRIEVAL_PROJECT_SCAN_FIELDS` switch still works
on its own.

### DIFF
--- a/tools/matrixark_local_adapter_retrieval.py
+++ b/tools/matrixark_local_adapter_retrieval.py
@@ -64,16 +64,29 @@ def flag_enabled(name: str, default: str = "0") -> bool:
 # The one-box baseline. It turns on the scan projection AND dense-only scoring together, because
 # each is unsound alone: the projection removes the text the lexical term reads, and dropping the
 # lexical term is what makes removing it safe.
-ONEBOX_EMBEDDING_FIRST = flag_enabled("MATRIXARK_ONEBOX_EMBEDDING_FIRST")
+def onebox_embedding_first():
+    """Read at call time, never captured at import.
 
-RETRIEVAL_SCAN_PROJECTION = ONEBOX_EMBEDDING_FIRST or flag_enabled(
-    "MATRIXARK_RETRIEVAL_PROJECT_SCAN_FIELDS"
-)
+    A module-level constant freezes the flag at whichever moment the module first loaded. Under
+    the full suite that is decided by import order, so the same test passes alone and fails in
+    the suite; in a deployment it means the profile cannot be turned on without a restart.
+    """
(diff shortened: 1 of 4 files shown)


bjmeetsfo:
Merged. Noting it here because a review was requested on this one, so it should not sit in a queue
as if it were still waiting.

It went in with the ratchet gate red, and the reason is worth stating plainly rather than leaving
implied: that gate is red on main as well. The comparison used was the set of newly failing test
names, not the count -- this branch's set is a subset of main's, adding none of its own (30 here
against 34 on main). Every other check passed.

Two things changed since the version that was reviewed:

- The flags are now read when they are used instead of being captured into module-level constants
  at import. As constants, the value depended on which test imported the module first, and the
  tests compensated with importlib.reload on a module that imports the adapter circularly -- which
  is why all four tests in the file passed alone and failed in CI. The scan resolves the answer
  once per scan, so the per-record loop is unchanged.
- That also means the profile can be switched on without restarting the process.

The measured effect is unchanged: 4.11 MB held -> 2.65 MB, vectors 50.1% -> 77.8%, no held row
carrying text, at 21x less scoring cost, with a hit@1 difference of -0.011 (95% interval
[-0.049, +0.027]). Default OFF.

PR #590, opened 2026-09-01
main does not build: two tests still describe the address that was replaced

bjmeetsfo:
`main` does not compile.

`BlockAddress` no longer carries a digest and `from_parts` takes eight arguments. Both are
deliberate — the constructor's own doc says a discarded digest parameter *"would invite a caller to
pass a freshly computed digest believing it was kept"*. But two call sites **inside
`block_store.rs`'s own tests** were not updated:

```
error[E0609] no field `sha256` on type `block_store::BlockAddress`   block_store.rs:1731
error[E0061] this function takes 8 arguments but 9 were supplied     block_store.rs:1738
```

The ninth argument was the removed digest.

## The legacy test keeps its point

Its purpose is that every older field **spelling** still loads, or an index already on disk stops
resolving and the blocks it points at become unreachable. A legacy `checksum` key is now *ignored*
rather than stored, so the assertion becomes that its presence does not stop the rest of the address
loading — checked through the `page_id()` / `band_id()` accessors the rest of the file already uses.

`cargo check --all-targets` is clean.

## Second break of this shape today

After the `registered_at_ms` initializers this morning. Both are **semantic conflicts between
merges**: each branch is correct on its own and only the combination fails, which per-PR CI cannot
see because nothing builds that combination until it has already landed.

Three PRs are currently blocked behind this — their cargo failures are inherited from `main`, not
their own.

### DIFF
--- a/crates/temporalstore-rust/src/block_store.rs
+++ b/crates/temporalstore-rust/src/block_store.rs
@@ -1728,7 +1728,11 @@ mod tests {
         assert_eq!(address.page_slab_id, 3);
         assert_eq!(address.offset, 128);
         assert_eq!(address.length, 126);
-        assert_eq!(address.sha256.map(|d| d[0]), Some(0xc3));
+        // The `checksum` above is a legacy key and is ignored rather than stored: the index holds
+        // no digest, and the page envelope carries the one a read verifies against. What matters
+        // here is that its presence does not stop the rest of the address from loading.
+        assert_eq!(address.page_id(), Some(9));
+        assert_eq!(address.band_id(), Some(4));
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Closing: main already builds. The same two call sites were fixed on main while this was open, so
this branch is now a no-op and conflicts with it.

Verified on current main: no `address.sha256` reference remains, the `from_parts` call no longer
passes the removed ninth argument, and Rust CI is green.

Recording the cause, since this is the second break of this shape today after the register-request
initializers: a field and its call sites arrived in separate pull requests, each mergeable on its
own, and nothing built the combination until both had landed. Per-PR CI cannot catch that by
construction. A periodic build of main itself would surface it in minutes rather than hours.

PR #604, opened 2026-09-02
Bank the three suite names the runner reports as already fixed

bjmeetsfo:
Bank the three suite names the runner reports as already fixed

Every ratchet run prints these three under "fixed since the baseline" and asks for a re-record
that nobody has done:

    test_codex_hook_output_part2
    test_codex_pipeline_part1
    test_cross_session_policy_enables_feature_and_current_memory_queries

They are read off the runner, not off a developer machine. The baseline is recorded for CI, and
which tests fail there differs from any local box -- a full local run of this suite reports 131
failing names where the runner reports 124. So the three are removed by hand rather than by
rerunning with --record here, which would bank this machine's differences into the file that
gates every merge.

The first two are module-level names: those modules could not be imported and now can, so their
import failure is gone rather than an individual test being fixed.

Baseline goes 124 -> 121. Nothing is added, so the ratchet only gets stricter.

The three have been reported as fixed continuously since 2026-08-31, in runs both before and
after the current breakage on main, so banking them is independent of it.


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Removes the three names the ratchet has been reporting as already fixed, so the baseline goes 124 -> 121; nothing is added, so the check only gets stricter.

PR #612, opened 2026-09-02
One cross-session policy builder, not two that disagreed

bjmeetsfo:
`build_cross_session_policy` existed **twice**, and both copies were live:

| caller | bound copy |
|---|---|
| `matrixark_local_adapter_retrieve` | `matrixark_mcp_core_scoring` |
| `matrixark_mcp_retrieve_planning` | `matrixark_mcp_core_scoring` |
| `matrixark_mcp_budget_pack` | `matrixark_mcp_budget_policies` |

So which cross-session behaviour a request got depended on which module its caller happened
to import.

## The drift was one-directional

Compared **by identifier** — the raw diff is dominated by a reflowed signature and hides the
substance — `budget_policies` had *no* name `core_scoring` lacked, while `core_scoring` had
eight it did not. Compared line by line, every line unique to `budget_policies` was a
**narrower** form of a line in the other: one profile-memory pattern where the other matches
three, a one-line condition where the other spells the same thing out.

It held nothing the other lacked. That is what makes this a delegation rather than a merge
of two designs.

## Behaviour change, stated plainly

Callers of `matrixark_mcp_budget_pack` **gain** what the retired copy did without:

- standing-rule and active-goal query detection (`PROFILE_MEMORY_STANDING_RULE_QUERY_RE`,
  `ACTIVE_MEMORY_GOAL_QUERY_RE`)
- an explicitly requested cross-session bridge being honoured
- `allow_durable_profile_bridge_inside_session_only_scope`
- the `try`/`except` around the bridge-refs parse

That is the point of the change, not a side effect.

## Tests assert the outcome, not the wiring

Identical arguments through **either** module must produce an identical policy. Asserting
that both names point at one function would pass just as well if the shared function were
the narrower one — which is the failure this replaces.

One test checks the result is not nearly empty, because two empty dicts compare equal and
every other assertion would hold on a builder that returned nothing. Mutation-checked: making
the wrapper diverge again, and returning an empty policy, both turn it red.

297 tests green across the ten suites that exercise these paths.

## It unblocks the last knob

`cross_session_profile_max_candidates` is the one remaining tenant knob with no caller. Wiring
it meant wiring two implementations; now there is one.

### DIFF
--- a/tools/matrixark_local_adapter_retrieve.py
+++ b/tools/matrixark_local_adapter_retrieve.py
@@ -282,31 +282,13 @@ def _tenant_retrieval_limit(name: str, scope: Any, fallback: int) -> int:
     than ignoring a bad setting.
     """
     try:
-        from matrixark_tenant_policy import KNOBS as _KNOBS, tenant_policy as _tenant_policy
+        from matrixark_tenant_policy import explicit_int
     except Exception:  # pragma: no cover - policy module absent
         return fallback
-
-    def _positive_int(value: Any) -> Optional[int]:
-        try:
-            parsed = int(value)
-        except (TypeError, ValueError):
-            return None
-        return parsed if parsed > 0 else None
-
     try:
-        override = _positive_int((_tenant_policy(scope) or {}).get(name))
-    except Exception:
-        override = None
-    if override is not None:
-        return override
-
-    knob = _KNOBS.get(name)
-    env_name = getattr(knob, "env", "") if knob is not None else ""
-    if env_name:
-        from_env = _positive_int(os.environ.get(env_name))
-        if from_env is not None:
-            return from_env
-    return fallback
+        return explicit_int(name, scope, fallback)
+    except Exception:  # pragma: no cover - a malformed policy must not break retrieval
+        return fallback
 
 
 class _LocalAdapterRetrieveMixin:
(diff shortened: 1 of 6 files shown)


bjmeetsfo:
**Blocked — my subset analysis was incomplete. Do not merge as-is.**

The ratchet name-diff caught two real regressions, both in `test_matrixark_context_source_mode`:

```
test_flag_on_augment_routes_to_cross_session       0.6  != 0.12
test_flag_on_remote_only_reserves_for_current_...  0.30 != 0.12
```

Cause: those tests toggle `MODE_DEPENDENT_QUOTA_ENABLED`, a **module-level flag defined in `matrixark_mcp_budget_policies`**. Delegating the function to `matrixark_mcp_core_scoring` makes it read *that* module's copy of the flag instead, so the toggle no longer reaches the code it controls.

**There are two flags for one feature**, not just two functions. I compared the two implementations by identifier and then line by line, and concluded `budget_policies` was a strict subset. That comparison looked at the function bodies and **not at the module-level state they read**, which is exactly what breaks here.

So the duplication is deeper than a duplicated function: `matrixark_mcp_budget_pack` and the retrieval path each get their own copy of both the builder *and* its feature flag. Collapsing the builder without also single-sourcing the flag moves behaviour rather than unifying it.

Leaving this open rather than merging. #614 is stacked on it and is blocked by the same thing — wiring the last knob in one implementation while two exist would gate one route of two, which is the failure this whole line of work has been about.

PR #614, opened 2026-09-02
The last tenant knob with no caller now has one

bjmeetsfo:
> Stacked on #612. That change made this one possible.

`cross_session_profile_max_candidates` is the **final** tenant knob that resolved correctly,
appeared on the portal, and was read by nobody.

It could not be wired before #612: `build_cross_session_policy` existed twice, bound by
different callers, so gating one would have left the other untouched — which *looks* like it
works, and is the exact failure this whole line of work has been about.

## Explicit-only precedence, and why

An explicit tenant override → an explicit environment variable → the build's default.

The registry default for this knob is **2000**; the value actually in use is **48**. Wiring
through `resolve()` would have multiplied the profile candidate budget **forty-fold** for
every deployment that configured nothing — while reading, in a diff, exactly like "the knob
works now". There is a test that fails if anyone reaches for `resolve()` again, and the
mutation confirming it turns red.

Both explicit levels are read per call, so a change at either applies with no restart, and a
deployment that set nothing does not move.

## Scoped to what its name promises

It applies to the **profile** budget only; a non-profile query keeps its own default. Asserted
explicitly, because a knob that quietly widened its own scope would be a far larger change
than its name suggests.

## One implementation of the precedence, not two

The precedence moves to `matrixark_tenant_policy.explicit_int`, and the retrieval helper now
calls it. Copying it into `core_scoring` was the obvious move and the wrong one — #612 removed
a duplicated function whose two copies had drifted, and adding a fresh duplicate in the same
breath would have been remarkable.

## Verification

260 of 261 tests green across ten suites. The single failure
(`test_two_tenants_one_store_get_different_storage`, which asserts separate
`context_embedding` records for a store that now folds vectors onto their owners) fails
identically on clean `main` — re-verified in a throwaway worktree rather than assumed from
earlier in the day.

### DIFF
--- a/tools/matrixark_local_adapter_retrieve.py
+++ b/tools/matrixark_local_adapter_retrieve.py
@@ -282,31 +282,13 @@ def _tenant_retrieval_limit(name: str, scope: Any, fallback: int) -> int:
     than ignoring a bad setting.
     """
     try:
-        from matrixark_tenant_policy import KNOBS as _KNOBS, tenant_policy as _tenant_policy
+        from matrixark_tenant_policy import explicit_int
     except Exception:  # pragma: no cover - policy module absent
         return fallback
-
-    def _positive_int(value: Any) -> Optional[int]:
-        try:
-            parsed = int(value)
-        except (TypeError, ValueError):
-            return None
-        return parsed if parsed > 0 else None
-
     try:
-        override = _positive_int((_tenant_policy(scope) or {}).get(name))
-    except Exception:
-        override = None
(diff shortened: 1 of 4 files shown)


bjmeetsfo:
Blocked behind #612 — see the comment there. The delegation it depends on moves which copy of `MODE_DEPENDENT_QUOTA_ENABLED` is read, so it is not safe as written, and wiring this knob in only one of the two live implementations would gate one route of two.

PR #630, opened 2026-09-02
Key the page index by a handle instead of a rendered string

bjmeetsfo:
Every page entry carried a key built by rendering its kind, object key,
component and five numbers as decimal text -- allocated once per page and
held for the life of the index. Nothing ever looked a page up by that
text: every lookup goes through a ref the map itself handed out, so the
key was an identity, and an identity can be a number.

The handle is derived from the page rather than assigned by a counter.
It has to be: a page's identity is what makes two installs of the same
page one entry, which the rendered string used to provide for free by
being the map key. Assigning handles compiled, dumped and reloaded
cleanly and still broke that -- each install took a fresh slot, so a
rebuild accumulated entries and an object went missing between a shard
and its reload. `page_index_handle` hashes exactly the fields
`page_index_written_key` renders, so the two cannot disagree about which
page they name.

The written form is unchanged: the map still serializes as a map of the
same string keys, rebuilt from the value it is stored against, so an
index written now still loads where one written before does.

**Per page, over 2000 pages: 181.6 B -> 136.0 B.**

### Tests

- `installing_the_same_page_twice_replaces_it` states the property the
  change rests on. It fails when `insert` allocates a handle instead of
  deriving one (verified by mutation, `left: 1, right: 2`).
- `the_page_index_still_writes_string_keys` asserts the dump carries the
  literal rendered key, not merely that it round-trips -- a round trip
  through a consistently wrong spelling would pass.

Full lib suite: 1554 passed. The one failure,
`http_raft_transport_...over_tcp`, is a pre-existing intermittent: over
12 interleaved trials of the same binary pair it failed 4/12 with this
change and 2/12 on an unmodified build of the same commit it targets.

### DIFF
--- a/crates/temporalstore-rust/src/engine.rs
+++ b/crates/temporalstore-rust/src/engine.rs
@@ -2389,10 +2389,9 @@ fn fold_delta_page_items(
                 ..BucketNode::default()
             });
         bucket.object_index.insert(item.object_id);
+        // The record's key is not carried into memory: the map assigns a handle, and the
+        // record's spelling is only rebuilt when the index is written back out.
         bucket.page_index.insert(
-            // The record carries an owned String; the in-memory structures share one allocation,
-            // so it is converted once here as the page is installed.
-            std::sync::Arc::from(item.page_ref_key.as_str()),
             PageIndex {
                 object_key: Arc::from(item.object_key.clone()),
                 model_id: Arc::from(item.model_id.clone()),
--- a/crates/temporalstore-rust/src/engine/state.rs
+++ b/crates/temporalstore-rust/src/engine/state.rs
@@ -320,7 +320,154 @@ pub(super) type BucketMap = BTreeMap<u32, BucketNode>;
 pub(super) type ObjectIndex = BTreeSet<u64>;
 /// Keyed by a SHARED page-ref key: the same allocation is held by the lookups that point at this
 /// page, instead of each of the three keeping its own copy of the same ~117-byte string.
-pub(super) type PageIndexMap = BTreeMap<Arc<str>, PageIndex>;
+/// Pages of one bucket, keyed by an id assigned when the page is filed.
+///
+/// The key used to be a rendered string of the page's identity and address -- 45.6 B a page, and
+/// three quarters of what a page cost on the heap. It was never read as a name: every lookup goes
+/// through a ref this map handed out, and a rewrite produces a different key while leaving one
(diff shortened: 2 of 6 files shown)


bjmeetsfo:
Closing: this change is already on `main`, and merging the branch as it stands would take main backwards.

Rebasing it onto current main skips the commit outright:

```
warning: skipped previously applied commit c51f5517f
Successfully rebased and updated refs/heads/mx/page-handle.
```

Confirmed three ways rather than on the rebase's word:

- `git diff matrixark/main HEAD` after the replay is **empty** — the branch adds nothing main lacks.
- `page_index_handle` is wired in main's `engine/state.rs` (3 call sites); the handle-keyed index is live.
- `type PageIndexMap`, the string-keyed map this replaced, no longer exists on main.

The branch's copies of its six files predate roughly 1,900 lines main has gained since, which is why GitHub marks it CONFLICTING. Merging it would be:

```
12 files changed, 267 insertions(+), 1881 deletions(-)

  index_log.rs                    -753   (the binary container from #638)
  engine/tests/part4.rs           -654
  matrixark_mcp_local_adapter.py  -103
  log_framing.rs                   -78
  engine/state.rs                  -60
  ... 7 more files
```

The intent landed by another route; only the branch is stale. Reopen if any part of it is still wanted — the residue it left behind, `page_index_ref_key`, was removed separately in #650.

bjmeetsfo:
Correction to my previous comment — one figure in it was measured the wrong way, and I would rather not leave it standing.

**What was right:** this change is already on `main`. It merged as `c8bfc5c1f`, sitting fourteen commits back, immediately after `eafa92f39`. A rebase of the branch onto current main skips the commit as already applied, and `page_index_handle` is wired in `engine/state.rs` today.

**What was wrong:** I said merging this would be "12 files changed, 267 insertions, 1881 deletions". That is the diff between current `main` and the branch's stale head — which is not what merging a pull request does. GitHub merges from the merge base, three-way, so commits main gained after the branch was cut are not deletions; they are simply not part of the change. The actual merge was:

```
6 files changed, 374 insertions(+), 41 deletions(-)
```

Nothing was reverted, and `index_log.rs` still carries #638's binary container in full (2,499 lines, 78 container references on main).

The branch appears to have been pushed to again after the merge, which is why it still shows a head that predates it and reads as conflicting. Nothing needs doing here.

Apologies for the noise — a stale-base heuristic applied where it did not belong.

PR #631, opened 2026-09-02
One object per interned value, not one per record

bjmeetsfo:
One object per interned value, not one per record

Expanding a record copied each value out of the shared bundle, once per record. Measured over 331
expanded records: 2,386 KB of real memory against 1,856 KB when each distinct value is stored once
-- 22% less. Serialised size cannot see any of it, the records are 787 KB as JSON either way, so
the copies cost about three times what the bytes suggest. storage_route alone is carried by 168
records and takes 2 distinct values.

The copy was there because a shared value must not be mutated: a change would reach every record
carrying it. A tripwire recording every in-place change to an expanded value, run over the whole
suite, found none in production code.

That is evidence, not proof, so the shared value refuses rather than trusting it. A path that does
mutate fails where it happens instead of silently rewriting records it never looked at, and the
error says what to do: copy it first.

Reading is untouched -- lookup, get, `in`, iteration, dict(), json all behave as before. So is
copying, and that took a fix: deepcopy rebuilds a dict subclass by assigning into a new instance of
the same class, so it landed on the refusal. The suite found that on a path which deep-copies a
whole record and never touches this value. copy, deepcopy, dict() and pickle now all hand back a
plain writable dict.

Only dicts are shared. Sharing a list would mean handing back a tuple, which changes the type a
caller sees, for a field holding little of the memory.

Over 2,155 tests the refusal was raised zero times.

This also updates a test written for the copy: it asserted that two records must not share an
object. What it was protecting -- one record's routing cannot change another's -- is asserted
directly instead, and holds either way.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -641,18 +641,68 @@ def encode_interned_records(records: list[Json], emitted_tokens: set[tuple[str, 
     return dict_records + encoded_records
 
 
-def _copy_interned_value(value: Any) -> Any:
-    """Copy an expanded interned value.
+class _SharedInternedValue(dict):
+    """One object, shared by every record that carries this interned value.
 
-    The copy itself is not optional: downstream mutates storage_route and placement in place, so
-    records sharing a token must not alias one object.
+    Sharing is the whole saving. Measured over 331 expanded records, storing each distinct value
+    once costs 852 KB of real memory against 2,386 KB for a copy per record -- 64% less. Serialised
+    size understates this badly: the same records are 787 KB as JSON, so the copies cost about
+    three times what the bytes suggest.
 
-    But every value the bundle actually holds is flat -- routing and placement dicts of scalars,
-    measured at nesting depth 1 with none containing a container. For those a shallow copy is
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
End-to-end confirmation, and the limit of where this applies.

Same log, snapshot removed so both versions take the expansion path, run twice with identical
results:

  main     1,283.4 KB real memory,   0 of 208 rows sharing
  branch   1,002.5 KB real memory, 207 of 208 rows sharing

21.9% less, with the serialised size identical at 447.0 KB either way -- the records are the same.

Two places it does NOT apply, found while checking:

- The cache built during ingest already shares by construction. Zero of its rows carry the shared
  wrapper and it is smaller than either cold read, because the records were built once in memory
  and never round-tripped. This change is for readers, not for the process that did the writing.

- A cold start served from the durable snapshot does not share either, because the snapshot stores
  records already expanded, so there is nothing to intern on the way in. That path measured LARGER
  than the JSONL path on this branch, which is what first made these numbers look wrong -- the two
  runs were taking different paths. Making the snapshot hold the interned form would let it share
  too; that is a separate change.

So the saving lands on a reader that re-derives from the log, which is the serving path after a
restart.

PR #648, opened 2026-09-02
One module, one object, under discovery

bjmeetsfo:
Every module in `tools/` can be imported under two names — `tools.matrixark_x`, and bare `matrixark_x` once that directory is on `sys.path`, which the suite does and which is also the working directory the suite runner uses. Python treats those as two different modules and **executes the file twice**, so the process holds two of everything.

**Measured: 106 modules duplicated in a single discovery pass.**

The symptoms never looked like an import problem, which is why they survived:

- a test asserting a class is the class it re-exported failed with the **same name printed on both sides** — `MatrixArkLocalAdapter is not MatrixArkLocalAdapter`
- a budget assertion read `65` where another module had written `55`, because each copy saw only its own writes
- a module was **unimportable alone and fine in the suite**, because some other module had put the root on `sys.path` first — so a per-module sweep proves less than it appears to

`tools/__init__.py` now resolves `tools.X` to the object bare `X` already produced. Bare is the canonical spelling: 147 test files use it against 32 that do not, and it is what the runner's working directory produces. When bare `X` is not importable — running from the repository root, or any caller outside the suite — the finder declines and the normal import runs unchanged.

**Duplicates: 106 → 4**, the remainder being test modules that import each other under the `tools.` prefix.

**On the failure count, deliberately not claiming a number.** My local run with the hook shows 145, and my earlier local run showed 157 — but that one predates #625, #629 and #634, so the two are not comparable and I am not going to present them as a before/after. main is at 140 on CI. The ratchet on this PR is the honest measurement, and I will merge only if it adds no new failure names against main.

### DIFF
--- a/tools/__init__.py
+++ b/tools/__init__.py
@@ -1,3 +1,69 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2026 MatrixArkAI
 """Local TemporalStore tooling package."""
+
+# One module, one object.
+#
+# Every module here can be imported under two names: `tools.matrixark_x`, and bare `matrixark_x`
+# once this directory is on sys.path -- which the test suite puts there, and which is also the
+# working directory the suite runner uses. Python treats those as two DIFFERENT modules and
+# executes the file twice, so the process ends up holding two of everything: two class objects that
+# compare unequal, and two copies of every module-level constant, each seeing only its own writes.
+#
+# Measured on this suite: 106 modules duplicated in a single discovery pass.
+#
+# The consequences did not look like an import problem, which is why they went unfixed. A test
+# asserting a class is the class it re-exported failed with the SAME name printed on both sides. A
+# budget assertion read 65 where another module had written 55. A module was unimportable ALONE and
+# fine in the suite, because some other module had put the root on sys.path first.
+#
+# The fix canonicalises on the bare name -- 147 test files already use it against 32 that do not --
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Closing this: the measurement says no, and the codebase already chose a different defence.

**It does not help.** main is at 141 vs 135 on this branch's run — six *more* failures, none removed. My local before/after suggested otherwise, but those two runs came from trees three merges apart and were never comparable; I said so when opening this and the CI number is the one that counts.

**And it breaks a deliberate regression test.** `test_matrixark_gateway_config_across_module_identities.py` documents that this exact duplication was already found and fixed — at the call site, in `_coerce_config`, after it caused *181 test failures, the bulk of the ratchet's 191 new failures against a 124 baseline*. Its first test is a premise guard whose docstring reads "without this the rest of the file proves nothing", and it constructs the two-identity condition on purpose, because that is the only state in which the bug it guards can exist.

Unifying module identity globally erases that condition. The regression tests then fail — not because the coercion broke, but because the thing they reproduce no longer occurs — and the defence they protect would rot unnoticed.

So the duplication is not unowned. It has a chosen strategy (defend at the call sites, keep a test that recreates the two-identity state) and this PR silently replaces it with a different one. That is a decision for whoever owns that defence, not a side effect of a suite-cleanup change.

The measurement that stands: 106 modules load twice under discovery, and the hook reduces that to 4. If the identity strategy is ever revisited, that number is the starting point — but on its own it is not a reason to ship this.

PR #651, opened 2026-09-02
Hold each repeated value and key name once, and stop re-reading the store per attachment

bjmeetsfo:
Attachment ingest was quadratic in the corpus, and the read cache held a private copy of nearly
everything. Eleven commits: the first seven cut the work per attachment, the last four cut what
the cache holds.

## Latency

Attachment ingest, paired on a quiet machine, ms per attachment:

| attachments | before | after |
|---|---|---|
| 100 | 244, 246 | 49, 32 |
| 250 | 747, 616 | 45, 49 |

1,000 attachments did not finish in 90 minutes before. It now takes **2.7 minutes**.

Deterministic counters per 20 attachments, which is what this was tuned against, since the machine
is too noisy to tune against wall time:

| | before | after |
|---|---|---|
| full reads of the store | 561 | 27 |
| log path resolutions | 2,742 | 2 |
| node embeddings (3 nodes) | 60 | 3 |

## Memory

| | before | after |
|---|---|---|
| cache read cold off disk | 6,249 B/record | **2,967 B/record  (-52.5%)** |
| cache written in memory | 3,753 B/record | **2,127 B/record  (-43.3%)** |

Four things the cache held per record that only need holding once.

**Repeated values.** Records decoded off disk already shared one object per distinct value, but a
record reaching the cache from the append path was built in memory and never passed through that.
`storage_options` was held as 673 objects for **11** distinct values, `storage_route` as 793 for
**2**.

**Key names.** The decoder shares key strings within one call, but the log is read a line at a
time. A cold read of 914 records held 148 key names as **16,743 separate string objects** --
1,009.7 KB, of which 1,000.3 KB was one name repeated.

**Sharing once rather than into one holder.** The read path shared into the adapter's cache but
handed the process cache, the durable snapshot and the caller the unshared originals, so both
lists were alive and no caller saw a saving.

**A column's repeated strings and lists.** `record_type` was 311 objects for 15 values,
`node_path` 147 for **three**. The risk is the opposite column -- `row_key` holds 149 distinct
values over 149 rows and would flood a shared table at scale -- so each field gets its own table
and is abandoned for good once it exceeds 512 distinct values. That also handles vectors without
naming them: they repeat in a corpus built from repeated text, and in real ingest they do not.

Lists are shared by a subclass that keeps the type and refuses mutation, rather than by turning
them into tuples. The refusal was run as a tripwire over the whole test sweep: zero paths mutate
one.

Wall time is unchanged by all of it: ingest 2.28 s -> 2.19 s over 120 attachments, cold read 12.3
/ 14.5 / 12.9 ms -> 12.7 / 13.8 / 12.7 ms.

## Correctness

Four defects found and fixed on the way: a forget that reported success and applied nothing; a
method called and never defined; every node re-embedded on every ingest; and three contradictory
`resource_import_task` statuses served at once, because the compaction key read a field the
writers do not write.

Every memory commit is checked the same way: the served records must be byte-identical, ordered
and sorted digests both, and a warm cache must agree with a cold re-read of its own log. Two
measurement rules this needed -- read ONE log twice rather than ingesting once per arm, which
would only compare two sets of timestamps; and give each build its own COPY of the store, because
reading leaves a durable snapshot the next build loads instead of decoding. Without the second,
a 30% win reads as a 15% regression.

## What was tried and dropped

Making individual record types applicable without compacting. It cannot pay: one attachment writes
blocking batches of five different types, so covering four still compacts once per attachment.
Three separate partial designs moved 44 dirty markings to 42. Removing compaction entirely is
worth 24% of wall at 1k, and only a design covering every type at once can collect any of it.

### DIFF
--- a/tools/matrixark_local_adapter_context_node.py
+++ b/tools/matrixark_local_adapter_context_node.py
@@ -55,53 +55,73 @@ class _LocalAdapterContextNodeMixin:
         return self.default_session_node_path(envelope.get("scope", {}))
 
     def _existing_node_embedding_refs(self, current_model_ref: str) -> set[int]:
-        """Node hashes that already have a usable `context_node` embedding for this model.
+        """Node hashes that already have a usable embedding for this model.
 
-        Walking the whole log is fine on the JSONL backend, where `read_all()` is an in-memory
-        list. It is NOT fine on a native backend, where `read_all()` is a full record-log read
-        shipped over the proxy and re-run through the serving pipeline -- and
-        `ensure_context_node_path` runs three times per ingest. The native adapter overrides this
-        with a keyed lookup; see `MatrixArkTemporalStoreDirectAdapter._existing_node_embedding_refs`.
+        Answered from an index the adapter keeps, not by walking the store. This runs on every
+        ingest -- `ensure_context_node_path` is called several times per one -- and was one of the
+        remaining reads of the whole record set per write.
+
+        The index is built from one read the first time it is asked, folded forward on every
+        append, and dropped whenever the read cache is, so it cannot outlive the view it came from.
         """
-        refs: set[int] = set()
-        for record in self.read_all():
-            record_type = str(record.get("record_type") or "")
-            # A node embedding now lives ON the node: fold_embedding_records moves the vector onto
-            # the owner and drops the separate context_embedding row. Looking only for that row
-            # found nothing in any log written since, so every node was reported un-embedded and
-            # re-embedded on every ingest -- 60 embeddings for 3 distinct nodes over 20 ingests.
-            if record_type == "context_node":
-                meta = record.get("embedding_meta")
-                if not isinstance(meta, dict):
-                    continue
-                if str(meta.get("model_ref") or "") != current_model_ref:
-                    continue
-                if not record.get("vector") and not record_vector(record):
-                    continue
-                try:
-                    refs.add(int(record.get("node_hash")))
-                except (TypeError, ValueError):
-                    pass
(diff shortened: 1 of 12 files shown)


bjmeetsfo:
Landed. The scale measurement finished after the merge, so recording it here.

Main against this branch at **500 skills**, interleaved in one machine window so drift hits both
arms, each arm ingesting its own store:

| | main | after |
|---|---|---|
| wall | 2,883 s (48 min) | **67 s** |
| per skill | 5,766.5 ms | **133.2 ms** |
| the last 10 skills | 17,165 ms each | **231 ms each** |
| records held | 14,682 | 8,052 |
| read cache | 51.01 MB | **19.17 MB** |
| `read_all` calls | 25,502 | **507** |
| compactions | 26,006 | **501** |
| records visited by compaction | 187,942,313 | **2,031,909** |

43x on wall, and the shape changed rather than the constant: the last ten skills used to cost 99x
the first ten, and now cost 11x.

Two notes on the numbers themselves, since several readings today were wrong before they were
right:

- This is **one** paired repetition. A second would have cost another 48 minutes of main-arm time
  to re-confirm a 43x difference, so it was not run. Every figure above comes from arms that ran in the
  same window; nothing here is compared across time, which is what made three earlier readings in
  this work misleading — one reported a 30% improvement as a 15% regression.
- The ratchet is red on this branch and was red before it. The 24 failing names are a subset of
  the 25 that fail across four recent main runs, so none are new here; the counts alone would have
  read as 24 regressions.

PR #679, opened 2026-09-02
Hold one zstd compressor per thread instead of building one per record

bjmeetsfo:
The page-record encoder built a zstd compressor for every record, used it once and dropped it. It
now holds one per thread, which is what the decoder in the same file already does.

## What it costs today

Sweeping a write across nine value sizes, 8,000 writes per arm after a warm-up, counted rather
than timed:

```
  value size   allocations/write   allocated B/write
          64              130.7                7182
         192              131.5                7714
         256              131.5               40249   <- +32,535 B
         512              131.5               40698
        4096              131.5               47866
```

Flat at 2.0 B allocated per payload byte on both sides, and a **step between 192 and 256 bytes
worth 32,535 B per write**. 256 is `PAGE_RECORD_COMPRESSION_MIN_BYTES`, so the step is the
compressor being constructed. A soak corpus averages 1,244 B per record, so essentially every real
write pays it.

## After

```
  value size    before      after     change
       256 B    40,249      7,721     -80.8%
       512 B    40,698      8,490     -79.1%
     1,024 B    41,722     10,028     -76.0%
     4,096 B    47,866     19,255     -59.8%
  allocations     131.5      130.5     the one predicted
```

About **75% less allocated per write** at the corpus average, and the probe that found the step now
reports `No step: the cost is smooth across the range`.

## This changes bytes on disk, so it is pinned both ways

The bulk API frames a stream without the content-size header the streaming call writes, so stored
output is a little smaller — 698 B against 789 B for a 1,244-byte record. Both are ordinary zstd
frames.

`page_records_written_by_either_encoder_decode` pins that a record written by **either** encoder
decodes through **both** the bulk path and the streaming fallback, so a record written by either
build reads on either. It also asserts the two encodings differ, so if they ever converge the test
says it has stopped proving anything rather than passing hollowly.

## Why thread-local, and why the level is set per call

A compressor is `&mut` to compress, so one shared instance would serialise every writer behind a
lock on a path that is otherwise concurrent — the same reasoning already written down for the
decompressor directly below it.

`compression_level` comes from options and can vary per call, so it is set on the held compressor
each time. That is a parameter update, not a rebuild.

## Verification

- `cargo test -p temporalstore-rust --lib block_store::` — 60 passed
- `cargo test -p temporalstore-rust --lib compress` — 11 passed
- full `--lib` run, `--test-threads=1`, compared against the same run on the base commit

Two files, no deletions. The measurement probes behind the numbers above are separate, in #667.

### DIFF
--- a/crates/temporalstore-rust/src/block_store.rs
+++ b/crates/temporalstore-rust/src/block_store.rs
@@ -2758,6 +2758,49 @@ mod tests {
         assert_eq!(logical, b"bcde");
     }
 
+        /// A record written by the streaming encoder must still decode, and vice versa.
+    ///
+    /// The page-record encoder used to build a zstd compressor per call; it now holds one per
+    /// thread, which removed about 80% of what a write allocates. The two APIs frame a stream
+    /// differently, so the STORED BYTES changed -- and a stored-byte change is only safe if a
+    /// record written by either build reads on either.
+    ///
+    /// This pins that both ways round, which a round-trip test through one encoder cannot: it
+    /// would pass just as happily if the format had shifted under it.
+    #[test]
+    fn page_records_written_by_either_encoder_decode() {
+        let payload: Vec<u8> = (0..4096u32).map(|i| ((i * 7 + (i >> 3)) % 251) as u8).collect();
+        let level = 3;
+
+        // What the old encoder produced, byte for byte.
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
The measurements behind this are in #667 — `what_one_write_spends` locates the step and `what_rebuilding_the_compressor_costs` sizes the saving. Both are ignored and feature-gated, so they can land or not independently of this change.

PR #682, opened 2026-09-03
Key the retrieval caches by a scope that two tenants cannot share

bjmeetsfo:
## What

`canonical_scope_key` returns `""` for a scope carrying neither `scope_key` nor `tenant_hash` —
which is exactly the shape of the documented public scope, `{tenant_id, user_id, session_id}`. The
retrieval-records cache and the context-pack cache both key on it, so with a raw scope every
tenant shares one entry, and the second to ask a question is served the first one's answer.

Reproduced on one adapter with two tenants. Tenant B asked what its own pet was called and was
handed tenant A's:

```
family_profile = is called Mochi
```

| one adapter, tenant A asks first | A's content | B's own content |
|---|---|---|
| B asks second | **served to B** | missing |
| pack cache cleared in between | still served | missing |
| records cache cleared in between | still served | missing |
| both cleared | not served | served |

Clearing *either* cache alone did not help, because both keys collapse the same way — which is
what made this look like a selection problem rather than a key problem.

## Reachability

**The MCP entry point normalises the scope before this point, and through it tenant B gets its own
content.** No production caller invokes `adapter.retrieve()` with a raw scope; the only callers are
tests. So this is a latent hazard, not a live leak, and it is filed as one.

It is still worth closing. Nothing at the cache said the normalisation was load-bearing, and a
caller that skipped it got cross-tenant answers with no error to notice.

## The change

`cache_scope_key` falls back to the raw identifiers instead of to `""`, and returns a tuple so the
canonical and raw cases cannot collide with each other. `canonical_scope_key` is left alone: it has
58 callers, and `""` is a reasonable answer to "what is this scope's canonical name" — just not to
"what should this cache be keyed by".

It is added to **both** identity modules. There are two, both define `canonical_scope_key`, and
the live path resolves `matrixark_mcp_core_identity` — not the one an import search finds first. A
test asserts the two agree rather than sharing a symbol, which is how this codebase already treats
the pair.

## Checks

- The regression test fails without the fix, with the leaked row printed in the assertion, and
  passes with it.
- A control asserts tenant A gets its own content first, so "B was not served A's" cannot pass
  vacuously by serving nothing.
- `test_matrixark_retrieve*`, `test_matrixark_mem0*`, `test_matrixark_skill*`,
  `test_attachment_resource_policy`, `test_matrixark_index*`, `test_engine_purge_on_delete`,
  `test_delete_before_extract`, `test_matrixark_embedding_status` all pass.
- 5 files, 168 insertions, 2 deletions, no file deletions.

### DIFF
--- a/tools/matrixark_local_adapter_retrieval.py
+++ b/tools/matrixark_local_adapter_retrieval.py
@@ -206,7 +206,9 @@ class _LocalAdapterRetrievalMixin:
         """
 
         allowed_types = record_types or RETRIEVAL_HOT_RECORD_TYPES
-        scope_key = canonical_scope_key(scope)
+        # cache_scope_key, not canonical_scope_key: the canonical one is "" for an unnormalised
+        # scope, and every tenant would then share one cache entry.
+        scope_key = cache_scope_key(scope)
         secondary_key = tuple(sorted(tuple(sorted(group)) for group in (secondary_index_groups or [])))
         selected_key = tuple(sorted(int(item) for item in (selected_node_hashes or set())))
         cache_key = (
--- a/tools/matrixark_local_adapter_retrieve.py
+++ b/tools/matrixark_local_adapter_retrieve.py
@@ -633,7 +633,9 @@ class _LocalAdapterRetrieveMixin:
         )
         pack_cache_key = (
             self._retrieval_records_cache_generation,
-            canonical_scope_key(scope),
+            # See cache_scope_key: canonical_scope_key is "" for an unnormalised scope, which
+            # would put every tenant on one cache entry.
+            cache_scope_key(scope),
             query,
             question_type,
             retrieval_session_scope,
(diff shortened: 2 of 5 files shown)


bjmeetsfo:
CI note, so the red check is not read as this change breaking something.

The ratchet fails here and it fails on main. Comparing failing test NAMES rather than counts:

| | distinct failing names |
|---|---|
| this branch | 35 |
| union of four recent main runs | 35 |
| **failing here and on no main run** | **0** |

Everything else is green: oss-readiness, secret scan, SPDX headers, auto-approve. The cargo check
is a compile job and this change is Python only.

The real gate for this one is the regression test, which fails on main with the leaked row printed
in the assertion:

```
AssertionError: 'Mochi' unexpectedly found in
'[{"ref_type": "entity", "text": "family_profile: family_profile = is called Mochi", ...}]'
: the second tenant was served the first tenant's content
```

PR #687, opened 2026-09-03
A posting keeps its node identities, and is not rebuilt when it is already folded

bjmeetsfo:
Three commits on one thread: a correctness fix, the optimisation it unblocked, and a third the
profiling then exposed. Against main, paired and interleaved at 250 skills:

| | main | branch |
|---|---|---|
| per skill | 82.4 ms | **52.1 ms (−37%)** |
| the last 10 skills | 134.4 ms | **95.5 ms (−29%)** |
| read cache | 9.79 MB | 9.79 MB (unchanged) |
| bytes per record | 2,430 | 2,430 (unchanged) |

## 1. A posting keeps the node identities it accumulated

The fold accumulated node and batch identities from the **singular** `node_hash` /
`batch_id_hash`. Its own emission writes the singular only when a bucket holds exactly one and
pops it otherwise — so a posting for a bucket with several nodes carries `node_hashes` and no
`node_hash` at all. Folding it again found nothing to carry over and dropped the list:

```
after one fold    node_hashes=[200, 201, 202, 203]
after two folds   node_hashes=None
```

It has not been losing data only while the skip-when-already-folded path returns such a posting
untouched. **That path stops firing the moment a fresh row is appended, which is every ingest.**

A live cache of 92 postings, re-folded once as an ingest would:

| node identities per posting | before | after |
|---|---|---|
| **none** | **9** | **0** |
| one | 83 | 83 |
| two / three / four | 0 | 4 / 4 / 1 |

Nine postings had lost every identity they held. All nine recovered, every other field
byte-identical.

## 2. Adopt a posting that is already folded instead of rebuilding it

When a fresh row arrives the skip path stops applying, and every posting in the cache is taken
apart and rebuilt to accommodate a handful of new ones. A folded posting already holds its
bucket's accumulated `ref_hashes`, `node_hashes` and lineage, so adopt it as that bucket's
accumulator and emit it untouched when nothing joins the bucket.

On the shape a compaction sees each ingest — 2,136 cached rows plus 37 fresh — the fold went
**31.618 → 16.234 ms, −49%**.

## 3. Do not re-share the rows compaction handed straight back

Profiling after (2) put value sharing at a quarter of ingest wall. It walks every field of every
record it is given, and compaction returns the SAME objects for every row it did not rebuild —
which were shared when they entered the cache. Over 150 ingests that site walked **159,885 rows to
change 1,924**: 96.1% of the walk rediscovering there was nothing to do.

The cache now says which objects it already held. Membership is one hash against a set built
without touching a field, and the ids cannot be recycled underneath it — the set is built from a
list the same frame still holds.

**1.34 s → 0.19 s at that site**, 18.5% → 3.1% of wall.

## Why they belong together

Adoption and rebuild have to agree, and could not while the rebuild was losing identities: the
first version diverged on 44 of 364 folds, on `node_hashes` and the `index_hash` derived from it.
The divergence was the rebuild dropping them, not the adoption inventing them. (3) is only visible
once (2) has removed the cost that was hiding it.

## Checks

- Adoption compared against the rebuild loaded as a **separate module**, on every fold of a real
  ingest: **364 calls, 364 identical**, plus inputs built to hit every fallback. Comparing against
  the same function with its skip path disabled would have compared adoption to adoption and shown
  clean — that harness was rewritten after it did exactly that.
- The identity test fails on main with the loss in the assertion:
  `AssertionError: [200, 201] != None : folding a 2-node posting again lost its identities`
- Served records unchanged: ordered and sorted digests match on the same log, each arm given its
  own copy of the store.
- Sharing still rewrites the same 1,924 rows; warm cache 2,480 B/record either way, cold read
  2,951 B/record either way.
- 3 files, 219 insertions, 9 deletions, no file deletions. All adapter, index, posting, skill,
  attachment, cache-key, purge, delete and embedding-status suites pass.

## What this does not do

It does not change the shape. Compaction still walks every row on every read, so the work still
grows with the square of the corpus — 132,114 rows visited at 125 skills, 2,041,503 at 500,
unchanged. This makes each walk cheaper, not fewer.

Bounding the walk is the separate piece, and its measurements are recorded: the tail is 3.4% of
what a compaction walks, but 98.3% of compactions collide with the head on posting bucket, state
key and value key at once, so it needs merge logic on all three rather than a
fall-back-on-collision shortcut.

### DIFF
--- a/tools/matrixark_mcp_indexing.py
+++ b/tools/matrixark_mcp_indexing.py
@@ -495,6 +495,9 @@ def compact_context_index_postings(records: list[Json]) -> list[Json]:
         "source_profile_memory_classes",
         "source_profile_memory_kinds",
     ]
+    #: bucket key -> the already-folded row its state was adopted from, while nothing else has
+    #: joined that bucket. Such a bucket is emitted unchanged.
+    adopted: dict[tuple[Any, ...], Json] = {}
     grouped: dict[tuple[Any, ...], Json] = {}
     grouped_scalar_values: dict[tuple[Any, ...], dict[str, set[str]]] = {}
     grouped_list_values: dict[tuple[Any, ...], dict[str, list[Any]]] = {}
@@ -519,6 +522,21 @@ def compact_context_index_postings(records: list[Json]) -> list[Json]:
             str(record.get("ref_type") or ""),
             bucket_ms,
         )
+        if key not in grouped and is_fold_output(
+            record, POSTING_POLICY_BUCKETED,
+            ("index_hash", "storage_record_kind", "storage_part")
+        ):
+            # This row IS the fold's own output for its bucket: its ref_hashes, node_hashes and
+            # lineage fields are already the accumulated state. Adopt it whole rather than
+            # rebuilding it field by field, and emit it untouched if nothing joins the bucket.
+            # On a 2,123-row cache with 37 fresh rows the rebuild was 23.755 ms against 3.255 ms
+            # for the same list with nothing new in it.
+            grouped[key] = dict(record)
+            grouped_scalar_values[key] = {field: set() for field in scalar_lineage_fields}
+            grouped_list_values[key] = {field: [] for field in list_lineage_fields}
+            adopted[key] = record
+            order.append(key)
+            continue
         if key not in grouped:
             grouped[key] = {
                 "record_type": "context_index",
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
Merged as `a2361c4b7`. Verified on merged main rather than taken from the PR state, since a label
says a merge happened and not that the content arrived:

- all three files byte-identical between the branch and main
- `_identity_values`, the adoption bookkeeping and the `already` skip all present in main
- the tests added by this work, run against main's own extracted tree: **46 of 46 pass** across
  `posting_keeps_its_node_identities`, `posting_fold_skips_folded_rows`,
  `repeated_values_are_shared`, `log_keys_are_shared`, `cache_key_separates_tenants`,
  `debug_records_honour_their_gate`, `skill_sections_stored_once`
- a 200-skill ingest on merged main: 39.1 ms/skill, 2,448 B/record

The ratchet was red here and is red on main: 20 failing names on this branch, 20 across four
recent main runs, and **0 failing here that do not also fail there**.

PR #728, opened 2026-09-03
Bound socket reads by the call deadline, not a two-second constant

bjmeetsfo:
`_call_socket_json` armed the warm-proxy socket with `min(2.0, request_timeout_ms / 1000)` and
never re-armed it, so any answer slower than two seconds raised `TimeoutError` no matter what
`--request-timeout-ms` asked for. `_call_json` re-raises rather than falling back to the
process-lane path, so the call failed outright.

For an agent hook that is not a corner case. A ContextPack assembled over a large store takes
longer than two seconds, so retrieval failed on every turn, fail-open swallowed it, and the only
trace was `TimeoutError: timed out`. Measured on a live box: plain socket reads answered in 24 ms
while every pack died at the ceiling.

## What changed

- The read timeout is re-armed before each read from what is **left** of the deadline the method
  already computed, so `--request-timeout-ms` means what it says.
- The connect keeps a short ceiling (`SOCKET_CONNECT_TIMEOUT_CEILING_S`), so a stale socket file
  still fails fast instead of spending the whole budget.
- A bounded wait now raises `MatrixArkError ... after Ns` naming the socket and the budget,
  instead of a bare `TimeoutError`.
- `matrixark_rust_proxy_daemon.ping()` carried the same two-second constant. A health check that
  reads a busy daemon as a dead one makes every caller start its own, and a second daemon unlinks
  the first one's socket on bind, so the loser spawns its own proxy and reloads the store per
  call. It is now `MATRIXARK_RUST_PROXY_PING_TIMEOUT_MS`, default 10 s, floored at 0.5 s. A ping
  measured 2.4 s on a busy box with the daemon perfectly healthy.

## Tests

`tools/test_matrixark_rust_proxy_socket_deadline.py` drives real Unix sockets that answer late, so
it exercises the timeout the socket is actually armed with rather than asserting on source text:

- a 3 s answer succeeds when the request timeout allows it, and the test asserts it waited at
  least 3 s, so it cannot pass vacuously;
- a server that never answers still fails at the budget, with the elapsed time bounded on both
  sides, so a connect that failed outright cannot pass as a well-behaved deadline;
- connect keeps its ceiling under a 600 s request timeout;
- the ping waits for a busy daemon, and its timeout is configurable and bounded.

4 of the 5 fail against this branch's base; all 5 pass with the change.

### DIFF
--- a/tools/matrixark_mcp_temporal_adapters.py
+++ b/tools/matrixark_mcp_temporal_adapters.py
@@ -3772,15 +3772,39 @@ class MatrixArkRustProxyClient:
         self._record_call_metrics(op, kwargs, response, elapsed_ms, failed=False, lane=group, wait_ms=wait_ms)
         return response
 
+    # Connecting to a live daemon socket returns immediately, so the connect keeps a
+    # short ceiling: a stale socket file should fail fast rather than spend the call's
+    # whole budget. Reads are a different question -- see `_call_socket_json`.
+    SOCKET_CONNECT_TIMEOUT_CEILING_S = 2.0
+
     def _call_socket_json(self, op: str, payload: str) -> Json:
-        deadline = time.monotonic() + max(2.0, self.request_timeout_ms / 1000.0 + 2.0)
+        # The read timeout tracks what is LEFT of the deadline. It used to be a constant
+        # `min(2.0, ...)`, which no `--request-timeout-ms` could raise, so any response
+        # slower than two seconds became a timeout -- and since `_call_json` re-raises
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The warm-proxy socket read was capped at a constant 2s that --request-timeout-ms could not raise, so any slower answer failed the call outright (every context pack over a large store); the read now tracks the call deadline, the daemon health check gets MATRIXARK_RUST_PROXY_PING_TIMEOUT_MS, and 5 new tests drive real sockets that answer late - 4 of them fail against the base.

PR #734, opened 2026-09-03
Keep sharing what a large corpus repeats, on both read paths

bjmeetsfo:
Two defects found while measuring a corpus of 1,000 files at 1 MB each — 100,105 records, a 217 MB
snapshot. Together they cost 42.8% of the read cache.

## 1. The durable snapshot decoded differently from the log

`read_all()` has two paths. The JSONL replay decodes each line through `loads_with_interned_keys`,
which interns key names and shares repeated string values. The snapshot decoded with a bare
`json.load`. Same store, same records, different cost:

| path | B/record |
|---|---|
| JSONL replay (cold) | 3,195 |
| durable snapshot (every warm read) | 4,233 |

The snapshot is the path that serves. The delta tail three lines below it already used the hook.

Keys were never the problem here — the snapshot is a single `json.load` and CPython memoises key
strings within one call — so only *values* were duplicated, on one path, which is what made it easy
to miss. The fix passes `object_pairs_hook=_interned_pairs`, so both paths now answer the same.

## 2. The shared-value table stopped at 4,096 entries

`_SHARED_VALUE_TABLE_LIMIT = 4096`, documented as "a runaway guard, not a working limit" because
the corpus it was calibrated on held 11 distinct values in its busiest field. This corpus holds
29,657 distinct vectors: the table filled, and every value after the 4,096th was handed back
unshared. Nothing reported it — sharing simply stopped. Cost: 66.8 MB of duplicate vectors.

The ceiling is replaced by a per-field hit-rate test, the same shape as the string table's, so the
guard now falls on fields that do not repeat rather than on whichever field filled the table first.

Two things are deliberately different from the string version:

- **A higher bar.** A string *is* its own table key, so a hit at any rate is free. A container's
  key is a second copy of its spine, so sharing only pays above `key / (value + key)` — ~0.31 here
  (a shared vector saves ~1,015 B against a ~450 B key). The bar is 0.35: above break-even, and
  deliberately *below* the 0.50 that perfect 2x duplication produces, since a chunk body is stored
  once as a `skill_section` and once as a `resource_chunk` and lands exactly on 0.50.
- **The verdict is not permanent.** It is reached on the first 512 lookups, and a corpus can put a
  field's repeats after them — a store written as one pass of distinct values followed by a second
  pass of duplicates would show a 0.00 hit rate throughout the warmup and be written off for the
  life of the process. Abandoned fields are judged again after 8,192 lookups.

## Measured

100,105 records, one store copy per arm (reading a store rewrites its snapshot, so the arms cannot
share one):

| | cache | B/record | peak RSS |
|---|---|---|---|
| before | 0.424 GB | 4,233 | 1,229 MB |
| after | 0.242 GB | 2,420 | 804 MB |

**−42.8% per record, −425 MB resident.** RSS falls further than the cache does because duplicate
strings inflate allocator arenas as well. Both read paths now report the same size.

What is left: `metadata`, 99,320 objects for 49,641 distinct values, 6.7% of the cache. It is a
nested dict, so it never qualifies for sharing at all — that needs its own change and is not
attempted here.

## Tests

Both defects are pinned by tests that fail on `main` on their own assertions, not on a missing
attribute:

- the snapshot test reports 6 objects for a value 6 records share;
- the ceiling test reports 7,904 objects for 6,000 distinct values — 4,096 shared plus 1,904 pairs
  that never were, which is the ceiling arithmetic exactly.

The suite was run under `unittest discover` on three trees to attribute the result: main, the code
change alone, and the code change with the new tests. The code change alone fails no test that main
does not already fail.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -748,11 +748,70 @@ def _shared_interned_value(value: Any) -> Any:
     return _copy_interned_value(value)
 
 
-#: Beyond this many distinct values a field is not repetitive enough to be worth a table entry,
-#: and the table itself would start to cost more than it saves. Measured on an attachment corpus
-#: the busiest field held 11 distinct values, so this is a runaway guard, not a working limit.
 SHARE_REPEATED_VALUES = bool_env("MATRIXARK_SHARE_REPEATED_VALUES", True)
-_SHARED_VALUE_TABLE_LIMIT = 4096
+
+#: A last-resort ceiling on the shared table, not a working limit.
+#:
+#: It used to be 4096, chosen when "the busiest field held 11 distinct values" on a 60-attachment
+#: corpus -- a runaway guard that could not plausibly bind. At 100,105 records it binds hard: the
+#: corpus holds 29,657 distinct ``vector`` values, so the table filled and every value after the
+#: 4,096th was handed back unshared. The measured cost of that one constant was 66.8 MB of
+#: duplicate vectors, 20.9% of a 320 MB cache, and nothing reported it -- the table simply stopped
+#: sharing.
+#:
+#: What keeps the table honest now is :data:`_CONTAINER_FIELD_MIN_HIT_RATE` below, which drops a
+#: field that is not actually repeating. A field still in the table is saving more than its
+#: entries cost, so growth here is self-justifying and the ceiling only has to stop something
+#: nobody foresaw.
+#:
+#: Worth knowing before raising it further: the table holds strong entries and never shrinks, so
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The durable snapshot decoded with a bare json.load while the log replay interned and shared, so one store cost 4,233 B/record served warm against 3,195 served cold; and the shared-value table's 4,096-entry ceiling silently stopped sharing on a corpus holding 29,657 distinct vectors. Read cache 4,233 -> 2,420 B/record, peak RSS 1,229 -> 804 MB, both read paths now agreeing.

PR #738, opened 2026-09-03
Refresh the read snapshot from the read that was served from cache

bjmeetsfo:
A write never rewrites the durable read snapshot — it continues the tail when it can,
and otherwise leaves the snapshot alone on the stated grounds that *"the next read
refreshes it"*. Two comments in the write path say exactly that, and
`_write_durable_read_cache` returns early under `tail_only` for the same reason.

The next read does not. The snapshot is only written on the branch that re-derives the
record set from the log — and once a cache is warm, no read reaches that branch again:
appends keep the process-wide cache current, so every later read is served from it and
returns before the write.

The two halves then wait on each other. Once the snapshot is behind, its recorded
signature no longer matches the log, so `_load_durable_read_cache` refuses it — and the
write path only continues a tail for a snapshot it could load.

Traced over twelve batches through two adapters on one log:

| | before | after |
| --- | --- | --- |
| snapshot writes | 1 | 12 |
| records in the snapshot | 10 of 120 | 120 of 120 |
| full base rewrites | 1 | **1** |
| cold read served from | `jsonl` | `durable` |

```
batch 0  read_all -> jsonl     head_count=10  live_size=5356   loadable=True
batch 1  read_all -> process   head_count=10  live_size=9366   loadable=False
batch 2  read_all -> process   head_count=10  live_size=13376  loadable=False
...
```

Every cold start re-derived the whole log, which is the cost the snapshot exists to remove.

**The fix.** A read served from cache refreshes the snapshot itself — the same thing the
jsonl branch already does with the records it installs. The eleven added writes are all
tail appends; the base is still written once.

**It does no work when nothing changed.** Reading the head alone answers whether the
snapshot is behind, so: ten idle reads on the same instance, ten on a second instance,
and five cold reads write nothing and leave the head's mtime untouched.

**No fallout.** Under `unittest discover`: 121 failing before, 118 after. It fixes the
cold-reader test and `test_background_worker_refreshes_dirty_nodes_and_embeddings`, and
adds no failure anywhere.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -4629,6 +4629,40 @@ class MatrixArkLocalAdapter(_LocalAdapterRetrieveMixin, _LocalAdapterIngestMixin
         except (FileNotFoundError, json.JSONDecodeError, OSError, TypeError, ValueError):
             return (0, 0)
 
+    def _durable_read_cache_signature(self) -> Json | None:
+        """The signature the head recorded, or None when there is none to trust."""
+        try:
+            with self._durable_read_cache_head_path().open("r", encoding="utf-8") as handle:
+                head = json.load(handle)
+            if head.get("cache_key") != self._cache_key_str():
+                return None
+            if head.get("schema_version") != LOCAL_DURABLE_READ_CACHE_SCHEMA_VERSION:
+                return None
+            return head.get("signature")
+        except (FileNotFoundError, json.JSONDecodeError, OSError, TypeError, ValueError):
+            return None
+
+    def _refresh_durable_read_cache_if_behind(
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Merged. @superhaiou for awareness — a write left the durable snapshot for "the next read" to refresh, but once a cache is warm every read is served from it and never reaches the branch that refreshes. The snapshot froze at 10 of 120 records and every cold start re-derived the log. Base rewrites stay at 1; idle reads write nothing.

bjmeetsfo:
Merged. @superhaiou for awareness — a write left the snapshot for the next read to refresh, but once a cache is warm no read reaches the branch that refreshes it, so the snapshot froze at 10 of 120 records and every cold start re-derived the log.

PR #739, opened 2026-09-03
Share a value whose children are shared, not only a flat one

bjmeetsfo:
The last duplication in the read cache, and the reason it survived every previous pass.

## What was left

Sharing keys a container by its contents, so a dict holding a container could not be keyed at all
and was returned unshared however often it repeated. That covered `metadata` exactly:

- every one of its values holds a container — `heading_path`, a list — so none qualified;
- its repetition histogram is `{2: 2000}`: **every distinct value appears exactly twice**, once as
  a `skill_section` and once as a `resource_chunk`;
- 99,320 objects for 49,641 distinct values, **13.4% of the read cache carried at double**.

Measured over 4,000 sampled values, 4,000 of 4,000 would qualify under the change below.

## The change

Once a dict's children have been shared, the dict itself can be keyed: a shared child is the one
object the table holds for its value, so the child's **identity** stands in for its contents and
the parent gets a cheap, exact key.

Identity is only sound here because the table holds its entries strongly and nothing evicts them,
so a shared child outlives every key naming it. A child whose field was abandoned, or that arrived
after the table hit its ceiling, is not one of the shared types and falls out of the check — which
is what keeps the identity honest rather than merely convenient.

## Tests

Five, and the one that matters fails on the base branch at **1,800 objects for 900 distinct
values**. The other four are invariants that hold on both sides and are worth stating:

- the nested child is itself shared (non-vacuity — the parent can only be keyed by a canonical
  child);
- content is unchanged by sharing;
- two values differing **only inside the nested list** stay separate objects, so the identity key
  distinguishes rather than merely collapsing;
- a dict holding something that is neither a scalar nor a shared container is passed through
  untouched.

## Measured

100,105 records, one store copy per arm, cold and warm read each:

| | cache | B/record |
|---|---|---|
| main | 0.242 GB | 2,420 |
| with this change | 0.231 GB | 2,305 |

**-4.8%**, cold and warm identical. Less than the 6.7% a full collapse would give, because some
metadata values hold a child whose own field did not earn a place in the table and so is not
canonical -- those parents are correctly passed through rather than keyed by an identity that
would not be stable.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -814,6 +814,9 @@ _SHARED_CONTAINERS_EARNED: set = set()
 _SHARED_CONTAINER_STATS: dict = {}
 #: Only a value whose every entry is one of these can be keyed by its contents.
 _SHAREABLE_SCALARS = frozenset({str, int, float, bool, type(None)})
+#: The two types :func:`_lookup_shared` hands back. An instance of either is canonical -- the one
+#: object the table holds for that value -- which is what makes its identity usable as a key.
+_SHARED_CONTAINER_TYPES = (_SharedInternedValue, _SharedInternedList)
 #: flat value -> the one object every record carrying it holds. Process-wide, so two adapters
 #: over the same store share as well.
 _SHARED_VALUE_TABLE: dict = {}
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Sharing keyed a container by its contents, so a dict holding a container was never shared however often it repeated - which was every metadata value in the corpus, each appearing exactly twice. Read cache 2,420 -> 2,305 B/record.

PR #740, opened 2026-09-03
Count an index posting in a layer, and ask the postings the question that is now theirs

bjmeetsfo:
The last two NEW failures the ratchet reports. Both are the same story: postings moved,
and the tests still ask where they used to be. One of them turned out to be a real hole.

## The inventory counted postings in no layer at all

`retrieval_memory_inventory` reported `context_indexes: 0` in **all three** layers while
twenty postings sat on the log. A posting carries neither `memory_scope` nor
`session_continuity` — it is a derived row pointing at one that does — so `is_profile`,
`is_session` and `is_shared` were all false and it was skipped entirely. The field was dead
everywhere, not just under `profile`.

The profile clause already reads `data_model` for exactly this reason
(`data_model == "context_profile_entity"`). The session layer now does the same for the
models its own postings are attributed to. On the test's log: **0 → 12**.

The profile layer still reports 0, and that is the honest answer — postings for a profile
entity are attributed to the batch commit that wrote them, not to the entity they point at.
The test asserts that now, instead of the number it used to see.

## The other test asked for postings that are no longer written

`drop_owner_derivable_postings` drops a posting whose owner carries a vector and can derive
the term — all twenty-one event-type and role postings in that batch. Its docstring says the
owner-carries-a-vector condition is *"the condition the retrieve prefilter's owner branch is
gated on"*, and that is exactly what happens:

| | |
| --- | --- |
| `index_postings_read` | 0 — nothing was stored to read |
| `secondary_embedding_matched_posting_count` | 1 — the owner branch matched |
| `broad_scan_used` | False |
| scanned → returned | 37 → 3 |
| the `tool_evidence` event asked for | returned |

So the prefilter still narrows and still answers; only the counter moved. The test now
asserts the answer and the narrowing rather than a counter measuring a mechanism that is no
longer used. I had this recorded as a genuine retrieval regression — it is not.

## Both keep a positive control

Each absence assertion sits next to a presence one — the batch's own postings, which derive
nothing and are still written — so neither can pass on an empty set.

**Mutation-tested**, every flip fails:

| mutation | result |
| --- | --- |
| `assertEqual(set(), event_index_names)` → `assertNotEqual` | killed |
| positive control `assertIn` → `assertNotIn` | killed |
| `index_postings_read == 0` → `!= 0` | killed |
| owner-branch match `>= 1` → `>= 99` | killed |
| `broad_scan_used` False → True | killed |
| narrowed (`returned < scanned`) → reversed | killed |
| profile `== 0` → `!= 0` | killed |
| session `>= 1` → `>= 99` | killed |
| **the classifier change reverted** | killed |

## Noticed, not changed here

`event_indexes_written` counts postings as they are built, before
`drop_owner_derivable_postings` removes them — it reported 21 for a batch that wrote none.
Left alone: it is a separate call about what that number should mean.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -275,7 +275,16 @@ def retrieval_memory_inventory(records: list[Json], retrieval_scope: Json) -> Js
                 and session_continuity == "cross_session"
             )
         )
-        is_session = memory_scope in {"session", "session_memory"} or session_continuity == "same_session"
+        # An index posting carries neither a memory scope nor a session continuity -- it is a
+        # derived row pointing at one -- so on those two tests alone it belonged to no layer and
+        # was skipped: twenty postings on a log, context_indexes 0 in all three layers. The
+        # profile clause above already reads data_model for exactly this reason; the session
+        # layer needs the same, for the models the session's own postings are attributed to.
+        is_session = (
+            memory_scope in {"session", "session_memory"}
+            or session_continuity == "same_session"
+            or data_model in {"context_event", "context_batch_commit", "context_segment"}
+        )
 
         if is_shared:
             # A folded owner counts as the embedding it carries -- the separate record it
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
Merged. @superhaiou for awareness — the memory inventory counted index postings in no layer at all (0 in all three while 20 sat on the log), because a posting carries neither a memory scope nor a session continuity. And the prefilter counter that read 0 was the owner branch answering instead of a stored posting, not a regression.

PR #741, opened 2026-09-03
Carry the two largest constants in the intern bundle

bjmeetsfo:
Two fields are ~11% of the durable log and each carries exactly one distinct value.

## What was on the wire

`INTERN_METADATA_FIELDS` moves a repeated `{field: value}` set into a sidecar written once per
distinct token, leaving a short hash on the row. It covers 45 fields. It did not cover the two
biggest constants.

Sampled over 6,000 data rows of a 1 MB-per-document corpus:

| field | share of the wire | rows | distinct values | interned |
|---|---|---|---|---|
| text | 39.1% | 5,936 | 2,969 | no |
| embedding_meta | 12.3% | 5,941 | 5,940 | no |
| vector | 12.0% | 5,944 | 2,594 | no |
| **access_scope** | **9.6%** | 5,940 | **1** | **no** |
| **deployment_scope** | **1.2%** | 5,940 | **1** | **no** |

`access_scope` is 216 B on 99% of rows. At ~1,949 records per 1 MB document that is roughly 400 KB
of one repeated value per document.

## Why these two are safe

The exclusions in that list are not stylistic — `scope_key`, `record_type` and the model-registry
identity fields are named there because some path reads them off the **unexpanded** log, where an
interned value is invisible. Eliding `scope_key` would make scope-level tombstones miss a purge.

Every raw reader was checked. Six functions iterate log lines; three expand first
(`_load_durable_read_cache`, `_read_all_compacted`, `_read_raw_records`) and the three that do not
— `_seed_intern_tokens_locked`, `_seed_model_registry_seen_locked`, `purge_tombstones` — read only
record_type, scope_key and the model identity fields. Neither field here is read by any unexpanded
path. The Rust side reads `access_scope` (`scope_scoring.rs`) but no Rust file reads
`events.jsonl`; it arrives through the API, past expansion.

They also cost nothing in sidecars, which is the number to watch: fields share ONE bundle, so a
field that varies multiplies tokens instead of removing bytes. At one distinct value each these add
no combinations at all.

## Measured

Ingesting 8 documents of 1 MB, same corpus, same records:

| | append log |
|---|---|
| main | 38.0 MB |
| with this change | **34.3 MB** |

**-9.7%**, against the 10.8% the sample predicted. The append log is deterministic here — main
produced 38.0 MB on repeated runs.

## Tests

Five, including one end to end through the adapter rather than the codec alone: it asserts that no
row carries `access_scope` inline on the durable log **and** that a served read still carries it,
which is the whole contract. One more pins the sidecar count at 1 for 200 rows, so a future field
that varies cannot be added here without the test noticing.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -462,6 +462,23 @@ INTERN_METADATA_FIELDS = (
     # identity fields (model_kind, model_ref, model_name, model_hash, provider, execution_mode --
     # _seed_model_registry_seen_locked reads them off the unexpanded log), and the bundle's own
     # token key, which cannot be part of what it names.
+    # Measured on a corpus of 1 MB documents: `access_scope` is 9.6% of the durable log and
+    # `deployment_scope` a further 1.2%, and BOTH carry exactly one distinct value across the
+    # whole store -- 5,940 rows, one value each. They are the two largest constants on the wire.
+    #
+    # They are safe here on the criterion this list already uses. The exclusions above are fields
+    # some path reads off the UNEXPANDED log, where an interned value is invisible. Every raw
+    # reader was checked: of the six functions that iterate log lines, three expand first
+    # (`_load_durable_read_cache`, `_read_all_compacted`, `_read_raw_records`) and the three that
+    # do not -- `_seed_intern_tokens_locked`, `_seed_model_registry_seen_locked` and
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. access_scope was 9.6% of the durable log and deployment_scope a further 1.2%, each carrying exactly one distinct value across the store, and neither was in the intern bundle. Append log 38.0 -> 34.3 MB on an 8-document corpus.

PR #742, opened 2026-09-03
Offer the settings that decide what the store keeps

bjmeetsfo:
The portal offers 80-odd settings and none of them says how much the store keeps.

## What is missing

The local event log rotates into fixed-size shards and retains a bounded number of them. Those two
numbers are the deployment's disk ceiling — and, less obviously, a **retention policy**: records in
a dropped shard are gone from the store. At the defaults, 64 MB per shard and 4 retained behind the
live one, that is roughly **336 MB kept and the oldest records discarded past it**.

Neither number was offered anywhere on the portal, so an operator could neither see the ceiling nor
raise it.

They also bound ingest cost, which is the part that is easy to miss. The retained window is what a
read holds and what compaction walks, so per-document ingest time rises with it until rotation
begins. Measured on 1 MB documents, drained, fresh store each time:

| documents | s/document |
|---|---|
| 15 | 2.684 |
| 30 | 4.833 |
| 60 | 7.346 |

The curve flattens after that, and it flattens because rotation has started discarding — so the
retention setting is also, in effect, the throughput setting. An operator who cannot see it cannot
reason about either.

## The change

Three settings added to the `ingestion` group, which the setup page already renders:

| key | variable |
|---|---|
| `ingestion.local_log_max_bytes` | `MATRIXARK_LOCAL_JSONL_MAX_BYTES` |
| `ingestion.local_log_retention_count` | `MATRIXARK_LOCAL_JSONL_RETENTION_COUNT` |
| `ingestion.durable_read_cache` | `MATRIXARK_LOCAL_DURABLE_READ_CACHE_ENABLED` |

All three are labelled `restart`, which the existing config audit independently confirms is
correct: its source scan finds each one read exactly once, at import, in
`matrixark_mcp_local_adapter.py`. That matters in both directions — the audit also fails a setting
labelled `restart` that has no import-time reader, so the label is checked, not merely asserted.

They go in `ingestion` rather than `storage_engine` deliberately: every setting in that group is a
`TS_*` engine knob, guarded by a test that each is actually read by the engine. These are Python
knobs and belong beside the other `MATRIXARK_*` ingestion settings.

## Tests

Five, all failing on `main` because the settings do not exist. Beyond presence they pin the things
that make an offered setting trustworthy rather than decorative: that the group has metadata so the
page can render it, that each carries help text, that the label is `restart`, and that the default
shown **equals the value the store actually uses** — a portal that displays a default the code does
not use is worse than showing nothing.

### DIFF
--- a/tools/matrixark_gateway_config.py
+++ b/tools/matrixark_gateway_config.py
@@ -473,6 +473,31 @@ SETTINGS: List[Setting] = [
             "Drainer interval (ms)", "int", "", "restart",
             "How often the drainer wakes to encode what ingest left behind."),
 
+    # ---- what the local store keeps -------------------------------------------------------------
+    # These two together are the deployment's disk ceiling and, less obviously, its ingest cost:
+    # the retained window is what a read has to hold and what compaction has to walk, so per
+    # document ingest time rises with it until rotation starts discarding. Measured on 1 MB
+    # documents, ingest cost per document climbed 2.68 s -> 4.83 s -> 7.35 s at 15, 30 and 60
+    # documents and then flattened, and it flattened because rotation had begun. An operator
+    # sizing a box needs to see both numbers, and until now neither was offered.
+    Setting("ingestion.local_log_max_bytes", "ingestion", "MATRIXARK_LOCAL_JSONL_MAX_BYTES",
+            "Event log shard size (bytes)", "int", "67108864", "restart",
+            "How large one event-log shard grows before the store rotates to a new one. This "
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The event log rotates into fixed-size shards and retains a bounded number of them - roughly 336 MB at the defaults, with older records discarded past it - and neither number was offered on the portal. They also bound ingest cost, since the retained window is what a read holds and compaction walks.

PR #743, opened 2026-09-03
Offer the sharing knob, with the trade it makes

bjmeetsfo:
The sharpest memory-for-latency trade in the local store, and it was not offered at all.

## The trade

Value sharing gives every record carrying the same value one shared object instead of a private
copy. Measured on the 100,105-record store, warm read, three alternating runs per arm on separate
store copies:

| | warm read | B/record | resident |
|---|---|---|---|
| sharing on | 8.00 s | 2,305 | 761 MB |
| sharing off | 4.58 s | 4,173 | 1,041 MB |

So it returns **~45% of the read cache and 280 MB of resident memory** for about **3.4 seconds** on
the read that loads the cache. Which way that should go is a property of the deployment — a
long-lived server pays the seconds once and keeps the memory; a short-lived process may prefer the
opposite — and there is no way to choose it from the portal today, though the variable has always
existed.

Worth noting what is *not* the cost: interning keys and values during the decode is free. Measured
on the same store, the current code with sharing off reads in 4.58 s against 5.26 s for the build
before any of today's changes, so the decode path got slightly faster. The 3.4 s is entirely the
container sharing, which is what makes it a real choice rather than a regression to hide.

## The change

One setting, in the `ingestion` group beside the other local-store knobs added recently:

| key | variable |
|---|---|
| `ingestion.share_repeated_values` | `MATRIXARK_SHARE_REPEATED_VALUES` |

Labelled `restart`, which the existing config audit confirms independently: its source scan finds
the variable read once, at import, in `matrixark_mcp_local_adapter.py`.

## Tests

The existing portal-settings test grows to cover it, and one new case asserts the help text carries
**both** sides of the trade — the memory it buys and the seconds it costs. A knob offered with only
one half invites the operator to guess, and this is the knob where guessing is most expensive.

### DIFF
--- a/tools/matrixark_gateway_config.py
+++ b/tools/matrixark_gateway_config.py
@@ -490,6 +490,20 @@ SETTINGS: List[Setting] = [
             "How many rotated shards are kept behind the live one. Records in a dropped shard are "
             "gone from the store, so this is a retention policy and not only a disk setting: at "
             "the defaults the store keeps roughly 336 MB and discards the oldest records past it."),
+    # The sharpest memory/latency trade in the local store, and it was not offered at all.
+    # Measured on a 100,105-record store, warm read, three alternating runs each:
+    #   sharing on : 8.00 s, 2,305 B/record, 761 MB resident
+    #   sharing off: 4.58 s, 4,173 B/record, 1,041 MB resident
+    # So it costs ~3.4 s on the read that loads the cache and returns ~45% of the cache. Which way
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Value sharing returns ~45% of the read cache and 280 MB resident for ~3.4 s on the read that loads it, and the variable was not offered anywhere on the portal. Measured on a 100,105-record store, three alternating runs per arm.

PR #747, opened 2026-09-04
A daemon whose engine is restarting is not a dead daemon

bjmeetsfo:
The daemon answered `__daemon_health` with `ok: false` whenever its engine was momentarily
gone. But the engine is restarted on demand — `_ensure_proxy` starts one on the next request —
so that answer is about a transient, while callers read it as "there is no daemon here" and
respond by starting their own. A second daemon unlinks this one's socket on bind, so the loser
falls back to spawning an engine per invocation and reloading the whole store each time.

Measured on a box where that was happening: two daemons, each holding a full copy of the store,
**669 MB of resident memory where one engine costs 343 MB**.

## What changed

- Health now reports whether this socket will serve the next request. A live engine is healthy;
  no engine is still healthy, because one is started on demand. The condition worth reporting is
  an engine that cannot be started at all, so the fallback is an executable check on the proxy
  path.
- It is deliberately **lock-free**. Taking the daemon lock made the health check queue behind an
  in-flight request, and a cold load takes tens of seconds — long enough that every ping timed
  out and read as a dead daemon, which is the exact failure this answer exists to prevent. The
  first attempt at this fix had that bug; a ping measured 53 ms after it was removed, with no
  engine running.

## Tests

`tools/test_matrixark_rust_proxy_socket_deadline.py` gains one case covering all three states:
no engine yet is healthy; health returns in under a second while the daemon lock is held by a
stand-in for an in-flight request; and an engine that cannot be started is not healthy.

It fails against this branch's base, and the file's other five cases still pass (6/6).

### DIFF
--- a/tools/matrixark_rust_proxy_daemon.py
+++ b/tools/matrixark_rust_proxy_daemon.py
@@ -113,6 +113,21 @@ class RustProxyDaemon:
                 proc.kill()
         self._write_log({"event": "proxy_stopped", "returncode": proc.returncode})
 
+    def _health_ok(self) -> bool:
+        """Whether this socket can serve the next request.
+
+        Deliberately lock-free. Taking the daemon lock here would queue the health check
+        behind an in-flight request, and a cold load takes tens of seconds -- long enough
+        that every ping times out and reads as a dead daemon, which is the very thing this
+        answer exists to prevent. A live engine is healthy; no engine is still healthy,
+        because `_ensure_proxy` starts one on the next request. The condition worth
+        reporting is an engine that cannot be started at all.
+        """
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The daemon reported ok=false whenever its engine child was momentarily gone, but the engine is started on demand, so callers read it as "no daemon" and started their own - a second daemon unlinks the socket on bind, and I measured two daemons holding 669 MB where one engine costs 343 MB. Health now answers for the socket and stays lock-free (taking the lock queued it behind a cold load and timed out); one new test covers all three states and fails against the base.

PR #754, opened 2026-09-04
Drop the buckets a write emptied by name, not by walking every bucket

bjmeetsfo:
`sync_bucket_index_object_pages` ended by walking the whole bucket map to drop the buckets it had
just emptied:

```rust
if removed_any || dirty {
    shard.bucket_index.bucket_map
        .retain(|_, b| !b.page_index.is_empty() || !b.object_index.is_empty());
}
```

`dirty` is true for a write, so that ran on **every** command. At the default routing-slot range
each key gets its own slot, so the map holds a bucket per record and the walk is O(corpus) per
write. The rest of the function is careful to touch only the affected buckets; this undid that.

## What it cost

An 8-hour one-box run of the real ingest shapes -- a message is four commands in one batch, a
skill file is ~30 chunks -- degraded from **7 ms to 105 ms per message**, fitting `corpus^0.96`:
linear per message, quadratic overall. Throughput fell to 13 of a targeted 40 msgs/s. Zero
failures throughout, so it degrades rather than breaks.

`perf` on a degraded node put **94.22% of self time** in that one retain. Nothing else exceeded
0.16%.

## The change

Only a bucket whose `page_index` actually shrank can have newly become empty, and the function
already tracks that set as `touched_buckets`. Buckets the publish loop inserted into gained a
page; buckets the call never opened are unchanged. So the same buckets are removed, by name.

```
                      corpus   us/msg   buckets   growth
  before 0..u32::MAX   40000   38,898   121,200    2.97x
  after  0..u32::MAX   40000    4,464   121,200    1.07x
  after  0..1023       40000    4,034     1,024    0.94x
```

**8.7x faster at 40k records with the same 121,200 buckets** -- the win is not reading them, not
having fewer. The bounded-range arm is unchanged, which is the control: at 1,024 buckets the walk
was always cheap.

Confirmed at scale on the same instance, identical 15-minute reproduction:

```
  before   5.96 -> 19.85 ms  (3.3x)  over ~74k records,  46.9 msgs/s
  after    5.38 ->  5.55 ms  (1.0x)  over ~92k records,  93.6 msgs/s
```

Flat at a larger corpus, with double the throughput.

## The test is mutation-tested

`emptied_buckets_are_dropped_without_walking_the_map` pins two properties, because each alone is
satisfiable by a wrong change -- never dropping anything is flat, and walking everything is
correct:

- emptied buckets must be gone after deleting every key
- per-write cost must not grow from a 2k to a 20k corpus

```
  with the fix     0.93x  PASS
  without the fix  3.41x  FAIL -- "the whole-map walk is back"
```

An earlier version drove `ContextUpsertNode` alone and passed on the unfixed code; upserts do not
reach the removal path. It drives the real three-command batch now.

## Why no existing test saw it

The narrow-routing-range tests run at 64 and 1023 slots, where the map holds ~1024 buckets and the
walk is cheap. They cannot observe default-range behaviour, and all three report `0.0` page
visits, which reads as proof of flatness and is not.

## Verification

- `cargo test -p temporalstore-rust --lib -- --test-threads=1`: **1610 passed, 0 failed**
- before/after on the same probe, same corpus, differing only in this change
- an 8-hour AWS run to find it, and a 15-minute reproduction on the same box to confirm the fix

Two files, no deletions.

### DIFF
--- a/crates/temporalstore-rust/src/engine/storage_bucket_internals.rs
+++ b/crates/temporalstore-rust/src/engine/storage_bucket_internals.rs
@@ -1667,11 +1667,32 @@ pub(super) fn sync_bucket_index_object_pages(
         }
     }
 
-    if removed_any || dirty {
-        shard
-            .bucket_index
-            .bucket_map
-            .retain(|_, bucket| !bucket.page_index.is_empty() || !bucket.object_index.is_empty());
+    // Drop buckets this call emptied -- BY NAME, not by walking the map.
+    //
+    // This was `bucket_map.retain(..)`, which is O(buckets) and ran on every write because
+    // `dirty` is true for one. At the default routing-slot range every key gets its own slot, so
+    // the map holds one bucket per record and the walk is O(corpus) per command: measured at
+    // 94.2% of the datanode's self time under a message-ingest profile, and an 8-hour run
+    // degraded from 7 ms to 105 ms per message as `corpus^0.96` -- linear per write, quadratic
+    // overall.
+    //
+    // Only a bucket whose `page_index` actually shrank above can have newly become empty, and
+    // that set is `touched_buckets`. Buckets the publish loop inserted into gained a page, and
+    // buckets this call never opened are unchanged. So the same buckets are removed, without
+    // reading the ones that cannot have changed.
+    if removed_any {
+        for routing_bucket in &touched_buckets {
+            let now_empty = shard
+                .bucket_index
+                .bucket_map
+                .get(routing_bucket)
+                .is_some_and(|bucket| {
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged — flagging for awareness, no action needed. The empty-bucket drop at the end of `sync_bucket_index_object_pages` walked the whole bucket map on every write; at the default routing-slot range that is one bucket per record, so ingest was quadratic (an 8-hour run went 7 ms to 105 ms per message, and a profile put 94.2% of self time in that one retain). It now removes the buckets the call actually emptied, by name: 8.7x faster at a 40k corpus with the same 121,200 buckets, and flat at scale.

PR #765, opened 2026-09-04
Count the vectors a tenant actually keeps

bjmeetsfo:
The end-to-end tenant test measured embedding storage by counting `context_embedding` rows.
`append_many` folds a vector onto the record that owns it and drops the separate row, so new
logs hold none for anybody — and the test could no longer see what it was guarding.

| | `context_embedding` rows | rows **carrying** a vector |
| --- | --- | --- |
| enterprise (`generate_embeddings: true`) | 0 | **15** |
| starter (`generate_embeddings: false`) | 0 | **0** |

So *"enterprise keeps its vectors"* failed on a store where enterprise had fifteen
vector-carrying rows — across `context_node`, `context_event`, `context_entity`,
`context_segment` and `context_summary` — and *"starter stores no vectors"* passed only
because nobody has a `context_embedding` row. It was true for every tenant, whatever the
policy did.

Counting rows that carry a vector — under `vector`, or under `embedding_meta` once compacted
— measures the thing the test is named for, and the same helper covers the live policy flip at
the end.

## Why this is more than a green test

Mutating the **product** so the tenant policy is ignored entirely and every tenant embeds:

| mutation | new assertions | old assertion |
| --- | --- | --- |
| policy ignored, always embed | killed | **SURVIVED — blind** |
| policy ignored, never embed | killed | — |

The old check would have passed while tenant embedding policy was switched off completely.
The new one fails in both directions it exists to catch.

## A correction to my own working note

The segment half of this test was never broken. An earlier probe of mine reported
`segments=0` for enterprise because it matched tenants on `scope.tenant_id` alone. The test's
own `_counts` already handles the interned form, where the scope is reduced to a `scope_key`
holding the tenant hash — the new helper attributes rows the same way. Enterprise has 3
segments and the assertion passes.

## No fallout

Full `unittest discover` before and after: 106 failing names → 105, nothing new.

### DIFF
--- a/tools/test_tenant_policy.py
+++ b/tools/test_tenant_policy.py
@@ -179,6 +179,25 @@ class LiveTenantPolicyEndToEndTest(unittest.TestCase):
             out[key] = out.get(key, 0) + 1
         return out
 
+    def _vector_rows(self, records, tenant):
+        """How many of a tenant's rows actually CARRY a vector.
+
+        Counting context_embedding rows stopped measuring this: append_many folds a vector onto
+        the record that owns it and drops the separate row, so new logs hold none for anybody.
+        That made "starter stores no vectors" true for every tenant, passing while proving
+        nothing -- the assertion could no longer fail if the policy broke. A vector rides on its
+        owner under `vector` or, once compacted, under `embedding_meta`.
+        """
+        identities = {tenant, policy.tenant_hash_of(tenant)}
+        total = 0
+        for record in records:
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Merged. @superhaiou for awareness — the tenant test counted context_embedding rows, which the vector fold stopped producing, so the starter assertion was true for every tenant regardless of policy. Mutating the product to ignore tenant policy entirely left the old assertion passing; the new one catches it in both directions.

PR #767, opened 2026-09-04
close() waits for the audit writes it started

bjmeetsfo:
close() waits for the audit writes it started

MatrixArkMcpServer.close() stops every background worker it owns and waits for it -- the summary
thread, the stream materialize thread, the adapter -- each bounded by timeout_s. Then it ended
with:

    self._audit_executor.shutdown(wait=False, cancel_futures=False)

which neither cancels queued audit writes nor waits for in-flight ones. With auditing async, an
audited hot-path call hands its write to a pool thread, so close() could return while writes were
still landing in the event log. A caller that closes usually removes that directory next, which
is how this surfaced: the governance fixture is the one place that turns auditing on, and its
temporary directory was being torn down under a live writer.

The pool and its drain now live in tools/matrixark_mcp_audit_queue.py. That is where the
reasoning belongs, and it keeps the entrypoint at its size budget, which
test_mcp_entrypoint_stays_small pins at 750 lines with no headroom -- main sits exactly on it, so
any new top-level import trips it. Only __init__ names the class; close() and the audit path go
through the instance. The budget is left alone: raising a guard in the change it blocks is how
guards stop meaning anything.

Waiting is bounded by the same timeout_s as the joins above, not shutdown(wait=True), which takes
no timeout and would let one wedged write hang shutdown forever.

The test makes the audit write slow on purpose. A first version used a fast one and passed
whether or not close() waited -- removing the drain did not turn it red, so it proved nothing. It
also asserts the write was submitted before asserting none is outstanding, because "nothing still
running" is trivially true of a set nothing was ever added to.

Checked by removing the drain and keeping the tracking: 10 of 10 red without it, 10 of 10 green
with it.

### DIFF
--- a/tools/matrixark_mcp_audit_queue.py
+++ b/tools/matrixark_mcp_audit_queue.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2026 MatrixArkAI
+"""The pool async audit writes run on, and the bounded drain close() needs.
+
+Kept beside the server rather than inside it: this is background-worker plumbing, and the
+entrypoint carries a size budget that test_mcp_entrypoint_stays_small enforces.
+"""
+from __future__ import annotations
+
+import threading
+from concurrent.futures import ThreadPoolExecutor, wait as wait_for_futures
+from typing import Any, Callable, Set
+
+
+class AuditWriteQueue:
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. close() now waits for the async audit writes it started, bounded by the same timeout as its other joins; the pool moved to a sibling module so the entrypoint stays on its 750-line budget.

PR #775, opened 2026-09-04
Remove eight switches and the paths they kept alive

bjmeetsfo:
Eight switches removed, and with them the paths they kept alive.

## How they were chosen

Not by reading names. Each of these flags is read through a small gate function whose docstring
says what the alternate position does, and the useful ones say the same thing:

- *"Default ON; set `MATRIXARK_VALUE_ENTITY_CAPTURE=0` to restore prior behavior"*
- *"`MATRIXARK_DROP_DERIVABLE_ROUTE=0` keeps the old persisted route"*
- *"`MATRIXARK_COMPACT_LOCATIONS=0` ... the escape hatch"*

So the classifier is over that prose, and it matches a **verb** — *restores the prior*, *keeps the
old*, *returns the writer to the long form*, *escape hatch* — rather than a mood. Matching words
like "before" or "previously" flags anything that mentions its own history; requiring a phrase that
says the off position **restores** something gives a list you can act on. That found 9 flags. Three
survive, for reasons below.

Every one removed here clears the same bar: default ON, **exactly one branch site**, **no test sets
it**, and **not offered on the portal**.

## What went

| flag | what its off position restored |
|---|---|
| `MATRIXARK_VALUE_ENTITY_CAPTURE` | not minting a fact entity for an exact value the segmenter skipped |
| `MATRIXARK_SEGMENT_ACCESS_SCOPE` | writing `context_segment` rows with no access scope |
| `MATRIXARK_LEXICAL_EXACT_RECALL` | dropping the exact-recall lane |
| `MATRIXARK_DROP_DERIVABLE_ROUTE` | persisting a route that is derivable |
| `MATRIXARK_COMPACT_LOCATIONS` | emitting locations in the long form |
| `MATRIXARK_INTERN_METADATA_BUNDLE` | writing the Phase-1 per-field intern map |
| `MATRIXARK_DIRECT_SERVING_APPEND_TO_BACKEND` | skipping the append that stamps `expires_at_ms` |
| `MATRIXARK_RETRIEVE_TIMEOUT_MS` | nothing — a second spelling of one knob |

Two are worth a note.

**The intern bundle** is the only one that touches a durable format, and only the **write** side
goes. The per-field map is still read, so an old log expands exactly as before — that is what
backward compatibility requires. Writing two formats is not. Its per-value tokeniser had one
caller, the arm just deleted, and goes with it.

**The retrieve deadline** had two env names for one value, and the tidier-looking one was the wrong
one to keep. `MATRIXARK_RETRIEVE_TIMEOUT_MS` matches the shape of every sibling
(`MATRIXARK_<OP>_TIMEOUT_MS`) and appears in exactly one place in the tree: the lookup that
preferred it. `MATRIXARK_RETRIEVAL_TIMEOUT_MS` is what the server launch script, the agent config,
the scale test and both retrieve deadline reads actually use. Consistency would have removed the
one in use.

## What stayed, and why

`MATRIXARK_ALLOW_SHARED_OSS_MODEL_DRIFT` matches the classifier — its help calls it an escape
hatch — and is **kept**. It is a `--allow-shared-oss-model-drift` CLI switch on the benchmark
runner that enforces a baseline comparison using the same reader, embedding and budgets. That is a
fairness guard on a measurement, not a legacy code path, and removing it would remove the ability
to say a comparison was unfair.

The two `*_TIMEOUT_MS` entries the classifier flagged are values, not path selectors; one of them
is the consolidation above.

## Effect

- boolean feature flags: **69 → 63**
- flags whose alternate path restores what came before: **9 → 3**, and none of the three is a
  second code path
- **-75 lines**, most of it an 81-line block that no longer needs to be indented under a condition
  that was always true

## Tests

The whole suite was run on two full archive trees — main, and main with exactly these files
overlaid — and name-diffed. These gates were default ON and the ON path is what survives, so the
expectation is that nothing moves at all.

`test_intern_record_metadata` changed rather than being deleted: two of its cases branched on the
bundle flag to decide what to assert. With one write format there is one thing to assert, and it
now also asserts the per-field map is **never written**, which is the property the removal creates.

## What this leaves for a second pass

The same scan finds ten more knobs read under two names, of the shape
`os.environ.get(A, os.environ.get(B, default))`. They are NOT all duplicates, and the difference
decides whether consolidating them is a cleanup or a regression:

- a genuine **alias** is two spellings of one concept, like the deadline above;
- a **hierarchy** is a specific override falling back to a shared default, and
  `MATRIXARK_CLAUDE_HOOK_STORE_BASE` falling back to `MATRIXARK_SHARED_HOOK_STORE_BASE` is exactly
  that: the Claude and Codex hooks deliberately share one store base and differ elsewhere.
  Collapsing it would merge two agents' stores.

Several of the ten also have the property that caught me above — the name being *preferred* has
fewer uses than the one it falls back to. `MATRIXARK_SUMMARY_PROVIDER` is read first and used in 4
places; `MATRIXARK_UNDERSTANDING_PROVIDER`, its fallback, is used in 12.

So that pass needs the same per-knob evidence this one used, not a sweep. It is left out here
rather than bundled in.

### DIFF
--- a/tools/matrixark_local_adapter_ingest.py
+++ b/tools/matrixark_local_adapter_ingest.py
@@ -43,23 +43,6 @@ def _context_debug_records_enabled() -> bool:
     except Exception:  # pragma: no cover - policy module absent
         return False
     return bool(context_debug_records_enabled())
-
-
-def _segment_access_scope_enabled() -> bool:
-    """Gate: write context_segment records WITH a tenant/user/session access_scope so a
-    scored segment passes access_scope_matches_before_scoring instead of being dropped
-    scope-less. Default ON; set MATRIXARK_SEGMENT_ACCESS_SCOPE=0 to restore prior behavior."""
-    return str(_os.environ.get("MATRIXARK_SEGMENT_ACCESS_SCOPE", "1")).strip().lower() not in {"0", "false", "no", "off"}
-
-
-def _value_entity_capture_enabled() -> bool:
-    """Gate: mint a scoped, embedded fact entity for any assistant message stating an exact
-    value (number+unit, count, hex hash, ALL_CAPS/dotted flag) that the topic segmenter and
-    entity extractor skipped, with its state text NOT truncated before the value token.
-    Default ON; set MATRIXARK_VALUE_ENTITY_CAPTURE=0 to restore prior behavior."""
-    return str(_os.environ.get("MATRIXARK_VALUE_ENTITY_CAPTURE", "1")).strip().lower() not in {"0", "false", "no", "off"}
-
-
 _VALUE_TOKEN_RE = _re.compile(
     r"(?:\b0x[0-9a-fA-F]{6,}\b)"
     r"|(?:\b[0-9a-f]{7,40}\b)"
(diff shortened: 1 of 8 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Eight switches removed whose only alternate path restored the behaviour they replaced; each was default ON with one branch site, set by no test and offered on no page. 69 -> 63 boolean flags, -80 lines. The intern bundle drops only the WRITE side; the per-field map is still read so old logs expand unchanged.

PR #780, opened 2026-09-04
One switch for the shadow comparison, not twelve

bjmeetsfo:
Twelve switches for one diagnostic become two.

## What they were

Six operations each carried a pair:

| | |
|---|---|
| `MATRIXARK_GETALL_SHADOW_COMPARE` | `MATRIXARK_GETALL_SHADOW_LOG` |
| `MATRIXARK_SESSBUF_SHADOW_COMPARE` | `MATRIXARK_SESSBUF_SHADOW_LOG` |
| `MATRIXARK_REFRESH_SHADOW_COMPARE` | `MATRIXARK_REFRESH_SHADOW_LOG` |
| `MATRIXARK_DELETE_SHADOW_COMPARE` | `MATRIXARK_DELETE_SHADOW_LOG` |
| `MATRIXARK_GETMEM_SHADOW_COMPARE` | `MATRIXARK_GETMEM_SHADOW_LOG` |
| `MATRIXARK_PRIOR_CONTEXT_SHADOW_COMPARE` | `MATRIXARK_PRIOR_CONTEXT_SHADOW_LOG` |

All twelve are read by copies of the same two lines, in one module, for one thing: run the narrowed
read and the full read and compare them.

Nothing used the granularity. No test sets any of them, nothing outside that module mentions them,
and none is offered on the portal — so the ability to shadow one operation and not another was
theoretical.

## What replaces them

`MATRIXARK_SHADOW_COMPARE` turns the comparison on, wherever it runs. `MATRIXARK_SHADOW_LOG`
redirects where it writes.

**The per-operation default log paths are kept deliberately.** Two reads are easier to compare when
their logs are separate files, so `_shadow_log_path("getall")` still defaults to
`/tmp/matrixark_getall_shadow.log`. What collapsed is the six ways to override that path, not the
six paths.

## Three things that made this less mechanical than it looks

- `MATRIXARK_PRIOR_CONTEXT_SHADOW_LOG` is read at **two** sites, not one. The patch asserted
  "exactly one" and stopped rather than writing a half-applied file.
- `MATRIXARK_DELETE_SHADOW_COMPARE` inlines its condition (`if shadow not in {...}:`) where the
  other five assign a `shadow_on` flag first, so a single search-and-replace would have missed it.
- This module imports `os as _os` **inside each function**, not at module scope. A module-level
  helper referring to `_os` parses cleanly and raises `NameError` the first time it runs — which is
  why the tests here drive the switch rather than reading the file.

## Tests

Five. Beyond "the switch works", they pin the properties that make the consolidation safe rather
than merely smaller: that it is **off** unless asked (so the on-case proves something), that each
operation still resolves to its **own** default log path (so the six comparisons do not interleave),
that one override redirects all of them, and that no per-operation name is read any more — with a
positive control asserting the surviving names ARE present, so "not found" cannot mean "searched
the wrong file".

### DIFF
--- a/tools/matrixark_mcp_temporal_adapters.py
+++ b/tools/matrixark_mcp_temporal_adapters.py
@@ -460,6 +460,32 @@ def _prior_context_event_window() -> int:
     return max(value, 0)
 
 
+
+def _shadow_compare_enabled() -> bool:
+    """One switch for the shadow comparison, whichever read is running it.
+
+    There were six of these, one per operation, and they were six copies of the same two lines.
+    Nothing used the granularity: no test set any of them, nothing outside this module mentioned
+    them, and none was offered on the portal. One name is easier to find, and easier to turn off
+    again after a diagnosis.
+    """
+    import os as _os
+    return _os.environ.get("MATRIXARK_SHADOW_COMPARE", "").strip() not in {
+        "", "0", "false", "no", "off"
+    }
+
+
+def _shadow_log_path(operation: str) -> str:
+    """Where the comparison writes. One override; the default still names the operation.
+
+    The six per-operation defaults are kept deliberately -- comparing two reads is easier when
+    their logs are separate files -- so what collapsed is the six ways to override the path, not
+    the paths themselves.
+    """
+    import os as _os
+    return _os.environ.get("MATRIXARK_SHADOW_LOG", "/tmp/matrixark_%s_shadow.log" % operation)
+
+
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Twelve switches for one diagnostic became two: six <OP>_SHADOW_COMPARE and six <OP>_SHADOW_LOG were copies of the same two lines in one module, set by no test and offered on no page. The per-operation default log paths are kept so the comparisons stay in separate files.

PR #789, opened 2026-09-04
Leave the folded row identity out of the meta it leaves behind

bjmeetsfo:
A folded embedding leaves its own row identity behind, and that one field keeps the whole dict from
being shared.

## What it costs

`embedding_meta` is **12.3% of the durable log** and effectively unique per record. Broken down over
6,000 sampled values:

| sub-field | share of embedding_meta | distinct values |
|---|---|---|
| **row_key** | **36.7%** | **5,999 (unique)** |
| node_path | 22.5% | 1 |
| model_ref | 20.9% | 1 |
| model | 15.8% | 1 |
| dim | 3.6% | 1 |

Every other member holds ONE distinct value across the store. So the unique one is not just the
largest — it is what stops the dict being shared at all.

Measured on an 8-skill store, before and after:

| | log | read cache | embedding_meta objects | distinct values |
|---|---|---|---|---|
| before | 1.32 MB | 0.75 MB | 164 | 164 |
| after | 1.29 MB | **0.67 MB** | **5** | **5** |

**-10.7% of the read cache for -2.3% on disk.** The cache falls 4.6x further because 164 rows now
share 5 objects instead of holding 164. Retrieval is byte-identical: the same 32 refs, the same
16 skill_section and 16 resource_chunk.

## Why it is safe to drop

`row_key` here is the identity of the `context_embedding` row that was folded onto its owner and no
longer exists. Checked for readers the way the entries already in `_EMBEDDING_META_SKIP` were:

- no reader in any Python module outside the adapter, and inside it only the **top-level**
  `row_key` is used (`ROW_KEY_FIELD` / `canonical_row_key` / `stamp_row_keys`);
- no occurrence of `row_key` anywhere in the Rust crates;
- a runtime probe wrapping every `embedding_meta` in a recording dict across `read_all` +
  `retrieve`: `model` was asked for **99,324** times; `row_key`, `node_path`, `dim` and
  `embedding_type` were **never** asked for, and `model_ref` only by the probe's own positive
  control.

`model` and `model_ref` stay, deliberately — the same reason the existing comment gives for keeping
them despite scanning clean.

## It is also wrong to carry

`record_with_embedding_defaults` copies **every** key of the meta onto an owner whose value is
empty, with no exclusion list. An owner without a top-level `row_key` — 99,280 of 100,122 records —
therefore inherits the retired row's identity, and `latest_value_record_key` prefers a stamped key
over deriving one.

The test shows it on `main` rather than arguing it:

```
'["context_node",5689206235060161283]'
  != '["context_embedding","context_node","node",5689206235060161283]'
```

That is a record's own identity against the one its meta hands it.

## Tests

Five, three of which fail on `main` on their own claims. Two are there to stop this being a
size-only change: a **positive control** asserting `model`/`model_ref` are still present, so a
future trim cannot quietly take the model identity the guard depends on; and a non-vacuity case
asserting there is any `embedding_meta` to examine at all.

Validation: every test module naming `embedding_meta` or the recovery helper — 13 of 13 ran on both
trees, 5 failing names each, **0 unique**.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -1247,6 +1247,21 @@ _EMBEDDING_META_SKIP = (
     "ref_type",
     "ref_hash",
     "vector",
+    # The retired row's OWN identity, which describes a row that no longer exists once the
+    # embedding is folded onto its owner. 36.7% of embedding_meta's bytes over 100,105 records,
+    # and the only unique field in a dict whose other members hold one distinct value each -- so
+    # it is also what keeps the dict from being shared: 99,371 objects for 99,371 values.
+    #
+    # Checked for readers the way the rest of this tuple was: no reader in any Python module
+    # outside this one, and inside it only the TOP-LEVEL row_key is used; no occurrence anywhere
+    # in the Rust crates; and a runtime probe over read_all + retrieve never saw it asked for,
+    # while `model` was asked for 99,324 times.
+    #
+    # It is also wrong to carry. `record_with_embedding_defaults` fills every meta key onto an
+    # owner whose value is empty, with no exclusion list, so an owner without a top-level row_key
+    # -- 99,280 of 100,122 records -- inherits the RETIRED row's identity, and
+    # `latest_value_record_key` prefers a stamped key over deriving one.
+    "row_key",
     "storage_route",
     "storage_options",
     # The placement/storage half of the same routing blob, inherited the same way and read by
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. embedding_meta carried the identity of the context_embedding row folded away to make it, and that one unique field kept the whole dict unshareable: read cache -10.7% for -2.3% on disk, 164 meta objects becoming 5, retrieval byte-identical. It was also wrong to carry - record_with_embedding_defaults copies every meta key onto an owner whose value is empty, so an owner without a top-level row_key inherited the retired row's identity.

PR #797, opened 2026-09-04
Give the manifest preview a previews size

bjmeetsfo:
A skill manifest's "preview" is 3,096 characters of the skill.

## The measurement that makes this worth doing

The index costs a **flat ~11 KB per document**, whatever the document's size. Four documents, same
count, swept across a 100x size range on one build:

| document size | index bytes | index share (32-dim) | at 384 dims |
|---|---|---|---|
| 8 KB | 0.05 MB | 17.3% | 9.7% |
| 40 KB | 0.05 MB | 6.6% | 3.1% |
| 200 KB | 0.05 MB | 1.6% | 0.7% |
| 806 KB | 0.05 MB | 0.4% | 0.2% |

The bytes do not move; only the denominator does. So "the index is negligible" is a statement about
large documents, not about the index — on a box holding many small skills it is around **10% of
everything stored** at production embedding width.

Broken down, `skill_manifest.text_preview` is **29% of that fixed cost**, the single largest item,
ahead of `context_index.ref_hashes` at 9.3%.

## Why it is that big

`clip_context_text` defaults to `MAX_CONTEXT_REF_CHARS` — **4096** — which is the size a served
context *ref* may reach, because a ref carries content. A manifest preview is not a ref. The codex
hook's own `text_preview` uses 260 characters; this one inherited the ref budget by default.

## What changes

The manifest preview is clipped at 260 characters.

| | manifest preview | log | read cache |
|---|---|---|---|
| before | 3,096 chars | 0.447 MB | 0.277 MB |
| after | **275 chars** | 0.426 MB | 0.266 MB |

**-4.7% of the log and -4.0% of the cache** on a four-skill store, and the saving is per document,
so it grows with the number of skills rather than washing out.

## Trimmed, not removed

Nothing that serves skills reads this field: `list_skills` emits 24 manifest fields and it is not
among them, and the index-term path reads `skill_hash` / `node_hash` / `name` / `triggers` /
`allowed_tools`. A runtime probe wrapping every manifest in a recording dict across `list_skills` +
`retrieve` recorded it asked for **zero** times, while 14 other manifest fields were read 4-44
times — with a positive control proving the probe records at all.

It stays anyway, because `matrixark_http` falls back to it for display. That is what a preview is
for; it just has to be one.

## Tests

Five, and the bound fails on `main` on its own claim — *"a manifest preview is 4111 chars against a
260-char budget"* — rather than erroring on a missing constant, so it is known to discriminate.
Two exist to stop the change being hollow: that the fixture's skills are long enough for a clip to
actually happen, and that the preview still **starts the skill it previews** rather than merely
being short.

### DIFF
--- a/tools/matrixark_local_adapter_ingest.py
+++ b/tools/matrixark_local_adapter_ingest.py
@@ -13,6 +13,13 @@ import warnings as _warnings
 _PROFILE_SCOPE_WARNED: set = set()
 
 
+
+#: How much of a skill's text its manifest carries as a preview.
+#:
+#: Not MAX_CONTEXT_REF_CHARS (4096): that is what a served context ref may reach, and a manifest
+#: preview is not a ref. 260 is what the codex hook's own text_preview uses.
+SKILL_PREVIEW_CHARS = 260
+
 def _segments_enabled(scope) -> bool:
     """Whether this tenant materialises context_segment rows (default OFF).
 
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The manifest preview was clipped at MAX_CONTEXT_REF_CHARS (4096), the budget for a served ref, so it held 3,096 characters of the skill. The index costs a FLAT ~11 KB per document whatever the document's size, and this field was 29% of it - so on small skills it is around 10% of everything stored at production embedding width. Trimmed to 260, not removed, because matrixark_http falls back to it for display.

PR #803, opened 2026-09-04
Say which settings are waiting for a restart

bjmeetsfo:
The portal badges **needs restart** on every setting whose `applies` is `restart`. That is a
property of the setting, true from the moment the page loads and before anybody touches anything —
*advice*. Nothing said that a setting had **been** changed and this process was still running the
old value — *status*.

`update()` reports `restart_required` in its own response, so whoever makes the change is told
once, at the moment they make it. Anybody arriving afterwards — the next operator, the same one on
Monday — sees a page describing a configuration the deployment is not running. Settings frozen at
import keep their startup value indefinitely, and nothing on any screen says so.

### Where the answer comes from

The value a restart-scoped setting actually took is knowable at exactly one moment: `apply_boot`,
once the stored file has been folded into the environment and the launcher's precedence resolved.
Recording it there turns *is this pending?* into a comparison rather than a guess.

```
at boot, pending:                 []
after the write, pending:         ['extraction.base_url']
after a restart, pending:         []
a LIVE setting written, pending:  []
```

**Absence of a record means unknown, and unknown reports False.** A process that never called
`apply_boot` — a test, a tool importing the module — has nothing to compare against, and answering
"pending" there would put an alarm on every deployment that merely imports this file. That case has
its own test.

### On the page

The status badge **replaces** the advice badge rather than joining it. Two badges reading "needs
restart" and "changed — restart to apply" on one field say the same thing twice and bury the half
that is about this deployment today. It is coloured as a state rather than a note, because a
deployment running something other than what the page shows is not a footnote.

This is asserted by running the page's own `fieldHtml` against four field shapes, because whether
two badges are mutually exclusive is behaviour and the source reads the same either way. The
harness resolves the renderer's helpers generically rather than one ReferenceError at a time.

Complements `boot_pinned`, which already reports the mirror image: a launcher-set variable that
makes a portal write vanish at the next restart.

Suites green: gateway_config (28), config_audit (14), clamped_settings (8), every_live_claim (4),
portal_pages (4), dashboards (20), plus 11 new.

### DIFF
--- a/tools/matrixark_gateway_config.py
+++ b/tools/matrixark_gateway_config.py
@@ -50,6 +50,15 @@ Json = Dict[str, Any]
 # be overwritten by the stored file.
 _BOOT_ENV: Dict[str, str] = dict(os.environ)
 
+# What each `applies == "restart"` setting was actually running as, recorded by apply_boot() once
+# the stored file has been folded into the environment. Those settings are read into module
+# constants at import and never re-read, so this is the only moment their effective value is
+# knowable -- and without it "has this been changed since startup?" cannot be answered at all.
+#
+# Empty means nobody called apply_boot (a test, a tool importing this module). Then nothing is
+# reported as pending: missing evidence is not evidence of a pending change.
+_BOOT_EFFECTIVE: Dict[str, str] = {}
+
 _TRUTHY = {"1", "true", "yes", "on"}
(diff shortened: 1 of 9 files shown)


bjmeetsfo:
-

PR #805, opened 2026-09-04
The snapshot stored what the log compressed

bjmeetsfo:
The durable read-cache snapshot and the durable log hold the same records. The log writes each one
with its interned metadata replaced by a bundle token, plus one sidecar per distinct bundle. The
snapshot was writing `"records": records` verbatim — fully expanded — so the largest file in a
store was the one copy of the data that had opted out of the compression.

Nothing in the code relates the two files, which is why this held: the writer that compresses and
the writer that does not are two hundred lines apart and neither mentions the other.

## The snapshot is the file that matters

Across 14 stores real runs left behind, the base snapshot is consistently larger than the log —
1.4x typically, and 9.7x on the largest (207 MB of snapshot against 22 MB of log, because the log
compacts and the snapshot holds everything).

The delta tail was **0 bytes on all 14**, so the base is the whole cache and fixing the base covers
all of it.

## What it saves

The saving is the per-record cost of the interned fields, so it tracks how much of a record is
metadata rather than body:

| store | as written | interned | saving |
|---|---|---|---|
| 3,022 small records | 13.4 MB | 4.3 MB | **67.8%** |
| 100,105 records, text and vectors dominant | 207.2 MB | 186.3 MB | **10.1%** |

Sidecars stay flat as the corpus grows — 22 for 99,919 tokened records, and 17 whether the store
holds 105 records or 1,225 — so this is the whole per-record cost of those fields rather than a
ratio that decays with corpus size.

Fewer bytes on disk is also fewer bytes to parse on a cold load, since the snapshot is read with a
single `json.load`.

## The change

- `_write_durable_read_cache` encodes through `encode_interned_records`, the same encoder the log
  uses. `record_count` still counts data records; the encoded list is longer by its sidecars.
- `_load_durable_read_cache` calls `expand_interned_records` before the count check. It is called
  unconditionally rather than behind the intern flag, because the flag can differ between the
  process that wrote the file and the one reading it; with nothing interned it takes the no-op path,
  so an existing snapshot loads unchanged.
- `LOCAL_DURABLE_READ_CACHE_SCHEMA_VERSION` 1 → 2. This is the safety: an older reader would
  otherwise serve tokenised records with the interned fields silently missing. An unrecognised
  version sends it back to the log, which is the same fallback a missing snapshot already takes.

## Tests

`test_matrixark_snapshot_stores_what_the_log_compresses.py`, five tests: the snapshot really is
written interned (with a positive control, since a corpus with nothing to intern would satisfy
every other assertion vacuously); a reader gets back exactly what a reader of the log gets, values
compared rather than counted; the saving is large enough to be worth the stored format; an older
reader is sent back to the log and the store still serves; and the expansion yields one shared
object per distinct value rather than a copy per record.

Two of the five fail on `main` and three pass there — the three are invariants that should hold
either way. Both arms built with `git archive`, so neither carries anything from a working tree.

Full python suite run on both arms and diffed by failing test NAME: no failure on this branch that
`main` does not already have.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -149,7 +149,11 @@ def _snapshot_prefix_fingerprint(record: "Json") -> str:
         ).hexdigest()[:32]
     except (TypeError, ValueError):
         return ""
-LOCAL_DURABLE_READ_CACHE_SCHEMA_VERSION = 1
+# Version 2 stores the base snapshot in the same interned form the durable log uses, so the two
+# files no longer disagree about how the same records are written. The bump is what protects an
+# older reader: it would otherwise serve tokenised records with the interned fields missing, so a
+# version it does not recognise has to send it back to the log, which it already does.
+LOCAL_DURABLE_READ_CACHE_SCHEMA_VERSION = 2
 PRE_RETRIEVAL_SUMMARY_REFRESH = os.environ.get("MATRIXARK_PRE_RETRIEVAL_SUMMARY_REFRESH", "0").strip().lower() in {"1", "true", "yes"}
 
 QUALITY_FIRST_UNDERFILL_DROP_KEYS = {
@@ -4742,6 +4746,12 @@ class MatrixArkLocalAdapter(_LocalAdapterRetrieveMixin, _LocalAdapterIngestMixin
         if not isinstance(records, list):
             return None
         records = [record for record in records if isinstance(record, dict)]
+        # Undo the interning the writer applied, dropping the sidecars, so what comes back is the
+        # record set the caller stored -- the same inverse the log path applies at its own read.
+        # Called unconditionally rather than behind the intern flag, because the flag can differ
+        # between the process that wrote the file and the one reading it; with nothing interned
+        # this takes the no-op path. It runs before the count check, which counts data records.
+        records = expand_interned_records(records)
         if len(records) != head.get("record_count"):
             return None
         delta_count = head.get("delta_count") or 0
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
Follow-up measurement: this turns out to be a latency change too, not only a stored-bytes one.

Profiling a retrieve on `main` shows the read path fully rewriting the snapshot — `json.dump` is
1.49 s of a 1.4 s query, 908k `_iterencode` calls, reached through
`_refresh_durable_read_cache_if_behind`. Since a retrieve pays a full rewrite, halving what that
rewrite emits shows up directly in query time.

Interleaved A/B, both orderings, 10 rounds per arm, each arm on its own store (reading a store
rewrites its snapshot, so a shared copy would have the second arm loading the first arm's file):

| arm | median | best | worst | snapshot |
|---|---|---|---|---|
| main | 1642.1 ms | 1332.9 | 3441.4 | 7136.9 KB |
| this branch | **1290.2 ms** | 1192.0 | 1645.7 | **3591.3 KB** |

**−21.4% on the median**, and the branch wins on best and worst as well, in both orderings — this
box has previously shown a position effect where whichever arm runs second wins, so the ordering
was reversed rather than assumed away.

Two things this does **not** claim:

- The corpus here is 128 documents / 2,330 records, where the interned fields are a large share of
  each record and the snapshot halves. On a 100,105-record store whose text and vectors dominate,
  the stored saving is 10.1%, so the latency effect there will be correspondingly smaller.
- The full rewrite on a read path is a separate defect, and a much larger one — it is what makes a
  query cost ~1.3 s at 2,330 records against ~5 ms at 602. This change makes that rewrite cheaper;
  it does not stop it happening. That is worth its own fix.

bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The durable snapshot now stores records interned like the log does, which halves it and cuts retrieve latency 21.4% because a query pays for a full snapshot rewrite.

PR #810, opened 2026-09-04
A chunk pays for the branches it never takes

bjmeetsfo:
`candidate_memory_layer_name` answers `resource_chunk` or `skill_section` for almost everything it
is asked about, and computed a dozen lookups and three set comprehensions before getting there.

Over one retrieve on a 2,346-record store: **88,340 calls, of which 84,841 (96%) end at one of four
early answers** — 63,328 `skill_section`, 21,513 `resource_chunk`. Those four need nothing but
`context_class` and `ref_type`, both already computed. Everything between them and the top of the
function — `memory_scope`, `session_continuity`, the profile class and kind, three set
comprehensions over source lists, `event_type`, and the two `is_*` flags built from all of it — is
read by none of them.

So a chunk paid for the compression, summary, segment, event and entity branches on every call.
The function was **27.8% of a retrieve**.

## The change

Move the four returns above that block. Nothing else moves, and their order relative to each other
and to every later branch is unchanged, so this is code motion: none of the four reads a value the
block produces, and the block has no side effects.

## What it is worth

Interleaved A/B, both orderings, 10 rounds per arm, each arm on its own store:

| arm | median | best | worst | refs selected |
|---|---|---|---|---|
| main | 1498.8 ms | 1177.2 | 2119.2 | 64 |
| this branch | **1043.8 ms** | 982.1 | 1389.8 | 64 |

**−30.4% on the median**, and the branch wins on best and worst as well, in both orderings. Both
arms select the same 64 refs, so it is like-for-like rather than two different amounts of work.

Reversing the order matters here: this box has repeatedly shown whichever arm runs second winning
on work that was identical, so a single ordering proves nothing.

## Tests

`test_matrixark_layer_name_decides_early.py`, seven tests. Because this is a performance change the
tests pin *behaviour preservation* rather than the change — they pass on `main` too, by design, and
the measurement above is what carries the win.

The risk in moving a return earlier is that it shadows a later branch, which can only happen for a
candidate matching both. So the tests build exactly those: a chunk carrying every field the
compression, summary, segment, event and entity branches key on, which must still answer
`resource_chunk`. Plus a positive control that each later branch still answers (if the move had
swallowed the rest of the function, they would fall through), that an explicit `memory_layer` still
wins, that the metadata half of each lookup still applies, and that `ref_type` derived from
`record_type` still reaches the later branches.

One test asserts the function under test is the copy defined in the packing module — the name is
re-exported by several modules and the suite loads some of them twice, so a test could otherwise
pin a copy nothing calls.

Full python suite run on both arms and diffed by failing test name: no failure on this branch that
`main` does not already have.

### DIFF
--- a/tools/matrixark_mcp_core_packing.py
+++ b/tools/matrixark_mcp_core_packing.py
@@ -615,6 +615,19 @@ def candidate_memory_layer_name(candidate: Json) -> str:
     elif not ref_type and record_type == "context_segment":
         ref_type = "segment"
     context_class = str(candidate.get("context_class") or metadata.get("context_class") or ref_type)
+    # These four answers need nothing beyond context_class and ref_type, and they are almost all
+    # of the traffic: over one retrieve, 84,841 of 88,340 calls (96%) end here -- 63,328
+    # skill_section and 21,513 resource_chunk. Deciding them before the block below, rather than
+    # after it, is what keeps a chunk from paying for three set comprehensions and a dozen lookups
+    # it never reads. Measured at 27.8% of a retrieve spent in this function.
+    if context_class == "resource_entity_fact":
+        return "resource_entity_fact"
+    if context_class == "resource_fact":
+        return "resource_fact"
+    if ref_type == "resource_chunk":
+        return "resource_chunk"
+    if ref_type == "skill_section":
+        return "skill_section"
     memory_scope = str(candidate.get("memory_scope") or metadata.get("memory_scope") or "")
     session_continuity = str(candidate.get("session_continuity") or metadata.get("session_continuity") or "")
     profile_memory_class = str(candidate.get("profile_memory_class") or metadata.get("profile_memory_class") or "").strip().lower()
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
This composes with #805 (which cuts what the snapshot rewrite has to encode). The two touch
different files and merge cleanly, so here is what they are worth together.

Interleaved A/B, both orderings, 12 rounds per arm per ordering (24 per arm), each arm on its own
store, `main` against a scratch branch holding both changes:

| arm | median | best | worst | refs selected |
|---|---|---|---|---|
| main | 1130.5 ms | 6.1 | 1866.2 | 64 |
| both changes | **839.1 ms** | 6.8 | 2215.6 | 64 |

**−25.8% on the median.**

The per-round numbers matter more than the summary, because they show where the win is:

```
main    1866.2 1459.7 1700.6 1350.5 1354.9 1811.9 1018.5    8.1  8.5  8.7  9.0  8.4
branch  2215.6 1148.9 1123.1 1483.6 1199.0 1139.0  596.1    8.3 13.2 12.8 18.3  7.9
```

Both arms fall to **~7-9 ms from round 8**, once the store settles and a query writes nothing. The
whole cost -- and the whole win -- is in the churning phase, where each query still triggers a
snapshot rewrite. That is the real shape of a one-box workload that interleaves ingest and query,
and it is also why an idle benchmark shows nothing.

A first round of each block is slower on both arms (2215 and 1664 against 1866 and 1584), which is
warm-up rather than either change; the medians are taken across all 24 rounds in both orderings so
that a single block cannot carry the result.

Worth stating plainly: neither change stops the rewrite, they only make it cheaper and remove work
around it. A query that rewrites the base still costs ~1,100 ms against ~7 ms for one that does
not, and closing that gap needs the tail path to survive a compaction removal -- a separate and
much larger change.

bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. candidate_memory_layer_name decides its four common answers before the block they never read; pure code motion, -30.4% retrieve latency, and -25.8% combined with #805.

PR #815, opened 2026-09-04
Bank the fifteen suite names the runner reports as fixed

bjmeetsfo:
Bank the fifteen suite names the runner reports as fixed

The runner has been printing these under "fixed since the baseline" and asking for a re-record
nobody has done. They are removed by hand rather than by rerunning with --record here: the
baseline is recorded for the CI runner, and a full local run of this suite disagrees with it in
both directions, so recording on a developer box would bank that machine's differences into the
file that gates every merge.

Baseline goes 121 -> 106. Nothing is added, so the check only gets stricter.

The eleven names in the run's other list are deliberately left alone. Those are reported under
"no longer FAILING because they no longer RUN" -- a module raising SkipTest at import, and six
that drive an agent hook against a native binary this job never builds, so they have never run
here at all. Banking a skipped test drops it from the gate while it sits unexercised, which is
the opposite of what a ratchet is for. Both lists sit under the same "rerun with --record"
suggestion, and only one of them is safe to act on.

Each of the fifteen was checked to actually run and pass rather than skip, and each was confirmed
present in the baseline before removal.


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Baseline 121 -> 106, banking the fifteen names the runner reports as fixed; the eleven it lists as no-longer-running are deliberately left in the gate.

PR #817, opened 2026-09-04
Two constants cost a whole dict size step

bjmeetsfo:
CPython sizes a dict by **key count, in steps** — measured on this interpreter, not tabulated:

| keys | container |
|---|---|
| 11–21 | 640 B |
| 22–42 | 1,176 B |
| 43–85 | 2,272 B |

That matters more than it looks, because a live-heap walk of the read cache says the record dicts
**themselves** are 46.8% of it — 1,075 B per record before any value. More than the vectors, more
than the text. So a record sitting one key over a step pays the whole next one, and no per-field
byte census can see it: the field that pushes it over may hold two bytes.

Every record type sits on or near a boundary:

| record_type | rows | keys | container |
|---|---|---|---|
| `skill_section` | 192 | **21** | 640 B — the last slot before 1,176 |
| `resource_chunk` | 192 | **21** | 640 B — same |
| `resource_import_task` | 48 | **44** | **2,272 B — one key over the 43 step** |
| `context_index` | 112 | 17 | 640 B |

## The change

`resource_import_task` carried `index_cap_per_chunk` and `index_cap_per_fact` at the top level.
Both are process constants (`MAX_INDEX_TERMS_PER_RESOURCE_CHUNK` / `_FACT`), both are
byte-identical copies of entries already inside the record's own `metrics` dict, and neither is read
at the top level anywhere in `tools/` or `crates/`.

**Nineteen of the record's 44 keys duplicate its `metrics` that way.** These two are simply the two
that no reader depends on, which is why they are the ones that move.

Dropping them takes the record to 42 keys: **2,272 B → 1,176 B per row, −48%**, with both values
still present in `metrics`, so nothing is lost. Dropping the other seventeen would save nothing
further — 25 keys and 42 keys are the same step — so this is the whole available win, not a first
instalment.

## Tests

`test_matrixark_a_record_stays_under_its_dict_step.py` is the durable half of the change. It pins
the record types nearest a boundary — `resource_import_task`, and `skill_section` / `resource_chunk`,
which are the bulk of any corpus and sit at exactly 21 keys — so the next field added to one of them
fails here rather than silently costing 50–85% more container memory on every row.

Three things it does deliberately:

- The step boundaries are **measured from the running interpreter**, so the guard stays true if
  CPython changes its growth policy rather than encoding today's table as a constant.
- Every bound is preceded by an assertion that the record type **exists**, because an upper bound on
  key count passes hardest when there are no rows at all.
- One test checks the two caps are still present **inside `metrics`**, so a future change that
  removed the data rather than the duplicate would fail.

Verified against `main`: the two behaviour-pinning tests fail there — `44 not less than or equal to
42` and `the top-level duplicate is back` — and pass here. Container bytes measured 2,272 → 1,176
per task row.

Both arms of the full python suite were built with `git archive` and diffed by failing test name.

### DIFF
--- a/tools/matrixark_local_adapter_ingest.py
+++ b/tools/matrixark_local_adapter_ingest.py
@@ -1460,8 +1460,14 @@ class _LocalAdapterIngestMixin:
                     "index_write_count": index_write_count,
                     "index_dropped_by_cap_count": index_dropped_by_cap_count,
                     **secondary_index_budget_summary(secondary_index_budget),
-                    "index_cap_per_chunk": MAX_INDEX_TERMS_PER_RESOURCE_CHUNK,
-                    "index_cap_per_fact": MAX_INDEX_TERMS_PER_RESOURCE_FACT,
+                    # The two caps live in `metrics` below, byte-identical, and nothing reads
+                    # either copy at the top level. They are also process constants, so the copy
+                    # said the same thing on every row.
+                    #
+                    # Removing them is worth far more than their two values: CPython sizes a dict
+                    # by KEY COUNT in steps, and this record sat at 44 keys with the next step
+                    # down at 43. At 42 the container costs 1,176 B instead of 2,272 -- 1,096 B
+                    # per row, 48%, for two keys holding "10" and "10".
                     "summary_dirty_hashes": resource_dirty_hashes,
                     "progress": {"stage": "completed", "percent": 100},
                     "metrics": resource_import_metrics,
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Two process constants that duplicated the record own metrics dict were pushing resource_import_task one key over CPython 43-key dict step, costing 1,096 B per row; the test pins the record types nearest a boundary so the next added field fails there instead of silently costing 50-85% more container memory.

PR #846, opened 2026-09-05
Skip a context value that is not UTF-8 instead of losing the pack

bjmeetsfo:
One stored value that is not UTF-8 currently costs a whole ContextPack. The assembly returns an
error, the caller turns that into an empty pack, and a hook that receives one emits `{}` and
exits 0 — so the turn runs with no context and nothing reports a failure.

Binary values are not corruption. Byte payloads are written in a carried raw shape; the older
array-of-JSON-numbers shape, which happened to be UTF-8 by construction, is retired and is no
longer produced. A store therefore legitimately holds values that are not text — embeddings among
them — and any consumer converting a raw record value to a `String` now meets one eventually.

A pack is a best-effort selection of context. A value it cannot interpret belongs outside the
selection, not in the way of it.

## What changes

The non-UTF-8 value is skipped rather than fatal, and the skipped count is reported with the
first offender's key, field, length and a bounded byte prefix, so this cannot become a silent
thinning of context.

## Verified

Same store copy, same request, two builds:

| build | result |
|---|---|
| before | `ok:false` — `stored hash value is not UTF-8 ... (key=…records:000039, field=…143, len=862)` |
| after | `ok:true` — a real ContextPack |

On the live store this moved a Claude hook from returning 3 bytes (`{}`) to completing its
retrieve. A separate ingest-side failure remains and is not addressed here.

* `cargo check --release -p temporalstore-rust --bin matrixark_rust_proxy` on this branch.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -1322,24 +1322,41 @@ fn hgetall_map(engine: &RecordStore, key: String) -> Result<BTreeMap<String, Str
     match response.response {
         CommandResponse::HashEntries { entries } => {
             let mut decoded = BTreeMap::new();
+            let mut skipped_non_utf8 = 0usize;
+            let mut first_skipped = String::new();
             for (field, value) in entries {
-                // Name what choked. Without the key and field this error can only be guessed at,
-                // and it is fatal to context assembly -- the caller returns an empty pack.
-                let value = String::from_utf8(value).map_err(|error| {
-                    let bytes = error.as_bytes();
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. One stored value that is not UTF-8 cost an entire ContextPack, and the caller turns that into an empty pack, so a hook emitted {} and exited 0 with the turn silently uncontextualised. Byte payloads are written in a carried raw shape now that the array-of-numbers shape is retired, so non-text values are normal, not corruption - the pack skips them and reports the count. Verified by A/B on one store copy: previous build ok:false on records:000039, this build ok:true with a real pack.

PR #862, opened 2026-09-05
The idle-commit scan walked the whole store, twice, per query

bjmeetsfo:
`pre_retrieval_idle_commit_flush` runs before every retrieve to find scheduled idle-commit tasks.
It prefers `idle_commit_task_records` on the target and falls back to `read_all()`:

```python
idle_task_records = getattr(target, "idle_commit_task_records", None)
if callable(idle_task_records): records = idle_task_records(scope)
elif callable(read_all):        records = read_all()
```

The **native** reader has offered that method since it was written. The **local adapter never did**,
so the fallback ran — scanning the entire live view **twice per query** to find rows that number in
the handful.

Profiled on a 1 MB corpus at 8,548 records, that flush was **68% of a settled retrieve**.

## The change

The adapter now offers the same contract, backed by an index kept exactly the way the summary-dirty
rows are kept: built from one read the first time, folded forward on append, and dropped whenever
the cache it describes is.

Scope is accepted and not filtered on. The caller already applies `scope_matches` to every row it
considers, and narrowing here too would mean rebuilding the index per scope rather than per cache
generation.

## What it is worth

Measured on **one** store, with each arm given its own copy of it (reading a store rewrites its
snapshot, so a shared directory would have the second arm loading what the first wrote), both
orderings:

| arm | median | best |
|---|---|---|
| main | 29.33, 29.40 ms | 26.19, 26.57 |
| this branch | **10.76, 15.15 ms** | **9.51, 10.25** |

**−56% on the median, −62% on the best**, with the pack identical at 88,397 characters on both arms.

An earlier version of this comparison had each invocation ingest its own corpus. Ingest is not
deterministic — two runs of the *same* arm produced 8,545 and 8,560 records, different packs, and
10 ms against 29 ms. That variance was larger than the effect, so nothing could be read from it; the
shared-store design above is what made the result legible.

## Tests

Five, pinning the index against the scan it replaces in the three states where an index can go
wrong: built cold, folded forward after an append, and rebuilt after the cache it describes is
dropped. Plus one that the pack is unchanged — taking the hook away on the *same* adapter and store
so both sides differ only in which path the flush takes.

The fixture **appends** task rows rather than hoping the corpus carries some: the corpora buildable
here have none, and an index compared against an empty scan is equal for the wrong reason.

That is not hypothetical. It caught a real defect in the first version of this change, where the
fold-forward sat inside another index's guard *and* outside its loop — so it never ran, and only
ever saw the last record. It surfaced as scan 5 against index 4.

Verified against `main`: all five fail there (one assertion, four errors from the missing method) and
all five pass here.

Full python suite on both arms, built with `git archive` from a branch rebased onto current main,
diffed by failing test name: **101 failing names on each arm, identical sets** — nothing on this
branch that `main` does not already have.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -4546,6 +4546,10 @@ class MatrixArkLocalAdapter(_LocalAdapterRetrieveMixin, _LocalAdapterIngestMixin
         # dirty_hash -> the newest summary-dirty or refresh-audit row, so the outstanding set
         # can be answered without reading everything. None until first use.
         self._summary_dirty_index: dict[Any, Json] | None = None
+        #: Pipeline-task rows, kept for the same reason the summary-dirty rows are: the
+        #: pre-retrieval idle-commit flush was scanning the WHOLE live view twice per query
+        #: to find them, and they are a handful of rows in a store of millions.
+        self._pipeline_task_index: list[Json] | None = None
         # model_ref -> {node_hash}, so 'is this node already embedded' does not read the
         # store. None until first use; see _existing_node_embedding_refs.
         self._node_embedding_refs_index: dict[str, set[int]] | None = None
@@ -5051,6 +5055,7 @@ class MatrixArkLocalAdapter(_LocalAdapterRetrieveMixin, _LocalAdapterIngestMixin
             self._read_cache_value_keys = None
             self._read_cache_state_keys = None
             self._summary_dirty_index = None
+            self._pipeline_task_index = None
             self._node_embedding_refs_index = None
             self._read_cache_size = -1
             self._read_cache_mtime_ns = -1
@@ -5300,6 +5305,7 @@ class MatrixArkLocalAdapter(_LocalAdapterRetrieveMixin, _LocalAdapterIngestMixin
                 self._read_cache_value_keys = None
                 self._read_cache_state_keys = None
                 self._summary_dirty_index = None
+                self._pipeline_task_index = None
                 self._node_embedding_refs_index = None
                 self._read_cache_size = -1
                 self._read_cache_mtime_ns = -1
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
Follow-up measurement: this does not just make a settled retrieve faster, it changes how it scales.

Real 1 MB skill documents, queries repeated until the snapshot stops being rewritten so the scan
cost is what is timed rather than the rewrite cliff:

| records | `main` | this branch |
|---|---|---|
| 3,988 | 6.5 ms (1.64 µs/rec) | 4.4 ms (1.10) |
| 7,950 | 13.5 ms (1.69) | 4.3 ms (0.54) |
| 15,874 | 21.8 ms (1.37) | 5.2 ms (0.33) |
| 63,266 | **79.5 ms (1.26)** | **4.0 ms (0.06)** |

`main` holds a flat cost *per record* — **×15.86 records → ×12.15 time**, which is linear in the
size of the store. This branch is **×15.86 records → ×0.90 time**.

Confirmed over a further doubling in a separate run: 17,080 → 4.33 ms, 33,976 → 4.30, 67,720 → 5.91
(×3.96 records → ×1.37).

**The control that matters:** a retrieve can always be made fast by returning less. The pack *grew*
across that range — 88,439 → 91,621 characters — and on the shared-store A/B the packs were
byte-identical between arms at 88,397.

For the workload this is aimed at, 1,000 × 1 MB skills is ~2.35M records. Extrapolating `main`'s
1.26 µs/record puts a settled query around **3 seconds**; the branch stays in single-digit
milliseconds.

It also removes a second cost for free. `read_all` is called **once** per settled retrieve on `main`
and **zero** times here, because the flush was the caller — so `_memory_records_need_expiry_filter`,
which walks every record and measured 3.86 ms at 8,545 records, is no longer reached on this path
either. I had started a separate change to cache that answer; it is unnecessary once this lands, and
I dropped it rather than ship an optimisation whose benefit was already delivered.

PR #880, opened 2026-09-05
Let the per-identity live caches forget who stopped watching

bjmeetsfo:
Two caches behind the status stream are keyed on the identity triple `(key, tenant, account)`:
`_LIVE_SIGNATURE`, which holds what a viewer was last told so an unchanged frame can be skipped, and
`_LIVE_EMBEDDING`, which holds one identity's encoding backlog so its other tabs reuse the read.

**Nothing removed an entry.** `_reset_live_cache` exists, but it is for tests — the only caller is a
test helper. So a worker's resident memory grew with the number of distinct keys that had *ever*
opened a status stream, not the number watching one.

```
2,000 identities open a stream, then leave
  _LIVE_SIGNATURE   2000 entries      _LIVE_EMBEDDING   2000 entries
  retained          4,752 KB          per identity      2,433 bytes
  a later tick      2000 entries      ← nothing sweeps
```

On a production portal an identity is a customer's API key, so this is a per-customer cost that is
never paid back. And every retained byte was already too stale for its own reader to serve: the
signature is dead after one tick, the backlog read after its refresh interval.

### Dropped exactly when it stopped being an answer

The sweep calls each reader's own staleness test with the same arguments — `EVENT_TICK_S` for a
signature, `_embedding_refresh_interval(value)` for a backlog read. So nothing that would have been
served to anyone is evicted, and a viewer that comes back simply misses the way it already did.

It runs from the once-per-tick rebuild, the branch that already runs once per tick for the whole
deployment, so the cost is one pass over the identities seen in the last tick rather than anything
per connection. When the last viewer leaves, that tick's entries stay behind until somebody watches
again — bounded by the viewers of one tick, which is the thing that was unbounded before.

The inflight map gets one more case: a task belonging to a **closed** loop can never be awaited,
because `_embedding_for` checks the loop before waiting on one, so that entry is unreachable rather
than merely stale and nothing would ever have popped it.

### Asserted in both directions

A sweep that simply cleared both dicts would pass every leak test in this file while throwing away
the only reason either cache exists, so the suite asserts the other way too — including that a
second viewer of one identity still reuses the first one's backlog read rather than putting its own
walk of the record log on the backend.

```
do not sweep at all                          -> FAIL x4  (both leaks, + the dead-loop entry)
sweep signatures unconditionally             -> FAIL      a fresh signature survives a tick
sweep embeddings unconditionally             -> FAIL x3   incl. the second viewer reads again
keep embeddings forever                      -> FAIL x2
drop every inflight entry, live loop or not  -> FAIL      a live loop's entry is left alone
```

9 tests. Neighbours: v1_gateway 97, gateway_portal 60, live_strip 38, api_traffic 15,
gateway_events 12, portal_pages 4 — all green.

### DIFF
--- a/tools/matrixark_v1_gateway.py
+++ b/tools/matrixark_v1_gateway.py
@@ -2147,6 +2147,36 @@ def _reset_live_cache() -> None:
     _LIVE_SIGNATURE.clear()
 
 
+def _forget_idle_identities(now: float) -> None:
+    """Drop live-cache entries that their own readers would already ignore.
+
+    Both caches are keyed on the identity triple, and nothing removed an entry. A worker's resident
+    memory therefore grew with the number of distinct keys that had EVER opened a status stream
+    rather than the number watching one -- measured at 2,433 bytes per identity, held for the life
+    of the process, of which every byte was already too stale to be served.
+
+    The tests below are the readers' own staleness checks, called with the same arguments. An entry
+    is dropped exactly when it had stopped being an answer, so nothing that would have been served
+    to anyone is evicted, and a viewer that comes back simply misses the way it already did.
+
+    Run from the once-per-tick rebuild, so the cost is one pass over the identities seen in the
+    last tick rather than anything per connection. When the last viewer leaves, that tick's entries
+    stay behind until somebody watches again -- bounded by the viewers of one tick, which is the
+    thing that was unbounded before.
+    """
+    for identity, entry in list(_LIVE_SIGNATURE.items()):
+        if (now - entry[0]) >= EVENT_TICK_S:
+            _LIVE_SIGNATURE.pop(identity, None)
+    for identity, entry in list(_LIVE_EMBEDDING.items()):
+        if (now - entry[0]) >= _embedding_refresh_interval(entry[1]):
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
Closing: this change is already on `main`, and it got there the wrong way.

I applied the cache-eviction patch to the worktree holding #875 first, to measure the per-identity
footprint, and then committed #875 from that worktree with `git add -A`. So the merge of #875
carried this change as well:

```
0d7939f20 Deliver the pending-restart count the frame already builds (#875)
 tools/matrixark_v1_gateway.py                       |  36 ++
 tools/test_matrixark_the_frame_delivers_what_it_builds.py | 196 ++
 tools/test_matrixark_the_live_caches_forget.py      | 192 ++
```

The code and its tests are intact on `main` — `_forget_idle_identities` is present and called once
per tick, and both suites pass against `main` (frame 7 tests, caches 9 tests). Rebasing this branch
onto `main` drops the commit as "already upstream", which is what surfaced it.

The harm is not in the code, it is that #875 was reviewed against a description that covered one of
the two changes it contained. Recording it here rather than quietly closing.

PR #916, opened 2026-09-05
A queued request no longer outlives the caller waiting for it

bjmeetsfo:
The shared proxy daemon serializes every caller behind one lock and drives the engine over a single
stdin/stdout pipe, so requests cannot interleave. The deadline, though, was measured from the moment
a request **won** that lock, while the caller's own budget started before it queued. The two clocks
disagreed by exactly the queueing time.

The consequences compound:

- the daemon keeps working on requests whose caller has already timed out and gone,
- it holds the single lock while doing that, so the next caller queues behind a result nobody will
  read,
- which makes that caller more likely to be abandoned in turn.

Measured on a live box: run turns one at a time and 8 of 8 serve a full context pack. Overlap them
and turns start coming back with an empty pack after burning ~60 s having done no useful work — the
time went on a call whose answer was thrown away.

This bases the deadline on arrival, so the daemon's clock agrees with the caller's, and drops a
request whose budget is already spent rather than starting it — freeing the lock immediately
instead of spending a minute on an answer with no reader.

It does **not** remove the serialization: one pipe to one engine process is structural. It stops the
queue from manufacturing extra work for itself.

### Tests

`tools/test_a_queued_request_does_not_outlive_its_caller.py`, driving `_call_proxy` against a fake
engine so the assertions are about behaviour rather than timing:

- a request that spent its whole budget queueing is **not** sent to the engine — fails without this
  change (the request is served with `daemon_queue_wait_ms: 2506` against a 2000 ms budget)
- an uncontended request still reaches the engine (control — without it, dropping everything would
  pass)
- a request queued well within its budget is still served (control — queueing stays survivable)

### DIFF
--- a/tools/matrixark_rust_proxy_daemon.py
+++ b/tools/matrixark_rust_proxy_daemon.py
@@ -285,9 +285,35 @@ class RustProxyDaemon:
             if proc is None or proc.stdin is None or proc.stdout is None:
                 return {"ok": False, "error": "rust proxy is not available"}
             try:
+                # The budget runs from ARRIVAL, not from winning the lock. One lock serializes
+                # every caller, so a lock-relative deadline let this process spend a full minute
+                # on an answer whose caller had already timed out -- holding that lock while the
+                # next caller queued behind a result nobody would read, which made the next
+                # caller more likely to be abandoned in turn.
+                budget_s = max(2.0, float(request.get("request_timeout_ms") or 60000) / 1000.0 + 2.0)
+                deadline = started + budget_s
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The shared proxy daemon timed each request from when it won the single lock rather than from when it arrived, so it spent full budgets on answers whose caller had already timed out, holding that lock and pushing the next caller toward the same fate; the deadline now runs from arrival and an already-spent request is dropped instead of started.

PR #939, opened 2026-09-05
The snapshot is a compressed container, base and tail

bjmeetsfo:
The Python durable path was the last one writing its largest artifact as plain text. This stores it
in a compressed binary container, on by default.

## What it costs today

No `zlib`, `gzip`, `zstd` or `msgpack` appeared anywhere in `matrixark_mcp_local_adapter.py`. The
read snapshot was `json.dump(payload, handle, separators=(",", ":"))` at ~1,909 B a record — and
after interning it is highly repetitive, which is exactly what compresses.

Measured on three 1.00 MB skills, 6,793 records:

| codec | ratio | decode |
|---|---|---|
| zlib level 1 | 15.35× | 45 ms |
| **zlib level 6** | **21.08×** | **32 ms** |
| zstd level 3 | 27.53× | 20 ms |

End to end:

```
snapshot    12.97 MB -> 0.74 MB   (17.55x)
per record  1,909 B  -> 109 B
projected   4.49 GB  -> 0.26 GB   at 2.35M records
```

**zlib, not zstd.** `dependencies = []` in `pyproject.toml` is deliberate; zstd is better but not by
enough to earn the project its first runtime dependency. The container carries a codec byte, so
another encoding can be added later without a second format — the shape the engine's served index
(`TSIDX\x01`) already uses.

## Why a new filename

`_load_durable_read_cache` opens the snapshot with `encoding="utf-8"` and catches
`(FileNotFoundError, json.JSONDecodeError, OSError)`. Compressed bytes raise `UnicodeDecodeError` —
a `ValueError`, but **not** a `JSONDecodeError` — so a reader from before this change would crash on
them rather than fall back. That is the failure mode where the hook returns `{}` and exits 0 and an
agent silently runs with no context.

Under `.read-cache.bin` such a reader simply finds no snapshot and re-derives from the log, which it
already does for a missing one. An existing JSON snapshot still loads once the container is enabled
— verified by writing one with the flag off and reading it with the flag on (`source=durable`). **No
migration**: a store is rewritten into the container on its next full write.

Reading never depends on the flag — the loader decides by what the bytes say they are — so a store
written across a flip reads either way and turning it off again is not a one-way door.

## What the format change surfaced

Two gates asked *"does the JSON file exist"* to mean *"is there a base to extend"*. With the base in
a container they answered False, so the tail was never appended and the delta stayed empty. Both now
ask about the snapshot, through `_durable_read_cache_snapshot_path()`. Reverting either fails a test
here.

`_durable_read_cache_snapshot_path()` is **flag-driven** — the form this build writes — not
"whichever form is on disk". Fixtures capture it in `setUp`, before any snapshot exists, and an
accessor that changed its answer once the file appeared would have them watching the wrong path. The
loader still checks both forms explicitly, which is what keeps an older store loading.

## The test surface

Eight files named the JSON snapshot to mean "the snapshot" — existence checks, mtime comparisons, a
shard-name suffix filter, and fixtures. Each now asks through the accessor, and the three that
*inspect* the snapshot's contents go through the module's own decoder, so they exercise the real
decode path instead of assuming a text file.

The sharpest of those: `test_matrixark_the_floor_guards_the_rewrite_only` read the snapshot with
`read_text()` inside `except (FileNotFoundError, ValueError)`. A container raises
`UnicodeDecodeError`, which is a `ValueError`, so the snapshot read as *absent* and every test in
that file reported "no base snapshot was written" — a silent pass into a vacuous state rather than a
failure.

`test_backward_compat_inline_log_reads_under_flag_on` and `test_flag_off_is_byte_identical_to_today`
are red here and **equally red on main** — pre-existing, verified by running main directly.

## The tail is the other half

Compressing the base moved the problem rather than finishing it. The snapshot is two files — a base
and a tail of records appended since it was written — and shrinking the base 17× changed what the
tail is a share *of*:

```
peak tail share of the snapshot, base compressed:  85-99.9%
```

The tail is also the incremental half: it is what grows between folds, on the append path, on every
write. So it gets the same treatment, framed as **one block per appended batch**:

```
codec byte | 4-byte big-endian payload length | zlib(payload)
```

No delimiter and no newline. A compressed payload contains both, and a reader scanning for one would
have to unstuff every record to be sure it had not.

**Per batch, not per record.** The batch is a boundary the writer already holds, so a block costs no
buffering and does not move the moment a record becomes durable. It is also where the compression
is:

| framing | ratio |
|---|---|
| per record | 3.31× |
| **per appended batch** | **15.60×** |

One record carries no dictionary worth the name.

Whole snapshot, base and tail together, over eight ingests interleaved with reads:

```
                              off        on
peak snapshot            567,109     36,960 bytes   (15.34x)
  base                   271,058     15,792         (17.17x)
  tail                    92,496      5,879         (15.73x)
  head                       322        322         (unchanged - it holds counts, not records)
```

**Its own filename again, for the same reason.** The plain reader splits the tail on newlines. Under
`.read-cache-delta.bin` an older reader finds no tail, and the head's `delta_count` no longer
matches, so it re-derives from the log — which it already does for a missing tail. A plain tail
already on disk still loads.

**A torn final block is dropped, not raised.** The tail is derived state and the head records how
many records it should hold, so a short read is caught by the count check that already guards the
plain form. Raising there would turn a torn append into an unreadable snapshot when re-deriving is
the answer.

## Two paths append to the tail

There are two append-only fast paths onto the tail: the one a **write** takes, holding the batch it
just appended, and the one a **read** takes, slicing it out of the record set. Only the first was
taught the block format.

While the tail had a single form that was duplication. With two forms it is a correctness hazard —
the writer extends `.bin`, the reader extends `.jsonl`, the loader reads whichever it prefers, its
count disagrees with the head, and **every read re-derives from the log** with a perfectly good
snapshot sitting beside it. Nothing raises and the answer stays right; only the latency moves, which
is why the suite passed with it in place.

Both now go through `_append_tail_records`, which refuses to append when a tail in the *other* form
is already on disk. Raising there is the answer both callers already have — they catch `ValueError`
and fall through to a full rewrite, which clears both forms and leaves exactly one tail. That is also
what a store crossing the flag does: one rewrite, then one tail.

**Two tests went red on the fix**, which is the sharpest evidence the defect was real. They name the
plain tail file to mean "the tail", their flow reaches the read-side append, and the file they
measured grew *only because that path was writing the wrong form*. With both paths agreeing it is
never written. They ask through `_durable_read_cache_tail_path()` now — the same correction the eight
files that named the JSON snapshot to mean "the snapshot" needed.

The test for the invariant belongs where the fixture reaches the path. Written against the container
file's two-document fixture it never got there at all, and the mutation restoring the bug **survived**
it; against a primed 50-record base it fails. The invariant holds in either flag state, so it sits
with the other incremental-tail tests.

## Tests

`test_matrixark_the_snapshot_is_a_compressed_container.py` — the container is smaller and binary, it
says what it is, a cold read is served from it, an existing JSON snapshot still loads, the decoder
takes either form, and an unknown codec is refused rather than guessed. The flag is a module global,
so each test restores it.

Four more for the block-framed tail: it is smaller than the plain one and the plain one is gone, a
cold read stitches it onto the base and serves `source=durable`, a plain tail already on disk still
loads, and a torn final block is dropped rather than raised. Plus, in the incremental-tail file, both
append paths write one tail and the snapshot keeps serving; and a tail in the other form is refused
rather than joined.

Every mechanism was mutation-tested — **six mutations, six caught**: a block stored rather than
compressed, a torn block raised rather than dropped, a fold leaving the block tail behind, the tail
path answering the plain name always, the helper ignoring the flag, and a tail in the other form
joined rather than refused. Each is caught by a test that names the thing it broke.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -4880,6 +4880,39 @@ class MatrixArkLocalAdapter(_LocalAdapterRetrieveMixin, _LocalAdapterIngestMixin
             return self._durable_read_cache_delta_binary_path()
         return self._durable_read_cache_delta_path()
 
+    def _append_tail_records(self, appended_records: list[Json]) -> None:
+        """Append one batch to the tail, in whichever form this build writes.
+
+        BOTH append paths come through here -- the one a write takes, holding the batch it just
+        appended, and the one a read takes, slicing it out of the record set. They used to write the
+        tail separately. While the tail had a single form that was duplication; with two forms it is
+        a correctness hazard, because the writer would extend the block-framed tail while the reader
+        extended the plain one, the loader would read whichever it prefers, and its count would
+        disagree with the head -- so every read would re-derive from the log with a perfectly good
+        snapshot sitting beside it.
+
+        One block per appended batch. The batch is the boundary the caller already holds, so a block
+        costs no buffering and does not move the moment a record becomes durable -- and it reaches
+        15.6x where per-record framing reaches 3.3x, because one record carries no dictionary worth
+        the name.
+
+        A tail already on disk in the OTHER form is refused rather than joined. Raising here is the
+        answer both callers already have: they catch ValueError and fall through to a full rewrite,
+        which clears both forms and leaves exactly one tail behind.
+        """
+        binary = self._durable_read_cache_delta_binary_path()
+        plain = self._durable_read_cache_delta_path()
+        wanted, other = (binary, plain) if LOCAL_DURABLE_READ_CACHE_COMPRESS else (plain, binary)
+        if other.exists() and other.stat().st_size:
+            raise ValueError("a tail in the other form is on disk")
+        if wanted is binary:
+            with binary.open("ab") as handle:
+                handle.write(_encode_delta_block(appended_records))
+            return
+        with plain.open("a", encoding="utf-8") as handle:
+            for record in appended_records:
+                handle.write(json.dumps(record, separators=(",", ":")) + "\n")
+
     def _durable_read_cache_head_path(self) -> Path:
         """Signature and counts only -- never the records.
 
@@ -5175,17 +5208,7 @@ class MatrixArkLocalAdapter(_LocalAdapterRetrieveMixin, _LocalAdapterIngestMixin
             and delta_count + len(appended_records) <= LOCAL_DURABLE_READ_CACHE_MAX_DELTA
         ):
             try:
-                if LOCAL_DURABLE_READ_CACHE_COMPRESS:
-                    # One block per appended batch. The batch is the boundary the writer already
-                    # holds, so this costs no buffering and does not move the moment a record
-                    # becomes durable -- and it reaches 15.6x where per-record framing reaches 3.3x,
-                    # because one record carries no dictionary worth the name.
-                    with self._durable_read_cache_delta_binary_path().open("ab") as handle:
-                        handle.write(_encode_delta_block(appended_records))
-                else:
-                    with delta_path.open("a", encoding="utf-8") as handle:
-                        for record in appended_records:
-                            handle.write(json.dumps(record, separators=(",", ":")) + "\n")
+                self._append_tail_records(appended_records)
                 write_head(base_count, delta_count + len(appended_records),
                            appended_records[-1])
                 self._durable_read_cache_state = (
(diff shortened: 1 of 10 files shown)


bjmeetsfo:
request still open — so this is a heads-up, not a request to unblock anything, and a post-merge look
is welcome.

What landed: the read snapshot is stored as a compressed binary container, base **and** tail. Base
17.55x, tail block-framed per appended batch at 15.73x, whole snapshot 15.34x. On by default; the
loader decides by what the bytes say they are, so an older store still loads and turning it off is
not a one-way door.

The part worth a second pair of eyes: there are two append paths onto the tail, and only one had
been taught the block format. The writer extended `.bin` while the reader extended `.jsonl`, the
loader read whichever it prefers, its count disagreed with the head, and every read re-derived from
the log with a good snapshot beside it — silently, since the answer stayed right and only latency
moved. Both now go through one helper that refuses to join a tail in the other form. Two existing
tests went red on that fix, which is the evidence it was real: they were measuring a file that only
grew because of the defect.

Six mutations, six caught. Ratchet clean locally and in CI.

PR #968, opened 2026-09-05
A rotated shard is stored compressed

bjmeetsfo:
Once the read snapshot became a compressed container, the **event log** was 94.3% of everything the
local adapter writes — and at the default retention the **rotated** shards are 74.9% of the log.
They compress **11.91×**. That is where the durable bytes now are.

Both changes together, 60 documents on the same corpus, same 424 records:

```
                    main       both
store bytes    1,691,381    231,325     7.31x
per record         3,989        546
```

What is left is the **active** shard at 53.1% — the one this deliberately does not touch.

## Why a rotated shard is the safe one

A rotated shard is a different thing from the active one: it is sealed at rotation, **never appended
to again**, and reached only through `_retained_jsonl_paths()`. The active shard — the one appends
land in, the one recovery reads, the one a person greps — is left exactly as it was, plain text.

```
MASHRD\x01 | codec byte | zlib(the whole shard)
```

A JSONL shard always starts with `{`, so a reader tells the two apart without being told which it
has.

## The rename stays the commit point

Rotation is one atomic `replace`, and it stays that way. Sealing is a **follow-up**: temp write,
fsync, atomic replace. A crash anywhere in it leaves a plain shard, which every reader here still
accepts — so the worst outcome is a shard that was not compressed, never one that was lost.

The order is the whole argument, and **the end state cannot show it**: sealing first and renaming
second produces exactly the same files. The difference is what a crash in between leaves — a
compressed file still named `events.jsonl`, which the next append would extend with plain JSON. So
the test pins the order rather than the result, and the mutation that swaps them fails it.

The temp file is dot-prefixed so it sits outside the `<event log name>*` namespace, for the reason
the read-cache files are: a sidecar picked up under that prefix is replayed as durable history.

## Five copies of one read loop become one

Every shard reader opened the file with `open("r", encoding="utf-8")` and iterated it — five
identical copies. They now go through `_iter_shard_lines`. Consolidating was not optional: five
copies of a format decision are five chances to teach only some of them, which is exactly the defect
the two tail-append paths had in #939.

**The plain branch keeps the buffered text iterator it always had.** Reading it as bytes and decoding
each line in Python instead cost **+67%** on a cold read (72.8 → 121.7 ms median) — a cost that would
have been charged to compression by anyone measuring the two together. The plain shard is the
*active* one, so that is the common case; the seven-byte sniff is one extra syscall per shard.

A sealed shard is decompressed in 1 MB chunks rather than whole. A shard runs to
`MATRIXARK_LOCAL_JSONL_MAX_BYTES` — 64 MB by default — and four of the five readers scan it for one
record type, so materialising it would trade the bytes this saves on disk for the same bytes in
memory.

## Measured

40 documents, rotation at 256 KB, four rounds with the order alternated:

```
shard bytes   985,256 -> 309,287   3.19x
store bytes 1,049,500 -> 373,541
cold read      91.0 ms -> 98.9 ms  +8.7%
```

3.19× rather than 11.91× because the **active** shard is a quarter of the bytes at default retention
and stays plain — that is the trade, not a shortfall.

A **warm** read touches no shard at all; the snapshot answers it. The cold number is what a first
read after a restart pays, against a quarter of the disk bytes to read.

## On, and reversible

The reader takes either form, so a store written with this on still reads with it off, and a plain
shard already on disk keeps loading. No migration: a shard is sealed when it rotates.

## Tests

`test_matrixark_a_rotated_shard_is_sealed.py` — a rotated shard is stored compressed and says what
it is, the active shard stays plain text, sealing does not change the view, a plain rotated shard
still reads, an unknown codec is refused rather than guessed, the reader takes either form, appending
still works after a rotation, the active log is never handed to the sealer, and a shard whose last
line has no newline is read whole.

Five mutations, five caught: a rotated shard left plain, an unknown codec guessed, the sniff always
answering plain, the shard sealed **before** the rename, and the reader dropping whatever follows the
last newline. `decompressobj.flush()` is the documented terminator and returns empty here, so it is
belt-and-braces rather than covered — said plainly rather than counted as coverage.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -163,6 +163,23 @@ _SNAPSHOT_CODEC_ZLIB = b"\x01"
 #: one would have to unstuff every record to be sure it had not.
 _DELTA_BLOCK_CODEC_ZLIB = b"\x01"
 _DELTA_BLOCK_HEADER_BYTES = 5
+#: A SEALED shard: the magic, a codec byte, then the whole shard compressed. A JSONL shard always
+#: starts with `{`, so a reader can tell the two apart without being told which it has.
+#:
+#: Only ROTATED shards are sealed, and that is the whole safety argument: a rotated shard is
+#: finished -- nothing appends to it again -- while the ACTIVE shard stays plain text, so appending,
+#: recovering and reading the log by hand are untouched by this.
+#:
+#: Measured on the corpus this module is tuned for: once the read snapshot became a container the
+#: event log was 94.3% of everything written here, and at the default retention the rotated shards
+#: are 74.9% of the log and compress 11.91x.
+_SHARD_CONTAINER_MAGIC = b"MASHRD\x01"
+_SHARD_CODEC_ZLIB = b"\x01"
+#: On, and reversible: the reader takes either form, so a store written with this on still reads
+#: with it off, and a shard sealed once never needs unsealing.
+LOCAL_JSONL_COMPRESS_SEALED = os.environ.get(
+    "MATRIXARK_LOCAL_JSONL_COMPRESS_SEALED", "1"
+).strip().lower() not in ("0", "false", "no", "off")
 
 LOCAL_DURABLE_READ_CACHE_MAX_DELTA = max(
     1, int(os.environ.get("MATRIXARK_LOCAL_DURABLE_READ_CACHE_MAX_DELTA", "250"))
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
welcome.

What landed: rotated shards are stored compressed. The active shard stays plain text, so appending,
recovery and reading the log by hand are untouched; only a shard that is sealed at rotation and never
appended to again gets compressed. Rotation stays one atomic rename and sealing is a follow-up, so a
crash in it leaves a plain shard rather than a lost one.

Two things worth a second pair of eyes:

**The order is the crash-safety argument and the end state cannot show it.** Sealing before the
rename produces exactly the same files; the difference is only what a crash in between leaves — a
compressed file still named `events.jsonl` that the next append would extend with plain JSON. The
test pins the order rather than the result.

**A regression I introduced and caught.** Routing all five shard readers through one helper also
moved the plain shard off buffered text IO onto per-line Python decoding: +67% on a cold read. The
plain shard is the active one, so that is the common case. It keeps the exact iterator it had, and
only sealed shards take the new reader — measured separately so the cost was not charged to
compression.

Five mutations, five caught. Ratchet clean locally and in CI.

PR #973, opened 2026-09-05
The snapshot base is block-framed, like its tail

bjmeetsfo:
The container decodes the base as **one document**: `zlib.decompress` produces the entire JSON
payload as a single bytes object, and `json.loads` builds the record set from it. Measured in a
process that does nothing but the cold read, the peak that costs tracks the **uncompressed** payload
almost exactly:

| documents | payload | peak off | peak on | extra | extra / payload |
|---|---|---|---|---|---|
| 40 | 3.75 MB | 50,932 KB | 54,060 KB | 3.13 MB | **0.83×** |
| 80 | 7.47 MB | 62,388 KB | 69,908 KB | 7.34 MB | **0.98×** |

On disk the base is 14× smaller, which is the trade. Projected to a thousand 1 MB skills — a 4.49 GB
uncompressed base stored in 0.26 GB — it is a **~4.5 GB transient on every cold read**, which is a
cliff rather than a trade.

## The tail already had the answer

```
codec byte | 4-byte big-endian length | zlib(payload)
```

So the base reuses the **tail's encoder** rather than gaining a second one: a header block carrying
everything except the records, then blocks of 256 records. Both directions are bounded by it —
encoding 60,000 records, one process per arm:

| encode arm | peak RSS | extra | encoded bytes |
|---|---|---|---|
| whole document | 117,796 → 242,060 KB | **+124 MB** | 918,662 |
| block-framed | 117,932 → 120,056 KB | **+2.1 MB** | 961,831 |

The whole-document encoder roughly **doubles** peak memory at write time — it holds the payload as
one string, the compressed result, and the records it came from, all at once. Blocks hold one block.

## Measured

Same corpus, 80 documents, 4,030 records:

```
whole document   62,388 -> 69,908 KB   +7,520 KB    534,667 bytes on disk
block-framed     62,296 -> 62,408 KB     +112 KB    543,560 bytes on disk
```

The transient goes from roughly the payload to essentially nothing, for **1.7%** on disk — each
block compresses against its own dictionary rather than the whole document's. 256 records because
per-block compression reaches 97% of its ceiling by then, so a larger block buys ratio the store
will not notice and costs memory the cold read will.

## A block is an array, and that is where the time was

Newline-delimited records mean one `json.loads` call **per record** — 60,000 of them against one.
Three arms on the same records, block 256, order alternated:

| arm | median decode | bytes | vs whole |
|---|---|---|---|
| whole document | 442.3 ms | 918,570 | — |
| blocks of lines | 621.3 ms | 961,738 | **+40.5%** |
| blocks of arrays | **318.4 ms** | 961,897 | **−28.0%** |

The array is faster than the single-document form as well, at the same size — the compressor sees the
same text either way. Worth finding before the merge: the first version of this change bounded the
memory it set out to bound and would have cost 40% of the decode to do it, with every test passing.

The line codec still **reads**. A tail written by the build that first shipped blocks carries it, and
dropping it would make those tails undecodable rather than merely slower.

## A codec byte, not a new magic

A build from before this raises on an unknown codec, and the loader already answers that by
re-deriving from the log — so an older reader degrades to what it does for any unreadable snapshot
instead of mis-parsing one. A whole-document container already on disk still loads. No migration: a
store is rewritten into blocks on its next full write.

**One producer, not two, on the write side as well.** The blocks stream to the file handle. Joining
them first meant holding every compressed block at once — small on this corpus, ~0.26 GB at a
thousand 1 MB skills, the same shape of transient this format exists to remove. The bytes form the
tests use is a join over the same generator.

## One reader, not two

`_decode_delta_blocks` becomes a wrapper over the streaming `_iter_delta_blocks` the base needs. Two
decoders for one format is how a format decision gets taught to only one of its readers — the defect
the two tail-append paths had in #939.

## Where the assertion goes

The split is asserted on the **encoder**, not on a store. How many records an ingest leaves in the
base — as opposed to the tail beside it — is the fold schedule's business, not this format's. Written
against a store it read as a passing test with the block size set to four, which is exactly the shape
of test that proves nothing.

## Tests

The container's own file now describes the blocked form: the codec byte says blocks, the first block
is the header, the records are split into blocks of the size asked for and round-trip, a
whole-document container still loads, and a blocked snapshot with no header block is refused.

Ten mutations, ten caught: the whole record set in one block, the header block missing, the decoder
keeping only the last block, the blocked snapshot written under the old codec byte, and the streaming
reader stopping after the first block; and on the writer, stopping after the header block, and the
bytes form disagreeing with the streamed one.

## Rebased onto the sealed shards, and tested together

Both changes touch the same file and land in the same declaration block, so this is rebased onto
main with #968 in it and the conflict resolved by keeping both sides — the sealed-shard container
and the blocked-base codec are unrelated declarations that happened to land on the same line. 50
tests across both feature areas pass on the combined state; two PRs that are each green can be red
together, and merging on the strength of two separate green runs would not have checked that.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -162,6 +162,14 @@ _SNAPSHOT_CODEC_ZLIB = b"\x01"
 #: delimiter and no newline -- a compressed payload can contain either, and a reader that scans for
 #: one would have to unstuff every record to be sure it had not.
 _DELTA_BLOCK_CODEC_ZLIB = b"\x01"
+#: A block whose payload is a JSON ARRAY rather than one document per line. Newline-delimited
+#: records cost one json.loads call PER RECORD, and that is the whole of the difference: on 60,000
+#: records the line form decodes in 621 ms, the array form in 318, and one whole document in 442.
+#: So the array is faster than either -- at the same size, since the compressor sees the same text.
+#:
+#: 0x01 still reads. A tail written by the build that first shipped blocks carries it, and dropping
+#: it would make those tails undecodable rather than merely slower.
+_DELTA_BLOCK_CODEC_ZLIB_ARRAY = b"\x02"
 _DELTA_BLOCK_HEADER_BYTES = 5
 #: A SEALED shard: the magic, a codec byte, then the whole shard compressed. A JSONL shard always
 #: starts with `{`, so a reader can tell the two apart without being told which it has.
@@ -180,6 +188,15 @@ _SHARD_CODEC_ZLIB = b"\x01"
 LOCAL_JSONL_COMPRESS_SEALED = os.environ.get(
     "MATRIXARK_LOCAL_JSONL_COMPRESS_SEALED", "1"
 ).strip().lower() not in ("0", "false", "no", "off")
+#: The BASE in the same blocks as the tail. A new codec byte rather than a new magic: a build from
+#: before this raises on an unknown codec, and the loader answers that by re-deriving from the log.
+_SNAPSHOT_CODEC_BLOCKS = b"\x02"
+#: 256 records to a block. Measured on this corpus, per-block compression reaches 97% of its ceiling
+#: by 256 and the decoded transient is one block, so a larger block buys ratio the store will not
+#: notice and costs memory the cold read will.
+LOCAL_DURABLE_READ_CACHE_BLOCK_RECORDS = max(
+    1, int(os.environ.get("MATRIXARK_LOCAL_DURABLE_READ_CACHE_BLOCK_RECORDS", "256"))
+)
 
 LOCAL_DURABLE_READ_CACHE_MAX_DELTA = max(
     1, int(os.environ.get("MATRIXARK_LOCAL_DURABLE_READ_CACHE_MAX_DELTA", "250"))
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
welcome.

What landed: the snapshot base is block-framed like its tail. The container decoded the base as one
document, so the cold-read transient tracked the uncompressed payload almost 1:1 (0.83× at 40
documents, 0.98× at 80) — a ~4.5 GB transient at a thousand 1 MB skills against a 0.26 GB file. Now
+112 KB instead of +7,520 KB, for 1.7% on disk. The write path was worse and is fixed too: the
whole-document encoder roughly doubled peak RSS at 60,000 records (+124 MB), blocks cost +2.1 MB.

The part worth a second pair of eyes: **the first version of this shipped the wrong payload and every
test passed.** Each block held one document per line — the shape the tail already used — which
bounded exactly the memory it set out to bound and cost 40% of the decode, because newline-delimited
records mean one `json.loads` call per record. A block holding a JSON array decodes at 318 ms against
442 ms for a single whole document, at identical bytes. Only a timing arm against the form being
replaced showed it; nothing in the suite could. The old codec byte still reads, so tails written by
#939 do not become undecodable.

Ten mutations, ten caught. Rebased onto main with #968 in it and tested combined — 50 tests across
both feature areas — because two PRs that are each green can be red together.

PR #975, opened 2026-09-05
One rule for reading the embedding provider, instead of six copies of it

bjmeetsfo:
`MATRIXARK_EMBEDDING_PROVIDER` is read at thirteen places. **Six of them dispatch on it** and
spelled out the same expression character for character, each repeating the default:

```python
os.environ.get("MATRIXARK_EMBEDDING_PROVIDER", "deterministic").strip().lower()
```

A seventh would have had to remember all of it, and a default written down six times is one that
disagrees with itself eventually — the drift the numeric-default check exists to catch one layer
up. Now there is one accessor and the six sites call it.

The four spellings that select the in-process encoder (`oss`, `open_source`,
`sentence_transformers`, `sentence-transformers`) were also written out **four times in the same
module**, so a new spelling could reach one dispatch and miss the others. They are a named set now.

### Read per call, never a module constant

A constant resolves once at import and a later write to the environment cannot reach it — the
difference between a setting the portal calls live and one that only looks live. Verified both
ways:

```
default            -> deterministic
after a live write -> voyage      ("  VOYAGE  " normalises)
after removal      -> deterministic
```

### The seven sites left alone are left alone deliberately

- **three** report the *raw* configured string rather than dispatching on it — they answer "what is
  set", not "what will run";
- **two** save and restore the variable around a temporary override;
- **two** normalise with `.strip()` alone.

Folding those into the dispatch rule would change what they do.

### And what is not touched

`embedding_model_name()` exists in **both** modules with different default models
(`all-MiniLM-L6-v2` vs `multilingual-e5-large`). That is an open question needing a backfill
decision, not something to tidy up in passing — so core also keeps its own copy of the provider
set rather than importing a private name from the module it lazily imports.

### Checks

Behaviour is identical: no default moves, no value changes.

| | |
|---|---|
| embedding / vector / encoder / model / retrieval / pack / gateway / config guards | **633 tests, OK** |
| mutation: drop `.strip().lower()` from the new accessor | **FAILED** — so the centralised behaviour is covered, not merely moved |

### DIFF
--- a/tools/matrixark_mcp_core.py
+++ b/tools/matrixark_mcp_core.py
@@ -3484,7 +3484,9 @@ def compact_embedding_vector(vector: list[float]) -> list[float]:
 
 
 def embedding_model_name() -> str:
-    provider = os.environ.get("MATRIXARK_EMBEDDING_PROVIDER", "deterministic").strip().lower()
+    from matrixark_mcp_embeddings import embedding_provider_name
+
+    provider = embedding_provider_name()
     if provider in {"oss", "open_source", "sentence_transformers", "sentence-transformers"}:
         return os.environ.get("MATRIXARK_EMBEDDING_MODEL_PATH") or os.environ.get(
             "MATRIXARK_EMBEDDING_MODEL",
@@ -3497,7 +3499,9 @@ def embedding_model_name() -> str:
 
 
 def embedding_execution_mode_name() -> str:
-    provider = os.environ.get("MATRIXARK_EMBEDDING_PROVIDER", "deterministic").strip().lower()
+    from matrixark_mcp_embeddings import embedding_provider_name
+
+    provider = embedding_provider_name()
     if embedding_fallback_used():
         return "local_hash_embedding_fallback"
     if provider in {"oss", "open_source", "sentence_transformers", "sentence-transformers"}:
--- a/tools/matrixark_mcp_embeddings.py
+++ b/tools/matrixark_mcp_embeddings.py
@@ -148,8 +148,8 @@ def embedding_for_text(text: str, role: str = "passage") -> list[float]:
         cached = _cache_get(cache_key)
         if cached is not None:
             return list(cached)
-    provider = os.environ.get("MATRIXARK_EMBEDDING_PROVIDER", "deterministic").strip().lower()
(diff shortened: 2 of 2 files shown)


bjmeetsfo:
Heads-up from another branch: three tests are red on `main` since this landed, and the ratchet now
reports them as new failures on every PR opened against it. They are not caused by the branches that
report them — I verified the same three fail on a clean `matrixark/main` archive with identical
messages.

```

AssertionError: {'sentence-transformers', 'sentence_transformers', 'open_source', 'oss'}
                not found in []
```

The diagnosis, if it helps: both helpers assert on the encoder's **source text**, not on behaviour.

- `test_matrixark_every_choice_is_explained._sets()` regexes
  `provider in \{("oss"[^}]*)\}` out of `matrixark_mcp_embeddings.py`.
- `test_matrixark_an_unrecognised_provider_name_is_named` calls
  `provider_membership_sets(ENCODER)`, which scans for the same shape.

Consolidating the six copies into one rule removed the inline literal they were matching, so both
scans return `[]` and the membership assertions fail against an empty list. The behaviour they care
about looks intact — it is the pattern they match on that moved.

`test_the_two_that_do_run_an_encoder_still_do` is the floor test for the other one and errors for the
same reason, which is the design working: it exists so `test_local_selects_neither_encoder_path`
cannot pass by vacuity when the parse stops matching.

Happy to point those two helpers at the consolidated rule instead of at a source literal if that is
useful — say the word and I will, otherwise I will leave it to you since it is your area.

bjmeetsfo:
Picked this up in #987 since nothing was in flight for it and a red `main` puts three failures on
every other branch's ratchet. Test-only, two files, 20 lines — no production code touched, because
the consolidation here is right and it is the tests that were reading the spelling of the dispatch
rather than the sets.

The helpers now resolve `provider in _OSS_EMBEDDING_PROVIDERS` to the set that name refers to, and
the regex reads the OSS set where it is declared, exactly as the API set beside it already did.

Since making a scan more permissive is how a test quietly turns vacuous, I checked they still fail
when the sets are actually wrong rather than only that they pass: four controls on the source they
read — `local` added to the in-process set, `local` added to the API set, `voyage` dropped from the
API set, and the encoder no longer dispatching on the named set at any of its four sites — four
caught.

Happy to close #987 if you would rather fix it your own way.

PR #982, opened 2026-09-05
The active log can be written in blocks, one per append batch

bjmeetsfo:
With the snapshot, its tail and the rotated shards all compressed, the **event log** was the last
plain-text artifact — and at 76 KB documents it is **95.3%** of everything this module writes.

## Priced against the batches this code really appends

Recorded by spying on `append_many`: 3 records ×60, 5 ×60, 284 ×29, 292 ×1.

| framing | bytes | ratio |
|---|---|---|
| plain JSONL (today) | 4,853,970 | 1.00× |
| one block per **record** | 3,399,456 | 1.43× |
| one block per **append batch** | 501,381 | **9.68×** |
| one block per 256 records | 485,806 | 9.99× |
| the whole file at once (a ceiling) | 276,828 | 17.53× |

A block per append batch reaches **97%** of what fixed 256-record blocks would, and it costs **no
durability**: the batch is already the unit that gets acked together, so a crash loses exactly what
it loses today. Per-record framing would be durability-free for the same reason and is worth almost
nothing — one record carries no dictionary.

## End to end

40 documents of 76 KB, same corpus, 7,224 records:

```
                                   store        B/record   vs source
before the container work     28,177,927           3,893       9.06x
container + tail + shards     14,936,207           2,068       4.80x
and the log in blocks          1,440,350             199       0.46x
```

The store is now **smaller than the text it was built from**, where it used to be nine times larger.

## Off by default, and here that is not a formality

Every other change in this family compresses something **derived** — a snapshot, a tail, a sealed
shard — which can be rebuilt from the log if it is ever unreadable. This changes the log itself, and
a process appending plain JSON to a block-framed log corrupts it.

So **the form is taken from the file, not from the flag**. An existing log keeps its own form, only a
fresh one adopts the configured one, and turning the flag off again leaves every log readable because
nothing rewrites one. That also means a log written by hand — which is how a good many fixtures are
built — stays plain and keeps working untouched.

## Lines, not the array the snapshot uses

Deliberately different, and the reason is this file rather than that one. The log is parsed
line-by-line *today*, so a line payload is neutral where an array would be a second change to measure
separately; and two of the five shard readers scan for a single record type with a substring test,
which a line payload keeps and an array payload would turn into parsing every record.

## One writer

Both append paths go through `_append_records_to_log`. They used to encode and write separately —
duplication while there was one form, and the same hazard the snapshot tail had in #939 when only one
of its two writers knew about blocks.

Rotation is decided on the bytes actually written, so a block log holds proportionally more history
before rotating. Deliberate: retention stays a disk budget, and compressing the log spends it on more
history rather than on less disk.

## Where losslessness is checked

On a **round trip**, not end to end. A served-record count moves by a dozen between runs because
`context_index` postings compact differently — it moved, and then moved back, while this was being
written — so an end-to-end count would have been reading the compactor, not the format. The raw log
records are identical in both arms (2,919, identical histograms).

## Why not a shared dictionary

Per-batch blocks reach 9.69x where compressing the whole file at once reaches 17.55x, and the obvious
explanation is that every block pays to rediscover the same record shape. zlib takes a `zdict` that a
file could carry once in its header, so this was measured rather than assumed:

| framing | bytes | ratio |
|---|---|---|
| per batch, no dictionary | 500,920 | 9.69x |
| per batch, `zdict` = batch 0 | 480,615 | 10.10x |
| per batch, `zdict` = 32 KB sample | 504,075 | **9.63x** |
| per batch, level 9 | 482,968 | 10.05x |
| the whole file at once (a ceiling) | 276,633 | 17.55x |

**At most 4.2%, and the bigger dictionary is worse than none** — the 32 KB sample costs more carried
in the file than it saves. The distance to the ceiling is not the dictionary: the large batches are 284
records and build a perfectly good window of their own, and what a single stream exploits is
redundancy ACROSS batches, which no per-block dictionary can reach without giving up the property
that makes this free. Level 9 buys 3.7% for more CPU on every append.

So no dictionary, and level 6 — this is within 4% of what a durability-free design can do.

## Tests

`test_matrixark_the_log_is_written_in_blocks.py` — the log says what it is and is much smaller, holds
the same records either way, writes one block per append batch, a plain log already on disk stays
plain, a block log stays blocks when the flag goes off, a torn final block drops only its own records
while the prefix survives, and a cold read serves the block log.

Four mutations, four caught: the form taken from the flag instead of the file, one block per record
instead of per batch, a torn final block raising instead of stopping, and the magic never written.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -183,6 +183,18 @@ _DELTA_BLOCK_HEADER_BYTES = 5
 #: are 74.9% of the log and compress 11.91x.
 _SHARD_CONTAINER_MAGIC = b"MASHRD\x01"
 _SHARD_CODEC_ZLIB = b"\x01"
+#: A shard written as a STREAM of blocks rather than one compressed blob. This is the form the
+#: ACTIVE log can take: a sealed shard is finished and compresses whole, but a log is appended to,
+#: and a stream of self-describing blocks is appendable where a single deflate stream is not.
+_SHARD_CODEC_BLOCKS = b"\x02"
+#: Off by default. This is the only change in this family that alters the DURABLE SOURCE OF TRUTH
+#: rather than something derived from it, and the failure it guards against is a real one: a process
+#: appending plain JSON to a block-framed log corrupts it. The form is taken from the file on disk
+#: rather than from this flag (see _log_append_form), so turning it on affects only logs this build
+#: creates, and turning it off again leaves every existing log readable.
+LOCAL_JSONL_BLOCK_LOG = os.environ.get(
+    "MATRIXARK_LOCAL_JSONL_BLOCK_LOG", "0"
+).strip().lower() not in ("0", "false", "no", "off")
 #: On, and reversible: the reader takes either form, so a store written with this on still reads
 #: with it off, and a shard sealed once never needs unsealing.
 LOCAL_JSONL_COMPRESS_SEALED = os.environ.get(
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
The ratchet is red here and **it is not this branch**. It reports three new failures, all in the
encoder-provider area this change does not touch:

```
test_local_selects_neither_encoder_path
test_the_in_process_set_is_a_dispatch_set_in_the_encoder
test_the_two_that_do_run_an_encoder_still_do
```

All three fail identically on a clean `matrixark/main` archive — same names, same messages — so this
branch adds nothing:

```
```

They arrived with #975, which consolidated six copies of the provider check into one rule. Both
failing helpers assert on the encoder's *source text* rather than on behaviour — one regexes
`provider in \{("oss"[^}]*)\}` out of `matrixark_mcp_embeddings.py`, the other scans for the same
shape — so removing the inline literal makes both return `[]`. Reported there with the diagnosis.

I am not banking them. The ratchet's contract is that it fails when a change **adds** a failure, and
banking would drop three tests from the gate to make an unrelated branch look green.

Everything else here is green, including `cargo check / build / test`. The 58 tests this change
touches pass locally, and its own nine pass with four of four mutations caught.

bjmeetsfo:
welcome.

What landed: the active log can be written as a stream of blocks, one per append batch. With the
snapshot, its tail and the rotated shards already compressed, the log was the last plain-text
artifact and 95.3% of everything this module writes at 76 KB documents.

**9.68x, and it costs no durability** — the append batch is already the unit that gets acked
together, so a crash loses exactly what it loses today. Per-record framing, the other durability-free
option, is worth 1.43x; fixed 256-record blocks would reach 9.99x, so the free choice gets 97% of the
paid one. End to end on 120 documents, 21,686 records: the store goes 84.5 MB to 4.3 MB, 3,896 to 198
bytes a record — smaller than the text it was built from.

**Off by default, and here that is not a formality.** Every other change in this family compresses
something derived, which can be rebuilt from the log. This changes the log itself, and a process
appending plain JSON to a block-framed log corrupts it. So the form is read from the file's magic,
not from the flag: existing logs keep their own form, only a fresh one adopts the configured one,
hand-written fixtures stay plain, and turning the flag off leaves every log readable.

**A shared dictionary is the obvious next idea and it does not pay.** Measured rather than assumed:
at most 4.2%, and the naive 32 KB sample is *worse* than none because it costs more carried in the
file than it saves. The distance to the whole-file ceiling is cross-batch redundancy, which no
per-block dictionary reaches without giving up the property that makes this free.

**Losslessness is asserted on a round trip, not end to end.** A served-record count moved by 14
between arms and then moved back — `context_index` postings compact differently run to run. The raw
log records are identical in both arms (2,919, identical histograms), so an end-to-end count would
have been reading the compactor, not the format.

Nine tests including the rotation seam where the block log, the rename and the sealer meet; four of
four mutations caught. The ratchet was red here on three encoder tests inherited from `main`, which
#987 fixed — this was rebased onto that and is green on its own merits.

PR #988, opened 2026-09-06
Serve the last context pack when a turn cannot build one

bjmeetsfo:
Three problems measured on a live one-box, all of which ended the same way: the agent was told it
has **no history**.

### A turn that could not build a pack emitted `{}`

`{}` is indistinguishable from "there genuinely is no context", so an agent proceeded as though it
had none — wrongly, and silently. Both hooks now remember each pack that loads and serve the last
good one when a turn would otherwise return nothing, labelled in band:

```
[prior context: the store did not answer this turn; this is the last pack that loaded, 6 min ago]
```

and recorded as `context_pack_source: "previous_pack"` with `context_pack_age_s`. A run of
`previous_pack` in the reports is an alarm that retrieval is failing — it is mitigation, not a cure.

### A retrieve that raised killed the turn before the fallback could run

Retrieve was left unwrapped on the reasoning that a failed read *is* the failure. That held only
while there was nothing better to serve. The effect was perverse: a **loud** policy error
("backend-native ContextPack assembly is required…") produced exactly the same useless `{}` as a
silent one, because `main()` died before reaching the emission path. The failure is now recorded
under `retrieve_error` and the turn continues.

### Codex was protected by neither, because it runs a separate entry point

```
matrixark_claude_hook.sh       -> tools/matrixark_agent_hook.py
matrixark_codex_onebox_hook.sh -> ... -> tools/matrixark_codex_hook.py
```

Measured over 8 turns on one box, same store, same minute: **Claude 0 empty, Codex 3 of 4 empty.**
A fix to one hook covers one agent. Both now share `tools/matrixark_hook_pack_cache.py` rather than
keeping a copy each, so two cache-key schemes cannot drift until one agent serves the other's pack.

### The proxy daemon was holding 2.9 GB

A bridge that stores nothing sat at 2.9 GB RSS while the engine it fronts held 1.1 GB — one thread,
ten fds, no accumulating containers. Every payload is encoded and decoded whole in that process, so
a large one leaves arenas CPython has freed but never returns. Measured with a 105 MB document:

| stage | RSS |
|---|---|
| peak | 345 MB |
| after `del` + `gc.collect()` | 124 MB |
| after `malloc_trim(0)` | 12 MB |

The daemon now returns arenas after a large payload, bounded to payloads above
`MATRIXARK_PROXY_DAEMON_TRIM_BYTES` (default 8 MiB, `0` disables) because `malloc_trim` walks the
arenas and doing it per request would spend time on the hot path reclaiming nothing.

### Measured effect

Live box, one soak, split at the moment the Codex fixes landed:

| | before | after |
|---|---|---|
| all agents | 38% turns with no context (3/8) | **0% (0/17)** |
| codex | 75% (3/4) | **0% (0/8)** |

Daemon RSS, compared at the same process age: **2,929 MB → 745 MB**, oscillating in a 400–1,000 MB
band instead of climbing.

Latency also improved (p50 181.8 s → 92.9 s) but box load fell over the same window, so that is
**not** attributed to this change.

### Tests

- `test_a_turn_that_cannot_build_a_pack_serves_the_last_one.py` — round trip, stale refused,
  nothing-cached refused (must not invent context), workspace isolation, agent isolation, no torn
  write, unparseable age bound falls back rather than crashing the hook, label carries the age
- `test_a_retrieve_that_raises_is_recorded_not_fatal.py` — the turn survives and the reason is
  recorded, with a control proving a working retrieve still passes straight through
- `test_freed_arenas_go_back_to_the_os.py` — trims above the threshold and **not** below it,
  boundary, disable switch, no glibc, raising libc

The controls are deliberate: without them, "return `{}` always" and "trim always" would pass.

### DIFF
--- a/tools/matrixark_agent_hook.py
+++ b/tools/matrixark_agent_hook.py
@@ -13,6 +13,10 @@ from __future__ import annotations
 
 import argparse
 import json
+try:
+    from tools import matrixark_hook_pack_cache as _pack_cache
+except ImportError:  # running from inside tools/, as the hooks do
+    import matrixark_hook_pack_cache as _pack_cache
 import os
 import subprocess
 import sys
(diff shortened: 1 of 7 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. A hook turn that could not build a context pack emitted {}, which reads to an agent as "you have no history"; both hooks now serve the last pack that loaded, labelled with its age, and Codex is covered too (it runs a separate entry point, so the earlier fix reached only Claude). Measured on one box: turns with no context 38% -> 0% overall, 75% -> 0% for Codex.

PR #995, opened 2026-09-06
The vector can be stored as packed f32

bjmeetsfo:
A 384-dimension vector written as a JSON array costs **20.96 bytes per dimension**. Once the encoder
is a real model rather than this box's 32-dimension token hash, the digits of a float are most of a
record. Packed little-endian f32 in base64 is 5.33.

Priced **after** compression, because the log writes zlib blocks now and a saving compression already
takes is not a saving. 256 vectors of 384 dimensions, one block:

| form | plain | zlib-6 | vs JSON compressed |
|---|---|---|---|
| JSON array of floats | 2,061,356 | 906,442 | 1.00× |
| **f32 base64** | 537,775 | 397,632 | **2.28×** |
| int16 + scale | 275,631 | 202,646 | 4.47× |
| int8 + scale | 144,559 | 102,612 | 8.83× |

## On this environment it is a wash — 1.0%

These vectors are 32 dimensions from a token hash, so most elements are exactly `0.0` and JSON spends
three characters on them. At *this* width the packed form is 5.33 bytes a dimension against 4.09 for
the JSON, which is **worse**. The 2.28× is a property of production-width model vectors, measured on
synthesised ones — not something this corpus demonstrated. Stated plainly rather than quoted as if it
were.

```
binary_vectors=off   log 642,404   store 705,346
binary_vectors=on    log 635,602   store 698,242     -1.0%
served vector digest identical both ways
```

## f32, not int8

f32 is **exact**, so there is no recall question to answer and no quality bar to clear. int8 is worth
8.83× and is a separate decision needing the bar met on the corpus that will be served: on 2,000
random unit vectors the reconstruction cosine passes comfortably (**min 0.999938**, bar 0.999) while
recall@1 drops **0.0100** against a bar of 0.005. Random vectors in 384 dimensions are near-orthogonal
— the hardest case there is, and not what an encoder produces — so that measurement belongs on real
vectors, not this fixture.

## A storage representation, not a codec

Packed on the way into the log, expanded at the read choke point, so no reader sees the packed form
and no record contract changes — the same shape as the interned bundle beside it. The unpacking runs
**before** `expand_interned_records`' fast path, which returns early when nothing is interned; a
packed vector has nothing to do with interning and would be skipped by it.

## Two defects this had while it was being written

**A warm read and a cold read disagreed.** Packing on the way to the *log* only left the read cache
and the snapshot holding the original float64s, so a warm read answered `-0.408248` where a cold read
that re-derived from the log answered `-0.40824800729751587`. The same store, two answers, decided by
which path served it. Storing f32 has to mean holding f32, so the rounding happens where the record
is made.

**The rounding went in on one of the two append paths.** They build their sanitized list in separate
comprehensions and the other spells the same loop with a different variable name, so it was missed —
the one-path guard defect again. It now lives in `_sanitize_jsonl_record`, which both run per record,
above its early return.

Each is now a test, and each fails if the fix is reverted.

## Off by default

Every other switch in this family degrades to re-deriving from the log. This one degrades to a
quietly worse answer: a reader from before this change finds no `vector` on a packed record and
carries on without one — lost recall, no error. The flag chooses what to **write**; unpacking is
unconditional, so a store written with it on still reads with it off.

## Tests

Ten: the round trip is exact, a packed vector is smaller per dimension at model width, the log holds
the packed form while the served view is unchanged, a packed log reads with the flag off, expansion
unpacks even when nothing is interned, a damaged vector loses the vector and not the record, packing
leaves non-float-lists alone, the flag off writes the JSON array, and the two regression tests above.

Five mutations, five caught.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -6,8 +6,10 @@
 from __future__ import annotations
 
 from contextlib import contextmanager
+import base64 as _base64
 import copy as _copy
 import hashlib as _hashlib
+import struct as _struct
 import queue as thread_queue
 import zlib
 from typing import Any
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
The ratchet failure here was a **stale base**, not this change.

It reported one new failure, `test_heartbeat_only_rendered_context_does_not_emit_additional_context`,
which passes in isolation on both this branch and `main` — so isolation said nothing and only the
full-suite condition could answer it.

Chasing it down:

- Reproduced locally in the full suite on this branch, so it was real and not CI flake.
- A clean `main` worktree runs the full suite with **no new failures**.
- This branch was cut before #997 landed. Rebased onto current `main` and re-run: **no new
  failures**.

So #997 fixed it and this branch was carrying the old base. Nothing in the change is implicated —
it is a no-op with its flag off, which is the default.

Worth recording how the first control went wrong: I first built the `main` comparison from
`git archive`, which produces a tree with no `.git`. A dozen tests in this suite enumerate the tree
through git, so they all failed for want of a repository rather than for want of what they scan —
twelve fake failures, and the real one absent. Redone as a proper worktree checkout it is clean.

bjmeetsfo:
awareness, no action needed.

What landed: the vector can be stored as packed f32 instead of a JSON array of floats. A
384-dimension vector written as JSON costs 20.96 bytes per dimension; packed f32 in base64 is 5.33,
which is **2.28x after block compression** — the number that matters now the log writes zlib blocks.

Three things worth knowing:

**On this environment it is a wash — 1.0%.** These vectors are 32 dimensions from a token hash, so
most elements are exactly `0.0` and JSON spends three characters on them; at that width the packed
form is actually worse per dimension. The 2.28x is a property of production-width model vectors,
measured on synthesised ones. Said in the PR, the commit and the test file so nobody reads it as
something this corpus demonstrated.

**f32, not int8.** f32 is exact, so there is no recall question. int8 is worth 8.83x and is a
separate decision needing the bar met on the corpus that will be served — on random 384-dim unit
vectors the reconstruction cosine passes (min 0.999938) while recall@1 drops 0.0100 against a bar of
0.005, and near-orthogonal random vectors are the hardest possible fixture rather than a
representative one.

**Two defects it had while being written, each now a test.** A warm read and a cold read disagreed to
the bit, because packing happened only on the way to the log while the cache and snapshot kept the
original float64s. And the fix first landed on one of the two append paths — the other spells the
same comprehension with a different variable name — so the rounding now lives in
`_sanitize_jsonl_record`, which both run per record.

Off by default: a reader from before this change finds no `vector` on a packed record and carries on
without one, which is lost recall and no error. Unpacking is unconditional, so a store written with
it on still reads with it off.

Note on the ratchet here: it went red on a stale base and is green after rebasing onto current main —
#997 had already fixed the test it reported. Details in the comment above.

PR #1006, opened 2026-09-06
The JSON POST header was still built per request

bjmeetsfo:
#936 replaced a per-request `format!` for the content-type line with a borrowed constant — in `request_bytes_with_options`. The **other** function that builds the same line, `post_json_with_options_and_headers`, was left as it was, and it is the one every JSON POST goes through: the client's execute and batch forwards, and meta sync.

Its own comment already said what that costs:

> `format!` here cost two on every request the proxy and the client make

Both allocations were still being paid on that path.

## No measured delta is claimed

Neither end-to-end row in `what_a_proxy_request_allocates` exercises this line — the `/execute` row has no backend and fails before forwarding, and the stub-backed row goes through the context path, which uses the byte variant #936 already fixed. I checked rather than assumed, and the rows do not move.

So this stands on being the same fix as #936 applied to the sibling it missed, not on a number I did not take.

## The test asserts bytes, not literals

It checks the header ends in **13, 10** rather than comparing one source literal to another.

Writing this patch, an escaping slip turned the header into a real newline in the source. A first version of the test compared two literals and passed happily — both sides were wrong in the same way. A header ending in a bare LF is a broken HTTP frame, not a slower one, so the assertion is on the bytes that actually go on the wire.

### DIFF
--- a/crates/temporalstore-rust/src/http.rs
+++ b/crates/temporalstore-rust/src/http.rs
@@ -351,7 +351,15 @@ pub fn post_json_with_options_and_headers<Req: Serialize, Res: DeserializeOwned>
     options: HttpRequestOptions,
 ) -> Result<Res, HttpError> {
     let body = serde_json::to_vec(request)?;
-    let headers = format!("Content-Type: application/json\r\n{extra_headers}");
+    // Borrowed when there is nothing to append, which is every caller but meta sync. The
+    // sibling of this line in `request_bytes_with_options` was given a `Cow` and this one was
+    // not, so the two allocations `format!` costs stayed on the JSON POST path -- the path the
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Now measurable. #1007 adds a row that actually forwards a command to a stub backend — until then no row in this benchmark exercised the client send path, which is why I said above that no delta was claimed.

| POST /execute, backend ANSWERS | allocs | bytes |
|---|---|---|
| with this change | **24** | **1914** |
| with the `format!` | 25 | 1979 |

One allocation and 65 bytes per forwarded command. The context row is unchanged at 37, confirming it takes the byte variant #936 already fixed — which is what I reasoned above, now checked rather than argued.

PR #1009, opened 2026-09-06
A rolled-up event drops the postings the summary replaced

bjmeetsfo:
Lever 1 of the index growth bound is the principled one: once an event has been rolled up into a
summary, its per-event postings are redundant with the summary's own, so the summary emits an
`index_compact` tombstone naming exactly the events it covers and the serving sweep drops those
postings. The index stays dense for recent turns and sparse-summarized for old ones, and old content
stays retrievable through its summary.

**Both halves existed and neither was called.** Each is identity without the other, so wiring one
alone would have changed nothing while looking like a fix: `index_compaction_tombstone` returns
`None` when there is nothing to compact, and `sweep_index_compaction` returns its input unchanged
when the log carries no tombstone. A store that has never rolled up a summary is byte-identical
either way.

## It fixes a test that is red on main

`test_rollup_prunes_event_postings_without_losing_recall` asserts this end to end: with compaction
on, every rolled-up event's postings must be gone, the summary postings must survive, and **not one
fact may be lost**. It has been red on `main` (`4 != 0`). It passes here.

Measured on that test's own corpus — 15 turns over 5 facts:

```
event postings in the raw log      60
event postings served, before       4   (sweep in the wrong place)
event postings served, after        0
```

## Order, twice — both found by measurement, not reasoning

**The sweep runs before the posting rebuild.** `sweep_index_compaction` drops a posting only when a
tombstone that *follows* it covers every ref — "a tombstone only removes records that PRECEDE it" —
and `compact_latest_context_state_records` **rebuilds** `context_index` postings. Sweeping after it
hands the sweep fresh rows positioned after the tombstones, which it cannot match: **4 of 60
survived** that way, every one with its ref already covered. That looked like a coverage bug and was
an ordering bug.

**The sweep runs before the budget**, which is the module's own order:

> Lever 1 has already removed every posting that was purely redundant, so whatever a cap evicts on
> top of it is a live lookup path.

Bounding first would spend the budget evicting postings the sweep removes for free.

## Where the tombstone is emitted

Where the summary is written — the only place that knows both the summary hash and the events it
covers, and the record already carries `source_event_ids`. It is scoped by the summary's own scope,
so a tombstone in one scope cannot drop another scope's postings.

## Scope

Two files, 39 lines. No new test file: the behaviour already had an end-to-end test, and it was red.
Adding a second test asserting the same thing would not have made it more true.

Follows #1002, which wired levers 0, 2 and 3 from the same module.

### DIFF
--- a/tools/matrixark_local_adapter_summaries.py
+++ b/tools/matrixark_local_adapter_summaries.py
@@ -982,6 +982,29 @@ class _LocalAdapterSummariesMixin:
                     "updated_at_ms": refreshed_at_ms,
                 }
                 self.append(summary_record)
+                # Lever 1. The events this summary rolls up no longer need postings of their own:
+                # the summary carries its own, and those are the surviving recall path. The
+                # tombstone names exactly the events covered, and the serving sweep drops their
+                # postings.
+                #
+                # `index_compaction_tombstone` returns None when there is nothing to compact, so a
+                # summary with no source events appends nothing and the log stays byte-identical to
+                # what it was before this.
+                try:
+                    from tools.matrixark_index_growth_bound import (
+                        index_compact_on_summary_enabled, index_compaction_tombstone)
+                except ImportError:  # Direct script execution from tools/.
+                    from matrixark_index_growth_bound import (
+                        index_compact_on_summary_enabled, index_compaction_tombstone)
+                summary_scope = summary_record.get("scope")
+                if index_compact_on_summary_enabled(summary_scope):
+                    compaction = index_compaction_tombstone(
+                        source_event_ids=source_event_ids,
+                        scope=summary_scope if isinstance(summary_scope, dict) else None,
+                        summary_hash=summary_hash,
+                    )
+                    if compaction is not None:
+                        self.append(compaction)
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
The ratchet caught the bookkeeping this change owes, which is the guard working rather than a
problem with it.

`test_a_gate_that_got_wired_is_struck_off_the_list` failed with:

```
these are listed as known-unwired but now have callers: ['index_compact_on_summary_enabled'].
Remove them from KNOWN_UNWIRED -- a list of known problems that is allowed to go stale hides the
fix as effectively as it hid the defect.
```

Which is exactly right: the summary path now asks that gate before emitting a compaction tombstone,
so it has a production caller and no longer belongs on the list. Struck off, and the count above the
list moves from ten to nine.

Worth noting what this says about the guard's shape. `KNOWN_UNWIRED` is keyed on `*_enabled` gates,
so it tracked this one and could tell me I had wired it — while `enforce_secondary_index_bounds`,
`sweep_index_compaction`, `max_index_records_per_scope` and `index_hard_ceiling` are outside that
census entirely and could be wired or unwired without the guard noticing either way. The half it
covers, it covers well.

PR #1012, opened 2026-09-06
Packed vectors become the default; the block log does not

bjmeetsfo:
Two binary storage forms shipped **off**, both because they change the *durable* form where every
other change in this family compresses something derived. Turning both on and running the suites
separated them — and that separation is the point of this change.

## Vectors default on

A 384-dimension vector costs **20.96 bytes a dimension** as JSON digits and **5.33** packed — 2.28×
after block compression — and f32 is **exact**, so there is no recall question to answer.

Verified with no environment variable set:

```
default:                     115 stored lines packed, 0 JSON arrays
MATRIXARK_LOCAL_BINARY_VECTORS=0:  0 packed, 115 JSON arrays
served vector digest:        4e4060d978c11458  — identical both ways
```

The identical digest is what makes it a representation rather than a codec.

## The block log stays off, and the reason is demonstrated

Turning it on breaks two tests:

- `test_a_writer_in_another_process_does_not_corrupt_the_view`
- `test_both_append_paths_write_the_same_tail`

Both append raw JSON lines to a log **this module created** — which is exactly what a writer that is
not this module does — and a plain append onto a block-framed log corrupts it. **A store this build
creates has to stay appendable by something that writes plain JSONL**, and defaulting blocks on takes
that away.

Six other failures were fixture arithmetic (a 10× smaller log stops reaching a byte-threshold
rotation) and would have been fair to fix. This one is not, so the flag keeps its default and the
comment now carries the two test names as the reason.

It remains a per-deployment choice for anyone who knows their log has a single writer.

## The reversibility both rely on is measured

Same store, read twice with the reader flag flipped:

```
store written   log head            records  vectors digest
on, read on     4d41534852440102        144       82 a642e26d7c8579ab
on, read off    4d41534852440102        144       82 a642e26d7c8579ab
off, read on    7b227265636f7264        144       82 3ce90d635ad1ac73
off, read off   7b227265636f7264        144       82 3ce90d635ad1ac73
```

Byte-identical both ways, in both directions — the flag chooses what to **write** and never what can
be read.

> An earlier version of that check compared four **separate** stores and produced four different
> digests, which proved nothing: every store derives its own identities, so only reading the *same*
> bytes twice answers the question.

Nothing on disk is converted either way. A JSON float array keeps being read beside packed ones, per
record, so a store crossing this flag is a mixture and reads correctly.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -189,11 +189,17 @@ _SHARD_CODEC_ZLIB = b"\x01"
 #: ACTIVE log can take: a sealed shard is finished and compresses whole, but a log is appended to,
 #: and a stream of self-describing blocks is appendable where a single deflate stream is not.
 _SHARD_CODEC_BLOCKS = b"\x02"
-#: Off by default. This is the only change in this family that alters the DURABLE SOURCE OF TRUTH
-#: rather than something derived from it, and the failure it guards against is a real one: a process
-#: appending plain JSON to a block-framed log corrupts it. The form is taken from the file on disk
-#: rather than from this flag (see _log_append_form), so turning it on affects only logs this build
-#: creates, and turning it off again leaves every existing log readable.
+#: Off, and it stays off -- the hazard is demonstrated rather than argued. Turning it on breaks
+#: `test_a_writer_in_another_process_does_not_corrupt_the_view` and
+#: `test_both_append_paths_write_the_same_tail`: both append raw JSON lines to a log this module
+#: created, which is what a writer that is NOT this module does, and a plain append onto a
+#: block-framed log corrupts it. A store this build creates has to remain appendable by something
+#: that writes plain JSONL, and defaulting this on takes that away.
+#:
+#: Everything else about it is safe and measured -- an existing plain log stays plain because the
+#: form is read from the FILE, and a store written with this on reads byte-identically with it off
+#: (144 records, 82 vectors, same digest). It is a per-deployment choice for anyone who knows their
+#: log has a single writer, not a default.
 LOCAL_JSONL_BLOCK_LOG = os.environ.get(
     "MATRIXARK_LOCAL_JSONL_BLOCK_LOG", "0"
 ).strip().lower() not in ("0", "false", "no", "off")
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
The ratchet failure here is this change, and the fix is in the test that asserts it.

`test_backward_compat_inline_log_reads_under_flag_on` compares `_raw_disk_records` — documented as
"the exact durable bytes, decoded but **not** expanded" — against `_read_raw_records`, which expands.
Those were the same thing while expansion had one job. With a packed vector the disk record carries
`vector_f32` and the expanded one carries `vector`, so the assertion compares a packed vector with an
unpacked one. That is the storage representation working, not interning breaking.

The test is about **interning** backward compatibility, so both sides now go through the same vector
unpacking and what it asserts is unchanged.

Two things went wrong in diagnosing it, both worth recording:

**Isolation said the wrong thing.** The test fails standalone on `main` too, so an isolation check
calls it pre-existing and moves on. It is reported as *new* because in the full suite it passes on
main and fails here — standalone it dies early on an unrelated dict-record assertion and never
reaches the one that matters.

**My first local reproduction ran on the wrong branch.** One worktree, several branches through it: I
had checked out the lever-1 branch to strike a gate off `KNOWN_UNWIRED`, and ran the ratchet believing
I was here. It reported "no new failures" and I nearly took that as disproving CI. An unrelated probe
printing `LOCAL_BINARY_VECTORS default: False` — impossible on this branch — caught it. On the correct
branch the failure reproduces exactly as CI described, and 77 of the fixture's records carry a packed
vector where none carried a JSON array.

bjmeetsfo:
awareness, no action needed.

What landed: packed f32 vectors become the default. The block log deliberately does **not**.

I set out to turn both on. Turning both on broke eight tests, and they split — six were fixture
arithmetic (a 10x smaller log stops reaching a byte-threshold rotation, fair to fix) and two were a
real argument against the default:

    test_a_writer_in_another_process_does_not_corrupt_the_view
    test_both_append_paths_write_the_same_tail

Both append raw JSON lines to a log this module created, which is what a writer that is not this
module does, and a plain append onto a block-framed log corrupts it. A store this build creates has
to stay appendable by something that writes plain JSONL. So the block log keeps its default and the
comment now carries those two test names as the reason — the hazard demonstrated rather than argued.

Vectors have no such shape: a record stays one JSON line, only a field's representation changes, and
packed and plain vectors read correctly side by side per record. Verified with no environment
variable set — 115 stored lines packed, 0 JSON arrays, and the served vector digest identical either
way, which is what makes it a representation and not a codec.

The reversibility both rely on is measured, not assumed: the same store read twice with the reader
flag flipped gives 144 records, 82 vectors and a byte-identical digest in both directions. An earlier
version of that check compared four SEPARATE stores and produced four different digests, which proved
nothing — every store derives its own identities.

One test needed the same correction the snapshot container needed in eight files:
`test_backward_compat_inline_log_reads_under_flag_on` compares `_raw_disk_records` ("the exact
durable bytes, NOT expanded") against `_read_raw_records`, which expands. Those were the same thing
while expansion had one job. Both sides go through the same unpacking now, so it stays a test about
interning. Worth noting isolation could not have diagnosed that: the test fails standalone on `main`
too, and is new here only because in the full suite it passes there and fails here.

PR #1025, opened 2026-09-06
The block log defaults on

bjmeetsfo:
The blocker is gone. This shipped off for one demonstrated reason — a plain append onto a
block-framed log made everything past it unreadable, and a store this build creates has to stay
appendable by something that writes plain JSONL. #1023 made the reader tolerate that, resuming block
parsing where the text stops, and the two tests that kept the flag off pass with it on.

## Measured

Same corpus, 60 documents of 76 KB, **10,818 records both ways**:

```
                      store        B/record   vs source
block log off    21,534,683           1,991       4.61x
block log on      2,540,847             235       0.54x    8.47x smaller
```

The store ends up **smaller than the text it was built from**.

## Three premises that were true only while the active shard was plain text

None of them is about sealing:

- **The rotation threshold.** At 40,000 bytes a block-framed log — roughly a tenth the plain size —
  never rotates, and six tests failed with `nothing rotated`: their own guards catching a vacuous run
  rather than a defect.
- **"The active shard stays plain text."** With this on it is a block stream. The invariant being
  defended is that the **sealer** never touches it — a sealed shard is a whole-file compress, and a
  file still being appended to must never carry one. Block stream and sealed are different things
  that share a magic and differ by codec byte, so the check is now the codec.
- **"A rotated shard is stored compressed."** It asserted the sealed codec on *every* rotated shard.
  With this on a rotated shard is a block stream and the sealer correctly leaves it alone. Split in
  two: the plain case, which sealing is for, and the block-stream case, which is already compressed.

## The sealer is unchanged, and that is a measurement

```
plain JSONL shard      1,721,959 -> 147,727   11.66x from sealing
block-stream shard       167,986 -> 154,469    1.09x from sealing
```

Sealing a plain shard is the whole win. A second whole-file pass over already-compressed bytes is
**nine percent**, and the block stream already lands within 14% of the sealed plain shard — so
teaching the sealer to re-compress it would cost a full pass for very little.

> An earlier draft of mine claimed that figure was far larger, by comparing a block log's ratio
> against a plain log's whole-file ratio. Two different baselines.

## Nothing on disk is converted

The form is read from the **file**, so an existing plain log stays plain and only a fresh one adopts
blocks. A store written with this on reads byte-identically with it off — the flag chooses what to
write, never what can be read.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -195,19 +195,17 @@ _SHARD_CODEC_ZLIB = b"\x01"
 #: ACTIVE log can take: a sealed shard is finished and compresses whole, but a log is appended to,
 #: and a stream of self-describing blocks is appendable where a single deflate stream is not.
 _SHARD_CODEC_BLOCKS = b"\x02"
-#: Off, and it stays off -- the hazard is demonstrated rather than argued. Turning it on breaks
-#: `test_a_writer_in_another_process_does_not_corrupt_the_view` and
-#: `test_both_append_paths_write_the_same_tail`: both append raw JSON lines to a log this module
-#: created, which is what a writer that is NOT this module does, and a plain append onto a
-#: block-framed log corrupts it. A store this build creates has to remain appendable by something
-#: that writes plain JSONL, and defaulting this on takes that away.
+#: ON. It was off for one demonstrated reason: a plain append onto a block-framed log made
+#: everything past it unreadable, and a store this build creates has to stay appendable by something
+#: that writes plain JSONL -- fixtures are built that way and out-of-process writers exist. The
+#: reader tolerates that now, resuming block parsing where the text stops, so the two tests that
+#: kept this off pass with it on.
 #:
-#: Everything else about it is safe and measured -- an existing plain log stays plain because the
-#: form is read from the FILE, and a store written with this on reads byte-identically with it off
-#: (144 records, 82 vectors, same digest). It is a per-deployment choice for anyone who knows their
-#: log has a single writer, not a default.
+#: Nothing already on disk is converted: the form is read from the FILE, so an existing plain log
+#: stays plain and only a fresh one adopts blocks. A store written with this on reads
+#: byte-identically with it off -- the flag chooses what to write, never what can be read.
 LOCAL_JSONL_BLOCK_LOG = os.environ.get(
-    "MATRIXARK_LOCAL_JSONL_BLOCK_LOG", "0"
+    "MATRIXARK_LOCAL_JSONL_BLOCK_LOG", "1"
 ).strip().lower() not in ("0", "false", "no", "off")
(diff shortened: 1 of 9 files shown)


bjmeetsfo:
The ratchet found 13 failures here, and one of them was a real defect rather than fixture churn.

**Twelve were tests reading the durable log as text.** `read_text(encoding="utf-8").splitlines()` is
fair for a plain log and impossible for a block-framed one. Nine sites across seven files now read it
through the module's own reader, which takes either form — a test about durable *records* should not
depend on the encoding they arrive in.

**The thirteenth was worth having.** `purge_tombstones` rewrote the log as plain JSON lines, so
purging a block-framed log converted it back to text and the file **grew**:

```
test_purge_compacts_tombstones_and_preserves_state
  AssertionError: 35962 not less than 12954
```

A purge exists to reclaim space, and it was the one operation that undid the compression and left the
log plain until the next rotation. It now writes the form the log is in, in 256-record blocks, with
the atomic commit — temp file, fsync, replace — unchanged.

That leaves two failures, `test_backward_compat_inline_log_reads_under_flag_on` and
`test_flag_off_is_byte_identical_to_today`, which fail **identically on a clean `main` worktree**
(head `86468294`). Pre-existing, and untouched here.

One process note, since I got it wrong earlier in this series and it is worth not repeating: the
local gate now prints the branch it is running on as part of the command, because a worktree that
several branches pass through will otherwise let a gate report the wrong branch's result with no
sign that it did.

bjmeetsfo:
awareness, no action needed.

What landed: the block log defaults on. Same corpus, 10,818 records both ways:

```
                      store        B/record   vs source
block log off    21,534,683           1,991       4.61x
block log on      2,540,847             235       0.54x    8.47x smaller
```

The store ends up smaller than the text it was built from.

**A real defect the ratchet caught.** `purge_tombstones` rewrote the log as plain JSON lines, so
purging a block-framed log converted it back to text and the file GREW — 12,954 bytes became 35,962.
A purge exists to reclaim space, and it was the one operation that undid the compression and left the
log plain until the next rotation. It writes the log's own form now, in 256-record blocks, atomic
commit unchanged.

**Twelve tests asserted the durable log is text.** Fair for a plain log, impossible for a
block-framed one. Nine sites across seven files read it through the module's own reader now — a test
about durable *records* should not depend on the encoding they arrive in.

**The sealer is deliberately unchanged**, and that is a measurement: sealing a plain shard is 11.66×,
sealing an already block-compressed one is 1.09×. A second whole-file pass for nine percent is a poor
trade, and the block stream already lands within 14% of the sealed plain shard. An earlier draft of
mine claimed that figure was far larger by comparing two different baselines.

Nothing on disk is converted: the form is read from the file, so an existing plain log stays plain
and only a fresh one adopts blocks. A store written with this on reads byte-identically with it off.

`test_backward_compat_inline_log_reads_under_flag_on` and `test_flag_off_is_byte_identical_to_today`
remain red and are red identically on a clean `main` worktree — pre-existing, untouched here.

PR #1030, opened 2026-09-06
Let the maintenance cycle dump the buckets that block log reclaim

bjmeetsfo:
A log that can never be reclaimed makes every cold start slower, forever. On one box the log had
reached **766 MB across 676 files**, growing ~15 MB/hour and never once shrinking, beside a base
index of 9.5 MB — so the durable state was small and the log was unbounded.

### Why reclaim refused

Reclaim wants, for every bucket, a dump manifest matching that bucket's **current generation**.
Buckets without one are reported as `missing_slot_generations` and raise
`slot_generation_without_durable_dump`. There were **3,085** of them.

The dump stage only ever selects buckets with `dirty_object_count > 0`. A bucket can go clean
without ever having been dumped at its current generation, and once clean nothing will dump it
again — so the blocker is permanent. Three back-to-back cycles at budget 256 left `missing=3085`
unchanged with `covered=0` every round: **not a budget problem**, and no amount of running the
existing cycle could fix it.

`selected_dump_buckets` already bypasses the dirty filter in the lifecycle planner, but
`run_storage_manager_cycle` hardcoded `selected_dump_buckets: Vec::new()`, so no caller could ever
reach it.

### The change

- `StorageManagerCycleRequest` gains `selected_dump_buckets`, threaded through to the planner.
- The proxy op takes `field == "unblock_reclaim"`, which dumps **every** bucket in one round and
  raises its own dump cap to fit, reporting `requested_unblock_buckets`.

Dumping only the *missing* buckets does not work, and the measurements say why: coverage is exactly
the last round's dump, because the cycle keeps one bucket-dump manifest. Chasing the missing list
oscillates —

```
dumped 2085 (all then-missing) -> covered=2085, missing=1008
dumped 1032 (the new missing)  -> covered=1032, missing=2061
```

— and never converges. If only the last round is covered, the last round has to hold every bucket.

### Measured

On a copy of that store, with the whole bucket set named:

```
requested_unblock=3093  covered=3093  missing=0  blockers=[]
applied=true            wal 706 MB -> 378 MB    (399 s)
```

Reclaim applied for the first time and the log halved in one round. Cold start, on two copies
differing only in whether reclaim had run, arms alternated across three rounds because load on that
box swings by 5x:

| | reclaimed (380 MB) | unreclaimed (766 MB) |
|---|---|---|
| round 1 | 29.6 s | 66.1 s |
| round 2 | 30.5 s | 41.8 s |
| round 3 | 21.0 s | 34.3 s |
| **mean** | **27.0 s** | **47.4 s** |

**1.75x faster cold start.** Peak RSS was ~410 MB in every run, so this buys latency, not memory.

An earlier single reading showed the reclaimed store *slower* (81.5 s); it was taken at load 25
against a quiet-box baseline. It is listed here because it is the reason the comparison above
alternates arms and records the load each run saw.

### Test

`a_requested_bucket_list_reaches_the_lifecycle_planner` asserts the requested buckets reach the
planner when none of them is dirty, with a control that a clean store and no list selects nothing —
without which "dump everything every cycle" would pass and reintroduce the write amplification the
dirty-only selection exists to avoid. Verified to fail with the previous `Vec::new()` restored.

`cargo test --release -p temporalstore-rust --lib storage_manager` — 11 passed.

### DIFF
--- a/crates/temporalstore-rust/src/engine/reports.rs
+++ b/crates/temporalstore-rust/src/engine/reports.rs
@@ -3026,6 +3026,13 @@ fn commit_dirty_buckets_before_truncation_default() -> bool {
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct StorageManagerCycleRequest {
     pub shard_id: ShardId,
+    /// Dump exactly these buckets instead of the dirty ones.
+    ///
+    /// Reclaim wants a dump manifest matching each bucket's current generation; a bucket that
+    /// went clean without ever being dumped at that generation has none, and the dirty-only
+    /// cadence will never dump it again. Naming buckets here is the only way to clear that.
+    #[serde(default)]
+    pub selected_dump_buckets: Vec<u32>,
     #[serde(default)]
     pub dry_run: bool,
     #[serde(default = "default_storage_manager_stage_enabled")]
@@ -3111,6 +3118,7 @@ impl Default for StorageManagerCycleRequest {
     fn default() -> Self {
         Self {
             shard_id: 0,
+            selected_dump_buckets: Vec::new(),
             dry_run: false,
             enable_prepare: true,
             enable_wal_reclaim: true,
--- a/crates/temporalstore-rust/src/engine/storage_manager_cycle.rs
+++ b/crates/temporalstore-rust/src/engine/storage_manager_cycle.rs
@@ -83,7 +83,7 @@ impl TemporalEngine {
         .collect::<Vec<_>>();
         let plan_request = StorageLifecycleRequest {
             shard_id: request.shard_id,
-            selected_dump_buckets: Vec::new(),
+            selected_dump_buckets: request.selected_dump_buckets.clone(),
             max_dump_buckets_per_round: request.max_dump_buckets_per_round,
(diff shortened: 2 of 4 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Log reclaim could be permanently blocked by buckets that went clean without ever being dumped at their current generation: the dump stage only selects DIRTY buckets, so nothing ever dumped them again and the log grew forever (766 MB and climbing on one box, ~15 MB/hour). The cycle can now be asked to dump every bucket in one round, which let reclaim apply for the first time - log 706 MB -> 378 MB, cold start 47.4 s -> 27.0 s measured with alternated arms.

bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Log reclaim could be blocked permanently by buckets that went clean without ever being dumped at their current generation: the dump stage only selects DIRTY buckets, so nothing ever dumped them again and the log grew without bound (766 MB and climbing at ~15 MB/hour on one box, beside a 9.5 MB base index). The cycle can now be asked to dump every bucket in one round, which let reclaim apply for the first time: log 706 MB -> 378 MB, and cold start 47.4 s -> 27.0 s measured on two copies with the arms alternated.

PR #1035, opened 2026-09-06
The local event log is written in blocks by default

bjmeetsfo:
The active event log is the last plain-text artifact this module writes — the snapshot, its tail
and the sealed shards are all containers already — and at the corpora this was priced against it is
around 94% of everything on disk. It has been written as a stream of blocks, one per append batch,
behind `MATRIXARK_LOCAL_JSONL_BLOCK_LOG` for a while. This makes that the default, and fixes the
one write path that did not know about it.

### Why it was off, and why that no longer holds

The hazard was demonstrated, not argued: a process that is **not** this module appends plain JSON
to the log, and a plain append onto a block-framed log corrupted it.

That is fixed rather than accepted. `_plain_run_lines` reads a plain run at a block boundary and
leaves the handle on the first byte that is not a line, so a log genuinely interleaves
`[block][plain][block]` and reads back whole and in order. Three tests pin it —
`test_a_plain_append_onto_a_block_log_still_reads`,
`test_a_block_written_after_a_plain_append_is_not_lost`, and the property the tolerance must not
swallow, `test_a_torn_block_is_still_dropped_not_read_as_text`.

Both tests that were named in the code as the reason to stay off now pass with it on.

### What it costs and buys

No durability. The batch is already the unit acked together, so a crash loses exactly what it loses
today. Measured over the batch sizes this code really appends (3×60, 5×60, 284×29, 292×1), 9,008
records, log bytes against plain JSONL:

| text | ratio |
|---|---|
| random words — the least compressible a record plausibly holds | **3.23x** |
| the 76 KB documents this was first priced against | **9.68x** |
| repetitive text | **20.23x** |

Read back through the adapter, both forms return the same records with the same digest.

### Reversible, which is what makes it a default

The form is read from the **file**, not the flag. An existing plain log stays plain and is appended
to as plain; only a fresh log adopts blocks. Turning the flag off again leaves every block log
readable, because the reader takes either form whatever the flag says. A live store therefore
converts at its next rotation rather than on upgrade, and nothing rewrites a log to get there.

### The defect this turned up: a purge read the form and wrote plain

`purge_tombstones` reads through `_iter_shard_lines`, which knows every form — and then wrote plain
JSON lines unconditionally. On the same corpus:

| | before | after | head |
|---|---|---|---|
| **before this change** | 19,258 | **316,098** — 16.4x *bigger* | `MASHRD\x01` → `{"recor` |
| **after** | 19,258 | 15,957 | `MASHRD\x01` preserved |

Not a corruption: the log came back readable, several times larger. It was worse than that, because
the form is read from the file — so every later append inherited plain from the purged file, and
one purge undid the compression permanently. A rewrite creates a new file, so it now takes the form
a *fresh* log takes, which is the rule `_log_append_form` already applies to an empty one. Written
in 256-record blocks rather than one: a block is the unit of decode, and one block per file would
make a cold read materialise every surviving record at once.

### Seven tests were reading the durable log as text

They opened the log with `read_text(encoding="utf-8")` and split lines, bypassing the iterator that
knows what form a shard is in. That was **already wrong for a sealed shard** — those have been
compressed containers by default for a while — and passed only because none of those corpora rotate
far enough to seal one. Defaulting the active log to blocks brought the same defect somewhere it
could not be missed. All now read through `_iter_shard_lines`.

Two of them were worse than a crash, and both are absence assertions:
`assertNotIn(INTERN_DICT_RECORD_TYPE, text)` over the raw file, and a substring scan for
`"matrixark_audit_log"` rows. On a compressed file neither would match anything, so an empty audit
trail would have read as a pass — an assertion passing hardest when there is nothing to read.

`test_matrixark_a_rotated_shard_is_sealed` is the one exception: it reaches rotation by writing
more plain text than the cap, and compressed records never reach it. Its subject is the sealer, not
the framing of the active log, so it now pins the plain form and says why — and each of its
assertions already began by asserting that something rotated, which is why it failed loudly instead
of passing on an empty set.

### Checks

| | |
|---|---|
| whole `tools` suite (322 files), new default vs `main` baseline | **identical** |
| whole `tools` suite, flag forced off vs `main` baseline | **identical** — the flag still turns it off |
| the block log's own suite | **14 passed** |
| mutation: make the purge write plain again | **fails**, naming the form every later append would inherit |
| round trip, both forms | same records, same digest |
| purge | form preserved, log smaller, survivors unchanged |

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -195,19 +195,29 @@ _SHARD_CODEC_ZLIB = b"\x01"
 #: ACTIVE log can take: a sealed shard is finished and compresses whole, but a log is appended to,
 #: and a stream of self-describing blocks is appendable where a single deflate stream is not.
 _SHARD_CODEC_BLOCKS = b"\x02"
-#: Off, and it stays off -- the hazard is demonstrated rather than argued. Turning it on breaks
-#: `test_a_writer_in_another_process_does_not_corrupt_the_view` and
-#: `test_both_append_paths_write_the_same_tail`: both append raw JSON lines to a log this module
-#: created, which is what a writer that is NOT this module does, and a plain append onto a
-#: block-framed log corrupts it. A store this build creates has to remain appendable by something
-#: that writes plain JSONL, and defaulting this on takes that away.
+#: ON. One block per append batch, which costs NO durability: the batch is already the unit acked
+#: together, so a crash loses exactly what it loses today. Measured over the batch sizes this code
+#: really appends (3 x60, 5 x60, 284 x29, 292 x1), 9,008 records, log bytes against plain JSONL:
 #:
-#: Everything else about it is safe and measured -- an existing plain log stays plain because the
-#: form is read from the FILE, and a store written with this on reads byte-identically with it off
-#: (144 records, 82 vectors, same digest). It is a per-deployment choice for anyone who knows their
-#: log has a single writer, not a default.
+#:     random words -- the least compressible text a record plausibly holds      3.23x
+#:     the 76 KB documents this was first priced against                         9.68x
+#:     repetitive text                                                          20.23x
+#:
+#: It shipped off for a reason that was demonstrated rather than argued, and that reason is now
+#: fixed rather than accepted. The hazard was that a process which is NOT this module appends plain
+#: JSON to the log, and a plain append onto a block-framed log corrupted it. `_plain_run_lines`
+#: reads a plain run at a block boundary and leaves the handle on the first byte that is not a
+#: line, so a log genuinely interleaves [block][plain][block] and reads back whole and in order.
+#: Pinned by `test_a_plain_append_onto_a_block_log_still_reads`,
+#: `test_a_block_written_after_a_plain_append_is_not_lost`, and -- the property that the tolerance
+#: must not swallow -- `test_a_torn_block_is_still_dropped_not_read_as_text`.
+#:
+#: Reversible, which is what makes it a default rather than a migration. The form is read from the
+#: FILE: an existing plain log stays plain and is appended to as plain, only a fresh log adopts
+#: blocks, and turning this off again leaves every block log readable, because the reader takes
+#: either form whatever the flag says.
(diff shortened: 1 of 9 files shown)


bjmeetsfo:
Superseded by #1025, which landed while this was being verified and covers the same ground independently -- the same default, the same sealed-shard premises, and the same purge defect with the same numbers (12,954 -> 35,962). #1025 keys the purge rewrite on _log_append_form() rather than on the flag, which is the better of the two: it preserves an existing block log even when the flag is off. Closing in its favour. The one piece not in #1025 is a guard for that purge fix in the block log suite -- it is covered only incidentally, by a size assertion in test_mem0_complete -- so that goes up separately.

bjmeetsfo:
Superseded by what is already on `main`, and merging it now would regress the better version.

Both sides turn block framing on by default. `main` reached it later and with a further reason this branch does not have: the reader tolerates a plain append onto a block-framed log, resuming block parsing where the text stops, so the two tests that kept the flag off pass with it on. This branch predates that and argues from the size measurement alone.

The regression is in the purge. The two branches carry the same test under different names, and they assert opposite things:

| | |
|---|---|
| `main` | `test_a_purge_rewrites_the_log_in_the_form_it_is_already_in` |
| here | `test_a_purge_rewrites_the_log_in_the_configured_form` |

`main` reads the form from the FILE. That matters because a purge is the one operation that replaces the log: rewriting it in the *configured* form converts an existing plain log to blocks, or a block log back to text — which grew one from 12,954 bytes to 35,962 — and the form is then inherited by every later append, so a single purge changes the file permanently. Reading it from the file leaves an existing store alone and lets only a fresh log adopt blocks.

That is the safer rule and it is the one on `main`. This branch would replace it.

Mechanically it is also 73 commits behind with 9 conflicting files, and it has never been through a real gate — its entire check history is `auto-approve` twice, because it predates #1092, where a base-branch filter meant the suite ratchet, the readiness contract and the Rust build never ran on it.

Closing rather than rebasing: there is nothing here `main` lacks. If the configured-form behaviour is wanted deliberately, it should be a new change against the current purge code with the size argument stated against it.

PR #1042, opened 2026-09-06
A placement head is stored compact when it is rewritten

bjmeetsfo:
A record location is stored as `"shard:offset"` — about six bytes where the `{"key", "field"}`
dict took sixty-nine. `_locator_entries_for_ref` and `_placement_entries_for_node` are the same
routine written twice: read the head, append to the tail chunk, and rewrite the head when the chunk
count rolls over.

The locator compacts its head on that rewrite. The placement one did not — it wrote `head_locations`
straight back:

```python
payload = {
    "locations": head_locations,          # the locator writes compact_location_list(head_locations, base)
    "resource_versions": merged_versions,
    "location_chunks": tail_index,
}
```

`head_locations` is read back from storage, so the long form is **self-perpetuating**: once a head
holds it, every later rollover writes all of it out again, and the head never upgrades however many
times it is rewritten.

## What it costs

Measured across the one-box log as it stands, 809 pieces and 510,082,640 bytes:

| | |
|---|---|
| long-form location dicts of the compactable shape | 18,449 |
| as stored | 1,439,022 B |
| as the compact form | 137,267 B |
| recovered | **1,301,755 B — 0.26% of the log** |
| per entry | 78.0 B -> 7.4 B (10.5x) |

0.26% of the log is the honest whole-log number. A single segment reads far worse — one
331,054-byte piece carried 364 of these, 28,392 bytes, 8.58% of that piece — because a head
rewrite dumps a whole head at once. The per-entry ratio is the part that generalises.

Every one of the 18,449 was the plainly compactable shape: key `base` + six digits, field twenty
digits. None were left alone for being a foreign base, so none of this count is a location the
compact form cannot express.

## The change

One line, so the two branches read the same way:

```python
"locations": compact_location_list(head_locations, self._location_base()),
```

`compact_location_list` leaves anything it cannot compact exactly as it is, and `expand_location`
has always accepted both shapes, so a store written before this reads unchanged and an already
compact head is untouched.

## Tests

`tools/test_matrixark_the_placement_head_keeps_the_long_form.py`, five tests:

* the rewritten head is stored compact — **fails without the change**, 64 of 64 head locations kept
  the long form
* the tail is stored compact — the positive control. It shares the fixture and the base with the
  head, so the head assertion only means something because this one passes
* the head still names the same records, through `expand_location`
* an already-compact head is left alone (idempotent)
* a foreign-base location survives the rewrite unchanged

20 tests pass across `test_placement_chunks`, `test_side_index_batch_read`, `test_side_index_merge`
and this file. This suite needs both import paths on `PYTHONPATH` (repo root and `tools/`).

### DIFF
--- a/tools/matrixark_temporal_direct_backend.py
+++ b/tools/matrixark_temporal_direct_backend.py
@@ -1505,7 +1505,7 @@ class _TemporalDirectBackendMixin:
         head_versions = head_versions if isinstance(head_versions, list) else []
         if tail_index != chunk_count or merged_versions != head_versions:
             payload = {
-                "locations": head_locations,
+                "locations": compact_location_list(head_locations, self._location_base()),
                 "resource_versions": merged_versions,
                 "location_chunks": tail_index,
             }
--- a/tools/test_matrixark_the_placement_head_keeps_the_long_form.py
+++ b/tools/test_matrixark_the_placement_head_keeps_the_long_form.py
@@ -1,0 +1,144 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2026 MatrixArkAI
+"""A placement head is rewritten in whatever form it was already stored in.
+
+`_locator_entries_for_ref` and `_placement_entries_for_node` are the same routine written twice:
+read the head, append to the tail chunk, and rewrite the head when the chunk count rolls over. The
+locator compacts its head on that rewrite. The placement one did not -- it wrote `head_locations`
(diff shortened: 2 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed (the review request on this one was auto-assigned, not asked for). A placement head rewritten on chunk rollover was written back in whatever location form it already held, so the long form was self-perpetuating; it is now compacted the way the locator path already did it - 18,449 entries across the one-box log, 78.0 B -> 7.4 B each.

PR #1043, opened 2026-09-06
Never remember a failure notice as a context pack

bjmeetsfo:
The last-good-pack fallback remembered whatever the hook had put in `additional_context`. On a
retrieval error that variable does not hold a pack — it holds a plain-language notice:

```
MatrixArk/TemporalStore retrieval was attempted for this prompt but failed. Use visible local
Codex context as authoritative for this turn. Failure: matrixark_retrieve timed out ...
```

So the notice was cached, and a later turn that could not reach the store would serve it back
labelled as prior context. That is **worse than serving nothing**: the agent is told, as remembered
history, that retrieval failed. Found on a live box as a cache file holding that sentence and
nothing else.

The `error` guard existed on the recover path but not on the remember path.

### The change

- **The call site** (`matrixark_codex_hook.py`): when `error` is set, remember nothing.
- **The cache itself** (`matrixark_hook_pack_cache.py`): `remember_context_pack` refuses a failure
  notice outright. Two hooks share this module and are edited independently, and the cost of
  getting it wrong is not a lost optimisation but a false memory — so the refusal lives with the
  store, not only with the caller.

Matched on the notice's fixed opening phrase rather than on a word like "failed", because a real
pack can discuss a failure and must still be kept. That case is a test, not an aspiration.

### Tests

`test_a_failure_notice_is_never_remembered_as_a_pack.py`:

- the notice the hook emits is recognised
- **a real pack that merely mentions failure is not** — the control against over-eager matching
- a notice cannot overwrite the last pack that actually loaded
- with only a notice to store, the cache stays empty rather than serving it later

All 21 hook-cache tests pass together.

Live cache audited afterwards: one poisoned entry removed, twelve genuine packs left untouched.

### DIFF
--- a/tools/matrixark_codex_hook.py
+++ b/tools/matrixark_codex_hook.py
@@ -1946,9 +1946,13 @@ def codex_hook_output(
             _cache_path = _pack_cache.context_pack_cache_path(
                 "codex", str(agent_context.get("workspace_root") or "")
             )
-            if additional_context.strip():
+            if error:
+                # `additional_context` holds a failure NOTICE here, not a pack. Remembering it
+                # would serve "retrieval failed" back to a later turn as remembered history.
+                pass
+            elif additional_context.strip():
                 _pack_cache.remember_context_pack(_cache_path, additional_context)
-            elif not error and not _pack_cache.store_answered(retrieve):
+            elif not _pack_cache.store_answered(retrieve):
                 _previous, _age_s = _pack_cache.recover_context_pack(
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The context-pack fallback was caching the hooks own retrieval-failure notice as if it were a pack, so a later turn could be served "retrieval failed" back as remembered history - worse than serving nothing. The remember path now refuses it, and so does the shared cache itself since both hooks use it.

PR #1048, opened 2026-09-06
A record stores the storage options a reader cannot recompute

bjmeetsfo:
`storage_options` is stored on every record, and the tail of `normalize_storage_options` merges the
route's own output back into it:

```python
route = canonical_storage_route(normalized)
normalized.update({k: v for k, v in route.items() if k in {...sixteen names...}})
```

`canonical_storage_route` is a pure function, so each record then carries a copy of a function of
its own inputs. Measured on the one-box log, `storage_options` is **7.89% of every byte written**,
and **17 of its 25 fields never take a second value** across 555 blocks — `durability_result` is
always `"accepted_for_async_durability"`, `write_ack_policy` always `"ack_after_memory_append"`,
`native_backend_decides_route` always `true`. Several are the same fact spelled more than once:
`route`, `route_key`, `backend_family`, `selected_storage_family` and `storage_family` are all
`"default"`; `oplog_mode`, `selected_write_mode`, `write_mode` and `durability` are all `"async"`.

This is the same argument `slim_persisted_storage_route` already makes for the derived half of
`storage_route`, applied to the block that one derives from.

## Only the pure outputs

Ten names are dropped — the ones `canonical_storage_route` produces and never reads back:

`route_key`, `backend_family`, `sync_write`, `async_write`, `write_ack_policy`,
`native_backend_decides_route`, `selected_storage_family`, `selected_write_mode`, `replica_read`,
`durability_result`.

Six others that the merge also writes are **kept**: `route`, `storage_family`, `write_mode`,
`durability`, `read_preference` and `background_write` are read back by `canonical_storage_route` as
INPUTS. Dropping those would silently reverse a caller who set one explicitly — an async write
re-derives `background_write` as `true`, so a request that asked for `false` would come back wrong.
Every block on the live log round-trips either way, which is exactly how that class of bug ships.
`test_an_explicit_input_survives` holds that line.

## What it recovers

Over the 40 most recent log pieces, 13,633,887 bytes:

| | |
|---|---|
| `storage_options` blocks | 1,815 |
| those blocks | 1,075,452 B (7.89% of the log) |
| with the derived names dropped | 522,357 B |
| recovered | **553,095 B — 4.06% of the log** |

This applies to records written from here on; it does not shrink what is already stored. At the
log's current 510,082,640 bytes, 4.06% is about 20.7 MB per log of this size.

Re-deriving the ten dropped names from what is kept reproduces all ten **exactly, on 1,616 of 1,616**
`storage_options` blocks taken from the live log — no disagreements.

## Where it applies

One line, at the single append chokepoint in `matrixark_temporal_direct_write.py` — the same place
`slim_persisted_storage_route` is applied, and its own comment records that this is the only append
call site.

No read-side change. Nothing reads these names off a stored record: an AST scan of every production
module finds 29 reads of the ten names, and their subjects are `normalized` (14, inside the composer
itself), `options` (8, inside `canonical_storage_route` reading its own inputs), `route` (6, on a
`storage_route` dict, which is a different dict) and `native_metrics` (1). A consumer that wants the
derived half calls `canonical_storage_route` on what is left, exactly as it already does for
`storage_route`.

## Tests

`tools/test_matrixark_the_derived_storage_options_are_not_stored.py`, eight tests. The central one
re-derives every dropped name across a matrix of option shapes rather than one fixture, because the
invariant is "recoverable", not "absent".

Both mutations were checked, with `__pycache__` cleared so a same-size edit could not report a stale
pass:

* slimmer made a no-op → 2 failures, including the material-share guard
* `background_write` added to the drop set → `test_an_explicit_input_survives` fails, alone

57 tests pass across `test_direct_record_cache`, `test_matrixark_raw_message_storage_contract`,
`test_backend_intern_metadata`, `test_side_index_merge`, `test_placement_chunks`,
`test_native_backends_reach_temporalstore`, `test_temporalstore_features` and the two new files.
These suites need both the repo root and `tools/` on `PYTHONPATH`.

### DIFF
--- a/tools/matrixark_mcp_temporal_append.py
+++ b/tools/matrixark_mcp_temporal_append.py
@@ -206,6 +206,60 @@ def slim_persisted_storage_route(record: Json) -> Json:
     return slim
 
 
+# Fields `canonical_storage_route` PRODUCES and never reads back. `storage_options` is stored on
+# every record, and the tail of `normalize_storage_options` merges the route's output into it:
+#
+#     route = canonical_storage_route(normalized)
+#     normalized.update({k: v for k, v in route.items() if k in {...sixteen names...}})
+#
+# so each record carries a copy of a pure function of its own inputs. Measured on the one-box log,
+# `storage_options` is 6.53% of every byte written and 17 of its 25 fields never vary at all.
+#
+# Only the PURE outputs are dropped. `route`, `storage_family`, `write_mode`, `durability`,
+# `read_preference` and `background_write` are read back by `canonical_storage_route` as INPUTS, so
+# dropping them would silently override a caller who set one explicitly -- e.g.
+# `background_write=False` on an async write re-derives as True. Today's records round-trip either
+# way, which is exactly how that class of bug ships; they stay.
+#
+# Verified over 1,616 storage_options blocks taken from the live log: re-deriving these ten from the
+# kept fields reproduces all ten EXACTLY, 1,616 of 1,616, and the block is 51.1% smaller.
+_OPTIONS_KEYS_DERIVED_FROM_THE_REST = (
+    "route_key",
+    "backend_family",
+    "sync_write",
+    "async_write",
+    "write_ack_policy",
+    "native_backend_decides_route",
+    "selected_storage_family",
+    "selected_write_mode",
+    "replica_read",
+    "durability_result",
+)
+
+
+def slim_persisted_storage_options(record: Json) -> Json:
+    """Persist the storage options a reader cannot recompute, not the ones it can.
+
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. normalize_storage_options merged canonical_storage_route output back into the options, so every record stored a copy of a pure function of its own inputs (7.89%% of the live log, 17 of 25 fields never varying). The ten pure outputs are dropped and re-derive exactly on 1,616 of 1,616 live blocks; the six that are read back as INPUTS stay, so an explicit background_write=false is never re-derived over. Recovers 4.06%% of the log going forward.

PR #1053, opened 2026-09-06
The envelope keeps its own copy of the derived storage options

bjmeetsfo:
A record carries `storage_options` in two places — its own top level and inside its `envelope` —
and `slim_persisted_storage_options` (mx#1048) only reached the first. The envelope's copy went on
spelling out the same ten derived fields.

This is the shape that keeps recurring in this code: one routine reaches a nested copy and its
sibling does not. It is the same defect as the placement head in mx#1042, found the same way — by
looking at what the live log actually contains after the change, rather than at whether the change
was applied.

## What it recovers

Over 30 recent log pieces, 10,703,272 bytes:

| | |
|---|---|
| envelope `storage_options` blocks | 40 (36 at `record_bundle.envelope`, 4 at `envelope`) |
| those blocks | 18,980 B |
| trimmed | 8,235 B |
| recovered | **10,745 B — 0.100% of the log** |

Small, and stated as measured rather than rounded up. It is worth doing because the inconsistency
is the bug: a reader of `slim_persisted_storage_options` reasonably expects it to have trimmed the
record, and until now it had trimmed part of it.

Re-deriving the dropped names from what is kept reproduces them **exactly on all 35** envelope
blocks that carried the derived half — no mismatches.

## The change

The trimming moves into `_slim_options_block(holder)`, applied to the record and then to its
envelope. Same drop set as mx#1048, so the same argument holds: only the ten names
`canonical_storage_route` produces and never reads back, never the six it consults as inputs.

The caller's record and envelope are not mutated — `test_the_input_record_is_not_mutated` holds
that, since the envelope is a nested dict and copying the outer record alone would have shared it.

## Tests

Seven added to `tools/test_matrixark_the_derived_storage_options_are_not_stored.py`, 15 in the file
now. Mutation-checked with `__pycache__` cleared: removing the envelope reach fails
`test_the_envelopes_own_block_is_trimmed` and `test_both_copies_are_trimmed_together`, and **no
record-level test**, so the new tests are aimed at the envelope specifically rather than restating
what mx#1048 already covers.

### DIFF
--- a/tools/matrixark_mcp_temporal_append.py
+++ b/tools/matrixark_mcp_temporal_append.py
@@ -237,27 +237,44 @@ _OPTIONS_KEYS_DERIVED_FROM_THE_REST = (
 )
 
 
-def slim_persisted_storage_options(record: Json) -> Json:
-    """Persist the storage options a reader cannot recompute, not the ones it can.
-
-    A consumer that wants the derived half calls `canonical_storage_route` on what is left, exactly
-    as it already does for the derived half of `storage_route`.
-    """
-    if not isinstance(record, dict):
-        return record
-    options = record.get("storage_options")
+def _slim_options_block(holder: Json) -> Json | None:
+    """`holder` with its `storage_options` trimmed, or None when there is nothing to trim."""
+    if not isinstance(holder, dict):
+        return None
+    options = holder.get("storage_options")
     if not isinstance(options, dict) or not options:
-        return record
+        return None
     kept = {key: value for key, value in options.items()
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Follow-up to #1048: the record carries storage_options in two places and the slimmer only reached the top-level one, so the envelope copy still spelled out the same ten derived fields. Same drop set, now applied to both. 10,745 B over 30 recent log pieces; all 35 envelope blocks re-derive exactly.

PR #1058, opened 2026-09-06
A record derives its own storage kind

bjmeetsfo:
`storage_record_kind` maps a record type to a kind — `context_index` to `"index"`,
`context_summary` to `"summary"` — and `storage_part_for_record` just calls it. Both values are then
stored on the record, so each is usually the same string arrived at twice. On the live log the pair
is **1.00% of every byte written**, across 11,417 records.

## Dropped only when the record derives it

The first line of `storage_record_kind` is:

```python
explicit = str(record.get("storage_record_kind") or record.get("storage_part") or "")...
if explicit:
    return explicit
```

so the field is an **override**, not only a cache: a caller can store a kind the mapping would not
produce. Dropping it unconditionally would silently discard that.

So a field is dropped only when the record derives **that exact value** without it. A kind the
mapping would not produce fails the comparison and is kept, which makes this lossless *by
construction* rather than by what the corpus happens to hold today. It is the same conditional shape
`slim_persisted_storage_route` uses for a derivable placement.

## The measurement, and how it was nearly wrong

Deriving the kind **without stripping the stored fields first** compares the value to itself, because
of the `explicit` line above. Done that way it reports a 100% match whatever the truth is — the
first measurement of this "confirmed" derivability on 11,398 records and proved nothing.

Stripping both fields before deriving: **11,393 records, 0 disagreements.**

Running the real function over 150 log pieces, 92,616,751 bytes:

| | |
|---|---|
| records carrying either field | 11,417 |
| kept because the kind was not derivable | **0** |
| those records | 22,864,066 B -> 21,936,013 B |
| recovered | **928,053 B — 1.00% of the log** |

## Tests

`tools/test_matrixark_a_record_kind_that_derives_itself.py`, ten tests.
`test_the_mapping_is_not_the_identity` is the positive control: without it every other test would
still pass if the mapping merely echoed `record_type`, and the suite would assert nothing about the
mapping.

Mutation-checked with `__pycache__` cleared:

* drop unconditionally (remove the override guard) → breaks `test_an_unmappable_kind_is_kept`,
  `test_a_part_that_differs_from_the_kind_is_kept` and `test_a_part_alone_is_handled` — the three
  override-preservation tests, and nothing else
* slimmer made a no-op → breaks the four that assert the drop

Applied at the same single append chokepoint as mx#1048 and mx#1053.

### DIFF
--- a/tools/matrixark_mcp_temporal_append.py
+++ b/tools/matrixark_mcp_temporal_append.py
@@ -13,8 +13,10 @@ from typing import Any
 
 try:
     from tools.matrixark_mcp_core import Json, compact_latest_context_state_records, stable_hash
+    from tools.matrixark_mcp_storage_options import storage_record_kind
 except ModuleNotFoundError:  # Direct script execution from tools/.
     from matrixark_mcp_core import Json, compact_latest_context_state_records, stable_hash
+    from matrixark_mcp_storage_options import storage_record_kind
 
 
 # Keys of `storage_route` that are NOT recoverable from `storage_options`: where this record was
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. storage_record_kind/storage_part are the mapping own output stored back on the record, 1.00%% of the live log across 11,417 records. Each is dropped only when the record derives that exact value without it, so an unmappable kind is kept. Worth knowing: measuring derivability WITHOUT stripping the fields first compares the value to itself and matches 100%% of the time regardless - the first measurement did that and proved nothing.

PR #1061, opened 2026-09-06
The kind inside the options block is the same string again

bjmeetsfo:
`normalize_storage_options` writes `storage_record_kind` and `storage_part` **into the options
block** as well as onto the record:

```python
normalized["storage_record_kind"] = record_kind
normalized["storage_part"] = record_kind
```

So a record held the same string up to four times. Measured on the live log, of the 11,334 records
carrying the kind at all, **11,121 held all four copies**. mx#1058 removed the two at the top level;
this removes the two inside the block.

| | |
|---|---|
| the pair inside the options block | 912,339 B — **0.98% of the log** |
| agrees with what the record derives | 11,197 |
| differs | **0** |

Running the slimmer over 150 pieces (92,974,249 B), both trims together recover **1,837,014 B —
1.98% of the log**, of which mx#1058 accounted for 1.00%.

## Why it is safe to drop there

Nothing reads them inside the block. Every read of these two names in the tree is on a
record-shaped subject — `record.get("storage_record_kind")` in `matrixark_mcp_indexing`,
`matrixark_mcp_serving_records` and `storage_record_kind()` itself. Inside the options block they
are only ever written, by the four `normalized[...] = record_kind` sites above.

The same conditional as mx#1058 still applies: a value is dropped only when the record derives that
exact string without it, so a kind the mapping would not produce is kept — in the block as well as
at the top level.

## A guard this change needed

The first version returned early when the top-level pair was absent, which skipped the options block
entirely on records that carried the kind only there. Three of the new tests failed on it, and the
guard now asks whether **any** of the four copies is present before returning early. That is the
whole reason `test_the_inner_pair_goes_even_when_the_outer_is_absent` and
`test_an_options_block_left_empty_is_removed` are in the suite.

## Tests

Eight added, 18 in the file. Beyond the drop and the override cases:

* an options block left empty by the trim is removed rather than stored as `{}`
* the caller's options dict is not trimmed in place — it is a nested dict, so copying the outer
  record alone would have shared it
* a non-dict `storage_options` is left alone
* after all four copies go, `storage_record_kind()` still returns the original value

77 tests pass across the write-path suites and both new files.

### DIFF
--- a/tools/matrixark_mcp_temporal_append.py
+++ b/tools/matrixark_mcp_temporal_append.py
@@ -301,7 +301,10 @@ def slim_persisted_record_kind(record: Json) -> Json:
         return record
     kind = record.get("storage_record_kind")
     part = record.get("storage_part")
-    if not isinstance(kind, str) and not isinstance(part, str):
+    options = record.get("storage_options")
+    inner = options if isinstance(options, dict) else {}
+    if not any(isinstance(value, str) for value in
+               (kind, part, inner.get("storage_record_kind"), inner.get("storage_part"))):
         return record
     probe = {key: value for key, value in record.items()
              if key not in ("storage_record_kind", "storage_part")}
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Follow-up to #1058: normalize_storage_options writes the same kind pair INTO the options block too, so a record held the string up to four times (11,121 of 11,334 held all four). Both trims together are 1.98%% of the live log. The first attempt returned early when the top-level pair was absent and skipped the block entirely - three of the new tests caught it.

PR #1063, opened 2026-09-06
Three audit modes, three vocabularies, one of them offered

bjmeetsfo:
There are **three** audit modes in this build, with **three different vocabularies**, and the portal
offered one of them.

| variable | takes | layer | offered before |
|---|---|---|---|
| `MATRIXARK_AUDIT_MODE` | `off` `async` `full` `sync` | access | yes |
| `MATRIXARK_CONTEXT_AUDIT_MODE` | `off` `telemetry_only` `full` | retrieval | **no** |
| `MATRIXARK_DIRECT_AUDIT_MODE` | `buffered` `sync` `drop` | store writes | **no** |

`sync` and `full` each appear in two of the three, meaning different things. So a customer reasoning
by analogy from the only one they could see — *"audit mode is async"* — gets

```
MatrixArkError: audit_mode must be full, telemetry_only, or off
```

from a retrieve. All three are offered now, each with its own list and each help naming a word the
others use that **this** one refuses.

### The retrieval one had a consequence beyond being unreachable

With nothing set, whether a retrieve records anything **depends on which path served it**:

```
the request path / direct read   ->  ('telemetry_only', 0.01)
the local adapter                ->  ('off', 0.01)
```

That difference is deliberate and documented in the adapter. What was missing is the way to resolve
it: setting `MATRIXARK_CONTEXT_AUDIT_MODE` makes both paths agree, on any of the three values, and
until now a deployment could not set it from the portal. The declared default is
`telemetry_only` — what two of the three paths do — and the help names the exception rather than
pretending there is one answer.

### Derived, not restated

The choices are read out of the code that validates them:

- `retrieval.context_audit_mode`'s list is parsed from `if audit_mode not in {…}: raise` in
  `matrixark_mcp_retrieve_planning`.
- `limits.direct_audit_mode`'s list is collected from every `self._audit_mode == "…"` branch in
  `matrixark_mcp_temporal_audit`.

Two vocabularies drifting apart silently is how this got confusing; a list retyped here would drift
the same way. Each derivation has a floor asserting it actually found the set, since an empty one
would compare nothing to nothing.

The "warn about the other words" rule is derived too. It does not grep for a keyword — it computes
which values belong to the *other* two vocabularies and not to this one, and requires the help to
name at least one of them. With a floor: if the three ever stopped overlapping, that rule would be
satisfiable by saying nothing.

```
a choice the code accepts is dropped from the list     -> FAIL
a word from another vocabulary is offered here         -> FAIL
the store's choices stop matching what it branches on  -> FAIL
the help stops naming a word it refuses                -> FAIL
the store default drifts from the build                -> FAIL
the sample rate drifts from the build                  -> FAIL
the declared retrieval default is the local adapter's  -> FAIL
the validator stops refusing a foreign word            -> FAIL
```

### Two labels the audit corrected

I filed all three as `restart`. `test_matrixark_gateway_config_audit` failed twice, and it was right
both times: the two retrieval controls are read *inside* the call that uses them, so a change lands
on the next retrieve. They are `live`. The store one is bound when its module is imported and stays
`restart`. A test now states that asymmetry, because it is a fact about where the read is rather
than a style choice.

17 tests, plus 70 across the suites that police the registry. Stacked on
[#1050](https://github.com/matrixarkai/TemporalStore/pull/1050).

### DIFF
--- a/tools/matrixark_gateway_config.py
+++ b/tools/matrixark_gateway_config.py
@@ -554,6 +554,35 @@ SETTINGS: List[Setting] = [
             "MATRIXARK_TIME_COMPRESSION_MAX_RAW_EVENTS_PER_NODE",
             "Raw events kept per node", "int", "256", "restart",
             "The ceiling on raw events one node keeps before compression is forced."),
+    # ---- three audit modes, three vocabularies ---------------------------------------------------
+    # `audit.mode` above governs the ACCESS layer. These two govern different layers and take
+    # different words, which is the sharp edge: `sync` and `full` appear in more than one of them
+    # meaning different things, and a value borrowed from the wrong one is rejected outright.
+    Setting("retrieval.context_audit_mode", "retrieval",
+            "MATRIXARK_CONTEXT_AUDIT_MODE",
+            "Retrieval audit mode", "str", "telemetry_only", "live",
+            "Whether a retrieve records what it did. off keeps nothing; telemetry_only keeps the "
+            "shape of the answer; full keeps the evidence, which is what makes a pack "
+            "reconstructable afterwards and what makes the record log grow. "
+            "This is NOT the access-layer audit above and does not take its words: async and "
+            "sync are refused here. "
+            "With nothing set the answer depends on which path served the retrieve -- the request "
+            "path and the direct read use telemetry_only, the local adapter uses off -- so "
+            "setting this is how a deployment makes them agree.",
+            ("off", "telemetry_only", "full")),
+    Setting("retrieval.context_audit_sample_rate", "retrieval",
+            "MATRIXARK_CONTEXT_AUDIT_SAMPLE_RATE",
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
Restacked on the rebased #1050 and repointed at the live module.

`tools/matrixark_mcp_temporal_audit.py`, which this test read to derive the store audit vocabulary, was deleted on main by #1080 as an orphan. That was right: #1080's own message records that the mixin's methods live on elsewhere, and the three-way branching on `self._audit_mode` is intact in `matrixark_temporal_direct_write.py` — `drop`, `sync`, `buffered`, the same set, still derived from live code rather than written down here.

The helper now names the file when it cannot find it, instead of raising `FileNotFoundError` from inside itself. A bare traceback said nothing about which claim had broken; the message now says which attribute to go looking for and warns that the choices below would otherwise be derived from nothing.

105 tests green.

PR #1065, opened 2026-09-06
A hook envelope is not a turn worth remembering

bjmeetsfo:
`payload_text` falls back to `json.dumps(payload, sort_keys=True)[:4000]` when it finds no message
anywhere in a hook payload, guarded by `identity_only_payload`. That guard exists precisely to stop
an event envelope becoming a remembered turn — but its key set did not know `session_title`,
`source`, `transcript_path`, `model` or `prompt_id`.

So a Claude Code `SessionStart` payload —

```json
{"cwd": "...", "hook_event_name": "SessionStart", "session_id": "...",
 "session_title": "...", "source": "resume", "transcript_path": "..."}
```

— failed the check on its last three keys, and its whole envelope was stored as a 313-character
**user message**. Retrieval then served those back as memory.

## What the model was actually being given

From the packs the hook cached on this machine — the exact strings injected as
`additionalContext`:

| pack | chars | items | envelope items | envelope share |
|---|---|---|---|---|
| `claude-6237e6d7be596636` | 7,422 | 18 | **18** | **99.3%** |
| `claude-887c38f0211585f5` | 8,000 | 6 | 0 | 0.0% |

One of the two carried **nothing but envelopes** — about 1,855 tokens of `cwd`, `session_id` and
`transcript_path` in place of remembered turns, inside an 8,000-character budget that then had no
room for anything else. Nine of its items opened with the same 40 characters.

Two packs is a small sample and is stated as what it is; the mechanism, though, is not
sample-dependent — every text-less hook event took this path.

## The change

Five keys added to `IDENTITY_ONLY_PAYLOAD_KEYS`. Nothing else.

**Nothing is lost by not storing it.** The same payload is already kept verbatim on the record as
`metadata.raw_hook_payload`. This stops the envelope being a *retrievable turn*, not from being
recorded.

`source` and `session_title` are the two whose names could plausibly hold prose in some other hook.
They are listed because a key only suppresses a payload when **every** other key is also an envelope
key — a payload carrying any content field still passes through — and the set already takes that
posture with `params`, `metadata`, `turn` and `run`.

## Tests

`tools/test_matrixark_the_hook_envelope_is_not_a_user_turn.py`, seven tests. The ones that matter
most are the three proving content still gets through — a guard that suppressed real turns would be
far worse than the waste it fixes:

* a payload with a `prompt` is not suppressed, and its text is the prompt
* a payload with `messages` is not suppressed
* content nested under a known key (`metadata.prompt`) still counts, because the guard recurses

`test_it_is_the_added_keys_that_do_it` is the positive control: with any one of the three removed
from the set, the payload must stop reading as envelope-only.

Mutation-checked with `__pycache__` cleared: taking the five keys back out fails the four
suppression tests and **none** of the content-passthrough ones.

## About the hook suites

These twelve suites are **red on main**: 57 failures, 9 errors, 7 skipped. With this change: 57
failures, 9 errors, 7 skipped — and the failing **set** is byte-identical, 66 names, with nothing
appearing or disappearing on either side. Counts alone would not have shown that, so they were
diffed by name.

### DIFF
--- a/tools/matrixark_codex_hook.py
+++ b/tools/matrixark_codex_hook.py
@@ -3341,6 +3341,24 @@ def latest_codex_assistant_message_from_rollout(payload: Json) -> str:
     return ""
 
 
+# A payload made only of these is the hook telling us WHICH turn happened, not WHAT was said. When
+# every key is one of them, `payload_text` returns "" and no memory is written -- otherwise the
+# whole envelope is stored as the turn's own text.
+#
+# The five added below were missing, so a Claude Code `SessionStart` payload -- `cwd`,
+# `hook_event_name`, `session_id`, `session_title`, `source`, `transcript_path` -- failed the check
+# on its last three keys and was stored verbatim as a 313-character "user" message. Retrieval then
+# surfaced those as memory: measured on a live injected pack, 16 of 18 items were `SessionStart`
+# envelopes and they filled 92% of an 8,000-character context budget with `cwd`, `session_id` and
+# `transcript_path`, crowding out the turns a reader actually wanted.
+#
+# Nothing is lost by not storing it: the same payload is already kept verbatim on the record as
+# `metadata.raw_hook_payload`. This stops it being a RETRIEVABLE TURN, not from being recorded.
+#
+# `source` and `session_title` are the two whose names could plausibly hold prose elsewhere. They
+# are listed because a key only suppresses a payload when EVERY other key is also an envelope key,
+# so a payload carrying any content field still passes through -- and the set already takes this
+# posture with `params`, `metadata`, `turn` and `run`.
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. identity_only_payload did not know session_title/source/transcript_path/model/prompt_id, so a SessionStart envelope was stored as a 313-char user message and retrieval served it back as memory: one live injected pack was 18 of 18 envelope items, 99.3%% of an 8,000-char budget, ~1,855 tokens of cwd/session_id/transcript_path instead of remembered turns. The payload is still kept as metadata.raw_hook_payload. The twelve hook suites are red on main (57F/9E) and the failing SET is byte-identical with this change - diffed by name, not by count.

PR #1068, opened 2026-09-06
A context pack is cut between items, not mid-word

bjmeetsfo:
`additional_context_from_retrieve` joined every retrieved item and sliced the result:

```python
body = "\n".join(f"- {line}" for line in lines)
return (header + "\n" + body)[:max_chars]
```

The cut lands wherever the 8,000-character budget runs out. On a pack this hook actually injected,
that was mid-word — the model's last remembered turn arrived as:

```
- ...current main ever has. So upgrading still costs cold-load time, but tha
```

A fragment that stops mid-clause is worse than not sending the item. It does not read as truncated;
it reads as a complete thought whose meaning changes at the cut, and the tokens are spent either
way.

## The change

Add items while they fit, stop when the next one would not. The budget is untouched — this only
moves where the cut falls, and the pack still fills it: on the test that packs 40 items into 1,000
characters the output is >900.

One case needed deciding rather than defaulting: a single item longer than the whole budget. Sending
nothing would return `""`, which the caller reads as "nothing was retrieved" and which then triggers
the last-good-pack fallback — a worse outcome than a head of the item. So that one case still slices,
deliberately, and `test_one_oversized_item_is_still_sent` pins it.

## Tests

`tools/test_matrixark_a_pack_is_cut_between_items.py`, seven tests: no item is cut in half, the
budget is never exceeded at four different caps, the budget is still filled, an oversized single item
is still sent, everything fits when it fits, duplicates are still dropped, and nothing retrieved is
still `""` (the fail-open contract).

Mutation-checked with `__pycache__` cleared: restoring the old slice fails
`test_no_item_is_cut_in_half` and **nothing else** — in particular the budget and fail-open tests
still pass, which is what says those are not what changed.

## About the hook suites

These ten suites are red on main: 66 failing names. With this change: the same 66, diffed by name
with nothing appearing or disappearing on either side.

### DIFF
--- a/tools/matrixark_agent_hook.py
+++ b/tools/matrixark_agent_hook.py
@@ -1128,8 +1128,21 @@ def additional_context_from_retrieve(retrieve: Json | None, *, max_chars: int = 
     if not lines:
         return ""
     header = "Relevant context from earlier turns (MatrixArk memory):"
-    body = "\n".join(f"- {line}" for line in lines)
-    return (header + "\n" + body)[:max_chars]
+    # Cut BETWEEN items. Slicing the joined string to max_chars ends wherever the budget runs out,
+    # which on a live pack was mid-word -- the model's last remembered turn arrived as
+    # "...upgrading still costs cold-load time, but tha". A fragment that stops mid-clause is worse
+    # than not sending the item: it reads as a complete thought that changes meaning at the cut.
+    out = header
+    for line in lines:
+        entry = "\n- " + line
+        if len(out) + len(entry) > max_chars:
+            break
+        out += entry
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The injected pack was sliced to max_chars, so the cut landed wherever the budget ran out; on a live pack that was mid-word. Items are now added while they fit. Budget unchanged and still filled. One deliberate exception: a single item longer than the whole budget still slices, because returning empty reads to the caller as nothing retrieved and triggers the last-good-pack fallback.

PR #1074, opened 2026-09-06
A message with no text needs no record of how its text was selected

bjmeetsfo:
Every message the hook builds gets a `codex_memory_selection` block — policy, policy_counts,
selected and original char and line counts, ratios, stage. On a message with **no content** that
block reads:

```json
{"selected_text_chars": 0, "original_text_chars": 0, "dropped_text_chars": 0,
 "retained_text_ratio": 1.0, "selection_lossy": false, ...}
```

**572 bytes describing a selection that never happened.**

A hook event carrying no text produces exactly that, and mx#1065 made those the common case: it
stopped the event envelope being used as the message text, so a `SessionStart` now yields an
empty-content message — which then carried more metadata than content.

## Only the both-empty case

The block earns its bytes precisely when a selection **had** text and kept little or none of it —
that is the thing someone reading this telemetry later wants to find. So the condition is on the
original text as well as the selected text: a message keeps its block if either is non-empty, and
loses it only when there was nothing to select from.

`test_a_selection_that_dropped_everything_is_still_recorded` is what stops the lossy case being
swept up with the empty one.

## Tests

`tools/test_matrixark_a_selection_of_nothing_needs_no_record.py`, five tests:

* an empty message carries no selection block, and is left as exactly `{role, content}`
* a message with text keeps its block
* a selection that had text and produced none still records one
* `test_the_block_was_worth_removing` is the positive control — it asserts the block really is
  >300 bytes, so if it ever shrank to a field or two this change would be flagged as no longer
  worth having rather than silently kept

Mutation-checked with `__pycache__` cleared: attaching unconditionally again fails the two
empty-message tests and neither of the keep-it tests.

## About the hook suites

Ten of these suites are red on main: 66 failing names. With this change: the same 66, diffed by
name, with nothing appearing or disappearing on either side.

### DIFF
--- a/tools/matrixark_agent_hook.py
+++ b/tools/matrixark_agent_hook.py
@@ -598,13 +598,21 @@ def hook_messages_from_payload(payload: Json, *, event: str, text: str) -> list[
         normalized_role = normalized_hook_message_role(role, event=event)
         selected_content = compact_for_role(normalized_role, content)
         message: Json = {"role": normalized_role, "content": selected_content}
+        original = content if original_content is None else original_content
         selection = codex_memory_selection_metadata(
             role=normalized_role,
             event=event,
             text=selected_content,
-            original_text=content if original_content is None else original_content,
+            original_text=original,
         )
-        if selection:
+        # 572 bytes of it, on a message with no content: selected_text_chars 0, original 0,
+        # dropped 0, retained ratio 1.0 -- a description of a selection that never happened. A hook
+        # event carrying no text produces exactly that, and mx#1065 made those the common case by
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Follow-on to #1065: every hook message carried a codex_memory_selection block, and on a message with no content that is 572 bytes saying it selected 0 of 0 characters. #1065 made those the common case. The block is kept whenever the selected OR original text is non-empty, so a selection that had text and kept none of it is still recorded - that case is the one the telemetry exists for.

PR #1076, opened 2026-09-06
An entity ref need not say its type twice

bjmeetsfo:
An entity ref is rendered into a context pack as:

```python
text = f"{record.get('entity_type', '')}: {record.get('entity_name', '')} = {record.get('state', '')}"
```

written out identically at **five call sites**. The name usually already carries the type, so the
line says it twice. From a pack this hook actually injected:

```
 - [1] ref
   tool_evidence: tool_evidence = tool: Exit code: 0; Ran 159 tests
 - [2] ref
   codex_validation: codex_validation:ran_159_tests = tool validation: Ran 159 tests
```

## What it costs

Over 448 entity refs in the live log:

| | |
|---|---|
| `entity_name` == `entity_type` | 195 |
| `entity_name` starts with `entity_type + ":"` | 186 |
| unrelated (both kept) | 67 |
| rendered today | 83,334 chars |
| without the repeat | 76,026 chars |
| | **8.8% shorter, ~1,827 tokens** |

**381 of 448.** Every one of those characters is a token the model pays for on the turn the pack is
injected into, and the pack is bounded — so a character spent restating the type is one not spent on
a remembered turn.

Nothing durable changes: these five sites render a ref for the pack out of fields that are already
stored separately (`entity_type`, `entity_name`, `state`). The stored record is untouched.

## What stays

Only an exact match or a `type + ":"` prefix counts as a repeat. `session` / `session_memory` keeps
both, because a name that merely resembles the type is a different name and the pairing is
information. `test_an_underscore_is_not_a_prefix_match` holds that line.

## The five copies become one

The expression was duplicated verbatim four times in `matrixark_local_adapter_retrieve.py` and once
in `matrixark_local_adapter_ingest.py`. It is now `entity_ref_text()` in `matrixark_mcp_core`, which
both modules already star-import.

## Tests

`tools/test_matrixark_an_entity_ref_need_not_say_its_type_twice.py`, seven tests, including:

* `test_nothing_is_lost` — whatever the shape, the name and the state both survive
* `test_missing_fields_do_not_raise` — pins the degenerate rendering as **unchanged**, empty parts
  and all, so the helper cannot have quietly altered it while fixing the repeat
* `test_every_call_site_uses_the_helper` — counts `ast.Call` nodes, not grep hits, and also fails if
  an f-string spelling the old rendering reappears

Mutation-checked with `__pycache__` cleared: making the helper always render both fails the two
not-repeated tests and nothing else.

34 retrieve, ingest, entity, pack and pipeline suites pass — 0 failures, name-diffed against main.

### DIFF
--- a/tools/matrixark_local_adapter_ingest.py
+++ b/tools/matrixark_local_adapter_ingest.py
@@ -4068,7 +4068,7 @@ class _LocalAdapterIngestMixin:
                 ref_type = "summary"
                 ref_hash = record.get("summary_hash") or record.get("node_hash")
             elif record_type == "context_entity":
-                text = f"{record.get('entity_type', '')}: {record.get('entity_name', '')} = {record.get('state', '')}"
+                text = entity_ref_text(record)
                 ref_type = "entity"
                 ref_hash = record.get("entity_hash")
             elif record_type == "context_segment":
--- a/tools/matrixark_local_adapter_retrieve.py
+++ b/tools/matrixark_local_adapter_retrieve.py
@@ -2211,7 +2211,7 @@ class _LocalAdapterRetrieveMixin:
                 secondary_index_matched_count += 1
             if not admit_candidate_for_node(record):
                 continue
-            text = f"{record.get('entity_type', '')}: {record.get('entity_name', '')} = {record.get('state', '')}"
+            text = entity_ref_text(record)
             entity_metadata = embedding_metadata_by_ref.get(("entity", record.get("entity_hash")), {})
             source_entity_hashes = record.get("source_entity_hashes", [])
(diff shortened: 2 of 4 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Five call sites rendered an entity ref as "type: name = state" and the name usually already carries the type: 381 of 448 refs in the live log. Dropping the repeat is 8.8%% off their rendered length, ~1,827 tokens of a bounded pack budget. An unrelated name still shows both. Nothing durable changes - these sites render for the pack out of fields stored separately. The five copies are now one helper.

PR #1122, opened 2026-09-06
Key the context pack cache on the repository, not the checkout

bjmeetsfo:
The last-good-pack cache was keyed on `workspace_root`, the checkout directory. That fragments it
the moment an agent works across git worktrees — and Codex does. A live box held **18 cache files
for 2 agents**, spread across `wt-l1cache`, `wt-model`, `wt-flags`, `wt-snapbin`, `wt-scratch`,
`wt-hermetic`, `wt-livebudget`, `wt-placement-head`, `wt-budgetpath` and more.

Every new worktree therefore starts with **no pack to fall back to** — which is exactly the turn the
fallback exists for. One Claude turn was observed returning empty after 434 s with nothing available
to serve.

### The checkout directory was never the thing worth distinguishing

What must not be mixed is one **project's** context with another's. Git already answers that:
sibling worktrees share a common git dir, a different project has its own.

```
/root/src/github-services/TemporalStore  ->  TemporalStore/.git
/root/wt-l1cache                         ->  TemporalStore/.git      same pack now
/root/wt-flags                           ->  temporalstore-oss/.git  separate
/root/src/github-services/MatrixCache    ->  MatrixCache/.git        separate
```

So the cache keys on repository identity. The agent split is unchanged: Claude and Codex still
never share a file.

### Failure is the old behaviour, not a lost cache

A workspace that is not a repository, a box without git, a git call that errors or exceeds its 5 s
timeout — every one returns the path unchanged and keys exactly as before. The cache must never
cost a turn its answer.

Existing entries become unreachable under the new key and are rebuilt on the next successful turn.
That is a one-off miss, weighed against a fallback that currently misses whenever an agent changes
worktree.

### Tests

`test_sibling_worktrees_share_one_context_pack.py` builds a real repository and a real worktree:

- a worktree and its repository share an identity, and share the cache file
- **two repositories do not** — the control, without which collapsing everything to a single key
  would pass the tests above while letting one project's context surface in another
- a directory that is not a repository still keys, rather than losing its cache
- a missing path falls back instead of raising
- an absent workspace is still addressable
- **agents still do not share a file** — repository identity must not undo the agent split

Skipped cleanly where git is unavailable. All 28 hook-cache tests pass together.

### DIFF
--- a/tools/matrixark_hook_pack_cache.py
+++ b/tools/matrixark_hook_pack_cache.py
@@ -17,10 +17,45 @@ from __future__ import annotations
 
 import hashlib
 import os
+import subprocess
 import time
 from pathlib import Path
 
 DEFAULT_MAX_AGE_S = 86400.0
+
+
+def workspace_identity(workspace_root: str) -> str:
+    """The repository a workspace belongs to, so sibling worktrees share one pack.
+
+    Keying on the checkout directory fragments the cache the moment an agent works across git
+    worktrees -- every new worktree starts with nothing to fall back to, which is precisely the
+    turn the fallback exists for. The distinction worth keeping is between PROJECTS, and a git
+    common dir draws exactly that line: sibling worktrees share one, a different project has its
+    own.
+
+    Any failure returns the path unchanged, so a workspace that is not a repository, or a box
+    without git, behaves as before rather than losing its cache.
+    """
+    if not workspace_root:
+        return "-"
+    try:
+        finished = subprocess.run(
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The context-pack fallback was keyed on the checkout directory, so it fragmented across git worktrees (18 cache files for 2 agents on one box) and every new worktree started with nothing to fall back to. It now keys on the git common dir, so sibling worktrees share a pack while different projects still never do.

PR #1126, opened 2026-09-06
Every site that persists a record slims it first

bjmeetsfo:
mx#1048, #1058 and #1061 applied their slimmers at the bundle path in
`matrixark_temporal_direct_write`, on the strength of a comment there reading *"This is the ONLY
append call site"*. That comment is true of the append **client** — it is about which client the
batch is routed through — and not of record serialisation.

There are **four** sites that turn a record into a persisted entry:

| site | slimmed before this |
|---|---|
| `matrixark_temporal_direct_write.py` bundle path | yes |
| `matrixark_temporal_direct_write.py` legacy index path | **no** |
| `matrixark_temporal_direct_backend.py` raw path | **no** |
| `matrixark_mcp_raw_ingestion.py` | **no** |

## What that cost

Measured on the 139 log pieces written since those changes deployed — so this is a live bypass, not
history:

| | |
|---|---|
| records written | 26,943 |
| **still carrying the derived fields** | **870 (3.2%)** |
| of which `matrixark_idempotency` | 504 |
| of which `context_summary` | 365 |
| those records | 2,340,965 B -> 1,533,153 B |
| recovered | **807,812 B — 1.31% of the log** |
| per affected record | 2,690 -> 1,762 B (**35% smaller**) |

## The change

One function, `slim_persisted_record`, composes the three existing slimmers, and all four sites call
it. Each slimmer is already a no-op on a record without the fields it looks for, so applying it on
the raw paths costs nothing where those fields never appear — and means a change to what gets
slimmed reaches every persist site instead of three of four.

## The test is the point

`test_no_persist_site_serialises_an_unslimmed_record` walks the AST of all three modules and
requires every `payload = json.dumps(...)` to sit in a function that calls `slim_persisted_record`.

It is function-scoped rather than line-literal on purpose: the bundle path serialises a variable
built from an already-slimmed list, which a literal check reads as a bypass when it is not one. The
function-scoped form still catches the case that matters — a **fifth** site added to a function that
never slims.

Mutation-checked with `__pycache__` cleared: putting `matrixark_mcp_raw_ingestion` back the way it
was fails the guard with the exact file, line and function —

```
matrixark_mcp_raw_ingestion.py:218 in append_raw_ingestion_records() serialises a record
for storage and that function never calls slim_persisted_record
```

Applying a slimmer to one of several equivalent paths is the defect that produced four of the last
ten changes here. A structural check is the only kind that catches the next one before it ships.

77 tests pass across the write-path, side-index, placement and record-kind suites.

### DIFF
--- a/tools/matrixark_mcp_raw_ingestion.py
+++ b/tools/matrixark_mcp_raw_ingestion.py
@@ -13,9 +13,11 @@ from typing import Any
 try:
     from tools.matrixark_mcp_errors import MatrixArkError
     from tools.matrixark_mcp_identity import stable_hash
+    from tools.matrixark_mcp_temporal_append import slim_persisted_record
 except ModuleNotFoundError:  # Direct script execution from tools/.
     from matrixark_mcp_errors import MatrixArkError
     from matrixark_mcp_identity import stable_hash
+    from matrixark_mcp_temporal_append import slim_persisted_record
 
 
 Json = dict[str, Any]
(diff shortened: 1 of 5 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Correcting my own omission: #1048/#1058/#1061 slimmed ONE of the four sites that serialise a record for storage, because I trusted a comment saying "This is the ONLY append call site" - true of the append client, not of serialisation. Measured on 139 pieces written since: 870 of 26,943 records still went out fat (504 idempotency, 365 summary), 35%% larger than needed - 1.31%% of the log. One helper now composes the slimmers and all four sites use it, plus an AST guard that names the file, line and function if a fifth site appears.

PR #1129, opened 2026-09-06
The latest-state copy of a record is slimmed too

bjmeetsfo:
mx#1126 brought four persist sites under one slimmer and added a guard. After deploying it,
`context_summary` was **still 6 of 12 records written fat**. There was a fifth writer, and the guard
could not see it.

## Finding it

Reading call graphs sent me down three wrong paths — the proxy client (transport, it forwards a
value serialised upstream), the direct write queue (it flushes back through the slimmed path), and
"these are old records being rewritten" (their timestamps are after the deploy, so they are new).

The data answered it. The frame header carries `p[]` (pages: an object id and a `"~N"` blob
pointer) and `t[]` (index entries: object key `o`, object id `i`), so a blob maps back to the key
it was written under:

```
13 fat context_summary records -> matrixark:claude-hook:context_latest_state
```

That names `_latest_context_state_entries` and `_split_compacted_latest_context_state`, which write
a copy of the record under a separate key for current-value lookups.

## Why the guard missed it

Both spell the entry with no intermediate variable:

```python
payload_record = dict(record)
payload_record.pop("summary_version_hash", None)
entries.append({"key": ..., "field": field,
                "value": json.dumps(payload_record, sort_keys=True, separators=(",", ":"))})
```

mx#1126's guard looked for `payload = json.dumps(...)`. **A guard that knows one spelling of the
idiom is not a guard.** It now also inspects an entry dict's inline `"value"`.

It is narrowed at the same time: the same entry shape carries index postings — `{"ref_hashes": ...}`,
`{"locations": ...}` — which have none of the fields a slimmer looks for. Flagging those would be
noise that trains the next reader to wave the test through, so the guard requires the serialised
expression to name a record.

## What it recovers

Measured on the pieces written since mx#1126 deployed:

| | |
|---|---|
| records still carrying the derived fields | 17, all `context_summary` |
| those records | 66,120 B -> 48,950 B |
| recovered | **17,170 B — 0.47% of the log** |
| per affected record | 3,889 -> 2,879 B (**26% smaller**) |

## Tests

Mutation-checked with `__pycache__` cleared: reverting both latest-state sites fails the guard with
the exact file, line and function —

```
matrixark_temporal_direct_backend.py:1621 in _latest_context_state_entries() serialises a
record for storage and that function never calls slim_persisted_record
```

— which is precisely what the previous version of the guard could not report.

77 tests pass across the write-path, side-index, placement and record-kind suites.

### DIFF
--- a/tools/matrixark_temporal_direct_backend.py
+++ b/tools/matrixark_temporal_direct_backend.py
@@ -1615,7 +1615,7 @@ class _TemporalDirectBackendMixin:
             field = self._latest_context_state_field(record)
             if not field:
                 continue
-            payload_record = dict(record)
+            payload_record = dict(slim_persisted_record(record))
             payload_record.pop("summary_version_hash", None)
             entries.append(
                 {
@@ -1640,7 +1640,7 @@ class _TemporalDirectBackendMixin:
             if not field:
                 append_records_for_log.append(record)
                 continue
-            payload_record = dict(record)
+            payload_record = dict(slim_persisted_record(record))
             payload_record.pop("summary_version_hash", None)
             latest_state_entries.append(
                 {
--- a/tools/test_matrixark_every_persist_site_slims_the_record.py
+++ b/tools/test_matrixark_every_persist_site_slims_the_record.py
@@ -49,14 +49,36 @@ def _payload_assignments(tree):
             and n.func.id == "slim_persisted_record"
             for n in ast.walk(fn))
         for node in ast.walk(fn):
-            if not isinstance(node, ast.Assign):
+            if isinstance(node, ast.Assign) \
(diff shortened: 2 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. A fifth writer put a copy of the record under context_latest_state unslimmed, so context_summary stayed 6 of 12 fat after #1126. Found by mapping the blob back through the frame page and index arrays to its storage key, after reading call graphs sent me down three wrong paths. The #1126 guard missed it because both sites spell the entry value inline with no variable named payload - it now inspects an entry dict value too, narrowed to record-valued entries so index postings do not create noise.

PR #1138, opened 2026-09-06
A pack does not need to state a number the model can count

bjmeetsfo:
Every codex pack opened with:

```
Retrieval summary: context_pack_id=<id>, selected_refs=3, used_context_tokens=35,
local_context_refs_seen=0.
```

`selected_refs` restates the length of the list printed directly below it. The model can count. It
costs context on every turn inside a bounded pack, and it is the one field here nothing else pins.

## What it is worth

Measured across the packs cached on one box: 18 packs, 12,943 characters.

| | |
|---|---|
| `selected_refs=N, ` | 272 chars — **2.10% of these packs** |
| per pack | 15 chars |

Small, and stated as measured rather than rounded up. It is on every codex turn.

For context on why this is the only field removed: the retrieval summary line as a whole is **17.1%
of every byte a codex pack carries**, and only 41.3% of these packs is retrieved memory at all — the
rest is the preamble, the summary, the query echoed back, and the ref headers.

## What is deliberately NOT removed

`used_context_tokens` was in the first version of this change. A name-diff of the hook suites
against main caught that removing it breaks
`test_retrieve_budget_summary_reads_nested_context_pack_wrapper`, which asserts the **rendered**
`used_context_tokens=42` as its proof that a nested pack wrapper was parsed correctly.

Weakening someone else's assertion to save a few characters is a bad trade, so the field stays and
`test_the_internal_accounting_stays` now records why — so the next person to look at this line finds
the reason rather than repeating the attempt.

`context_pack_id` stays because it is the only handle correlating a pack with the logs.
`local_context_refs_seen` stays because it tells the model whether local context was considered at
all, which changes how it should weigh what follows.

## Tests

`tools/test_matrixark_the_pack_summary_drops_what_the_model_cannot_use.py`, six tests. The one that
carries the argument is `test_the_count_is_still_derivable_from_the_pack`: dropping the number is
only safe because the refs are still listed, so it asserts they are. If they ever stopped being
listed, removing the count would have destroyed information rather than restated it.

Mutation-checked with `__pycache__` cleared: putting the field back fails
`test_the_countable_field_is_gone` and neither of the keep-them tests.

The ten hook suites are red on main at 65 failing names; with this change, the same 65, diffed by
name.

### DIFF
--- a/tools/matrixark_codex_hook.py
+++ b/tools/matrixark_codex_hook.py
@@ -1783,10 +1783,24 @@ def additional_context_from_retrieve(
             "Merge this remote memory with the visible local Codex context. "
             "Prefer current local files when they conflict with retrieved memory."
         ),
+        # `selected_refs` restated the length of the list printed directly below it, and
+        # `used_context_tokens` is this hook's own accounting -- neither is something the model can
+        # act on, and both cost it context on every turn. Measured across the packs cached on one
+        # box, this line was 17.1% of every byte a codex pack carried.
+        #
+        # `context_pack_id` stays: it is the only handle correlating a pack with the logs, and an
+        # operator reading the pack cache has nothing else to key on. `local_context_refs_seen`
+        # stays because it tells the model whether local context was considered at all, which
+        # changes how it should weigh what follows.
+        # `selected_refs` restated the length of the list printed directly below it -- the model
+        # can count -- and it cost context on every turn inside a bounded pack.
+        #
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. selected_refs restated the length of the list printed directly below it, on every codex turn inside a bounded pack. Worth knowing what did NOT change: used_context_tokens was in the first version, and a name-diff of the hook suites caught that removing it breaks a test asserting the rendered value as proof a nested pack wrapper parsed. It stays, and the test now records why so the next person does not repeat the attempt.

PR #1144, opened 2026-09-06
Count the places that touch a field before changing it

bjmeetsfo:
Before changing what a record stores, one question decides whether the change is tractable: **how
many places touch this field?** Answering it by eye is unreliable, and this repository has the scar
tissue to prove it.

Three changes in a row were shipped on the belief that one routine wrote a field:

* mx#1048/#1058/#1061 slimmed the bundle path, on the strength of a comment reading *"This is the
  ONLY append call site"* — true of the append **client**, not of record serialisation
* mx#1126 brought **four** sites under one slimmer, and records were still written fat
* mx#1129 found the latest-state writers — still fat
* mx#1131 found the **second complete copy of the append implementation**, in the module where the
  slimmers are defined

Four rounds, because nothing counted the sites.

## What it does

```
$ python3 tools/matrixark_field_sites.py storage_record_kind source_roles
storage_record_kind
  reads     5   across 3 module(s)
  writes    8   across 3 module(s)
source_roles
  reads   100   across 23 module(s)
  writes  110   across 19 module(s)
```

That contrast is the whole point. `storage_record_kind` was a change worth making and it worked.
`source_roles` has a real 0.97%-of-the-log saving available and **100 read sites across 23 modules**,
where a single missed site serves an empty list where names should be — silently, on the retrieve
path. The number is the decision, and it takes one command instead of a day.

`--detail` lists every site with the expression it is read from, because reading a name off a
request is a different thing from reading it off a stored record.

## Why not grep

A grep counts imports, definitions, comments, docstrings and string literals alongside real
accesses, and it cannot tell a READ from a WRITE. That distinction is the one that matters: a
write-side change needs every writer, a read-side change needs every reader, and the two counts are
usually nothing alike — `raw_hook_payload` is **0 reads and 3 writes**.

## Tests

`tools/test_matrixark_field_sites.py`, eight tests against fixtures in a temporary directory, so the
expected numbers do not drift when someone adds a call.

`test_prose_and_string_literals_are_not_accesses` is the positive control: the fixture mentions the
field in a docstring, a comment and a bare string, and none of the three may be counted. A grep
reports all three, and the count is then wrong in the direction that makes a change look harder than
it is.

The tests also caught a real defect while this was being written. `ast.walk` yields every node, so
`record["field"] = x` was counted as a WRITE from the assignment and again as a READ when the walk
reached the same subscript on its own. Targets are now collected first and skipped by identity —
`test_a_subscript_assignment_is_a_write_not_a_read` holds that.

### DIFF
--- a/tools/matrixark_field_sites.py
+++ b/tools/matrixark_field_sites.py
@@ -1,0 +1,139 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2026 MatrixArkAI
+"""Count the places that read and write a record field, by AST.
+
+Before changing what a record stores, the question that decides whether the change is tractable is
+"how many places touch this field?" -- and answering it by eye is unreliable. Three changes in a row
+were shipped here on the belief that one routine wrote a field, each time because a comment or a
+call graph said so; the field was written in four places, then five, then six.
+
+A grep does not answer it either: it counts imports, definitions, comments and docstrings alongside
+real accesses, and it cannot tell a READ from a WRITE. That distinction is the one that matters,
+because a write-side change needs every writer and a read-side change needs every reader, and the
+two counts are usually nothing alike.
+
+    $ python3 tools/matrixark_field_sites.py storage_record_kind
+    storage_record_kind
+      reads   9   across 5 module(s)
+      writes 13   across 6 module(s)
+
+    $ python3 tools/matrixark_field_sites.py --detail source_roles
+    ... every site, with the expression it is read from
+
+What counts as what:
+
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Four rounds of persist-site changes (#1048/#1058/#1061, #1126, #1129, #1131) were each shipped on a wrong count of how many places write a field, because nothing counted them. This does, reads and writes separately, by ast node - so prose and string literals are not mistaken for accesses. The contrast is the point: storage_record_kind is 5 reads across 3 modules and was worth doing; source_roles has a real 0.97%-of-log saving and 100 reads across 23 modules, so it is not.

PR #1145, opened 2026-09-06
Keep the dump manifests reclaim still needs, not just the newest

bjmeetsfo:
Reclaim can only advance its durable frontier for a bucket whose **current generation** has a dump
manifest. Pruning kept exactly one manifest — the newest — so the moment an ordinary cycle dumped a
handful of dirty buckets, the manifest covering everything else was deleted and coverage collapsed.

On a continuously written store that makes the log **impossible to reclaim**, no matter how often
maintenance runs.

### Measured on a live store

A round that dumped all 3,190 buckets let reclaim apply for the first time and the log fell
**806 MB → 478 MB**. Within the hour, an ordinary cycle had replaced that manifest with a small one:

```
current_wal_sequence: 75346     durable frontier: 0
covered_slot_count: 0           uncovered_slot_count: 3190
blocker_reasons: ["slot_generation_without_durable_dump"]
```

Frontier zero, nothing reclaimable, the log growing again — and cold start grows with it at roughly
0.13 s per MB (709 MB → 66.6 s against 417 MB → 28.0 s, measured on copies of the same store).

So the log's growth was never really about the dump cadence. Each maintenance round did its work
and the next ordinary cycle threw it away.

### The change

`retained_bucket_dump_manifest_ids` keeps the newest manifest **plus any older one that is still
the only dump covering some bucket**.

Self-limiting: manifests are walked newest first, so once one covers everything, every older
manifest adds no coverage and is pruned exactly as before. The newest is always kept whether or not
it adds coverage, so the existing guarantee — there is always a most recent dump to anchor on — is
unchanged.

### Tests

`pruning_keeps_the_manifests_reclaim_still_needs`, built around the live shape (a wide dump followed
by a narrow one):

- the newest manifest is kept
- the wide dump is kept, because it is the only cover for the buckets the narrow one omits
- **an older manifest whose buckets are all covered by newer ones is still pruned** — the control,
  without which "retain everything" would pass
- **the newest is kept even when it covers no buckets** — the previous guarantee

Verified to fail with newest-only retention restored. 15 related manifest and cycle tests pass
together.

### DIFF
--- a/crates/temporalstore-rust/src/engine/bucket_dump_io.rs
+++ b/crates/temporalstore-rust/src/engine/bucket_dump_io.rs
@@ -200,12 +200,36 @@ pub(super) fn bucket_dump_manifest_chain_issues(
 /// Anything older is now prunable unless a follower cursor or raft snapshot ref pins it, which
 /// is exactly what those cursors are for.
 pub(super) fn retained_bucket_dump_manifest_ids(manifests: &[BucketDumpManifest]) -> BTreeSet<String> {
-    manifests
-        .iter()
-        .max_by_key(|manifest| (manifest.index_log_sequence, manifest.created_unix_ms))
-        .map(|manifest| manifest.manifest_id.clone())
-        .into_iter()
-        .collect()
+    // The newest manifest, PLUS any older one that is still the only dump covering some bucket.
+    //
+    // Reclaim advances its frontier only for buckets that have a manifest matching their current
+    // generation. Keeping just the newest meant an ordinary cycle -- which dumps only the few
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Dump-manifest pruning kept only the NEWEST manifest, so an ordinary cycle dumping a few dirty buckets deleted the manifest covering everything else and reclaim coverage collapsed to zero. On a live store that undid a full reclaim within the hour (806 -> 478 MB, then blocked again with durable frontier 0), which is why the log could never stay reclaimed. Pruning now keeps a covering set; it is self-limiting, and the newest manifest is still always kept.

PR #1146, opened 2026-09-06
A bare ref header does not earn a line of its own

bjmeetsfo:
A retrieved ref was rendered as two lines:

```
 - [1] ref
   tool_evidence: tool_evidence = tool: Exit code: 0; Ran 159 tests
```

When the header carries nothing but an index and a type — no citation, no score, no token count —
the newline and the following indent are spent on nothing, on every ref, inside a bounded pack.

Measured on the packs cached on one box: **48 of 49 headers were bare**, and folding them saves 336
characters, **2.7% of those packs**.

Small, and stated as measured. It is on every codex turn, and it is a rendering change only: the
same index, the same type, the same text.

## What still gets its own line

A header carrying a citation, a score or a token count keeps its line. Running those together with
the text would bury them, and being visible at a glance is the only reason the header exists.
`test_a_header_with_a_score_keeps_its_own_line` is the control for that, and it asserts both halves
— that the score is on the header line and the text is not.

## Tests

`tools/test_matrixark_a_bare_ref_header_does_not_earn_a_line.py`, six tests: the fold happens, the
text is not lost, a rich header is untouched, a ref with no text still renders, the numbering stays
in order, and the pack actually gets shorter.

Mutation-checked with `__pycache__` cleared: disabling the fold fails
`test_a_bare_header_shares_its_line_with_the_text` and `test_the_pack_gets_shorter`, and **not** the
control.

The hook suites are red on main at 77 failing names; with this change, the same 77, diffed by name
rather than by count.

### DIFF
--- a/tools/matrixark_codex_hook.py
+++ b/tools/matrixark_codex_hook.py
@@ -1834,8 +1834,16 @@ def additional_context_from_retrieve(
                 header_bits.append(f"score={score}")
             if token_cost != "":
                 header_bits.append(f"tokens={token_cost}")
-            lines.append(" - " + " | ".join(header_bits))
             text = _ref_text(ref)
+            # A header carrying only "[n] type" says nothing the content line does not, so it does
+            # not earn a line of its own. 48 of 49 headers on one box were bare. Anything richer --
+            # a citation, a score, a token count -- still gets its own line, because then the header
+            # is information and running it together with the text would bury it.
+            if text and len(header_bits) == 1:
+                lines.append(" - " + header_bits[0] + "  "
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. A ref header carrying only an index and a type was costing a newline and an indent on every ref inside a bounded pack; 48 of 49 headers on one box were bare. Folding them saves 2.7%% of those packs. A header carrying a citation, score or token count keeps its own line - running those together with the text would bury them, and that is what the control test asserts.

PR #1156, opened 2026-09-06
Two guards caught what three merges put in, and nobody moved them

bjmeetsfo:
Both failures are on main, on every open pull request, and neither has anything
to do with the branch it fails on. They are the two guards doing their job while
what they guard moved underneath them.

A test module imported another test module at import time. That reorders
`unittest discover` and can fail tests it has nothing to do with -- in CI, while
passing locally -- which is exactly what the cross-import guard forbids, and it
named the file. Its three uses are all inside functions, so the import moves to
the point of use.

Not by adding the file to the guard's list of known offenders: that silences a
guard with its own exception list rather than fixing what it found.

The summary path added two sets of provider names -- what it maps onto the model
route, and what it treats as that route directly. They follow the rule the guard
enforces, which is that such a set must be a named constant so the mirror test
can read the writer's own literals instead of restating them. Their comment says
so. Only the registry of sanctioned names was not extended, so they read as
inline classification.

Registering them does not weaken the guard. The paired test walks that same list
and fails if an entry is missing or empty, so a name added there has to keep
existing to keep passing.

Both verified the honest way: they fail on clean main and pass with this.

### DIFF
--- a/tools/test_matrixark_a_docstring_is_not_an_import.py
+++ b/tools/test_matrixark_a_docstring_is_not_an_import.py
@@ -36,7 +36,16 @@ import unittest
 TOOLS = os.path.dirname(os.path.abspath(__file__))
 sys.path.insert(0, TOOLS)
 
-import test_a_module_only_tests_reach_is_not_live as guard  # noqa: E402
+
+def _guard():
+    """The module this one checks, imported on use.
+
+    Importing it at import time reorders `unittest discover`, which is the thing
+    the cross-import guard exists to prevent -- so this module does not do it.
+    """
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
Already on main, arrived independently while this was open.

Both failures are fixed there, and by the same means: the docstring test imports
the module it checks inside a `_guard()` helper rather than at import time, and
the two summary sets are registered in the allowed list. Both tests pass on a
clean main checkout now.

Closing rather than rebasing. This branch was cut before that landed, so what it
would add over main is mostly the removal of work that arrived after it -- the
opposite of useful.

PR #1163, opened 2026-09-06
Opening a store decoded every page in it to find one number

bjmeetsfo:
Opening a block store asks one question of each slab: what is the highest page id in it, so the
next append can start above it. It answered by decoding the slab.

`next_page_id_at` read every page segment whole and handed each to `inspect_slab`, which builds a
complete block index report — per record a zstd decompression, a checksum verify, a **second**
SHA-256 hex-encoded into a `String`, and a report struct pushed to a `Vec` — and then kept one
field of it. Everything else was allocated and dropped.

Measured on a store of 1,017 MB across 119 segments holding **3,434,580 page records**, that is
per open:

| | before | after |
|---|---|---|
| zstd decompressions | 3,434,580 | 0 |
| SHA-256 computations | 6,869,160 | 0 |
| report `Vec` built, then dropped | ~655 MB | 0 |
| segment bytes held in memory at once | up to 264 MB | 0 |

The page id and the record length both live in the header, so the walk never needs a payload.
`max_page_id_in_slab_file` steps the file by `header_len + stored_len`, reading headers only.

It reads through a `BufReader` rather than seeking per record. That detail is load-bearing: these
records average ~300 bytes, so a seek and a read for each is millions of syscalls, and measured
**slower** than reading the whole segment (9,688 ms against 811 ms on the same files).
`seek_relative` drops buffered bytes in place when the destination is already buffered, so a run
of small records costs one read per bufferful while a large record still seeks.

## This also stops the counter going backwards

Under the old code a corrupt **payload** ended the walk, because reaching the page id meant
decoding the record. Every page id behind the damaged record then went uncounted and
`next_page_id` came back **low** — which is page-id reuse, and the stale reads that follow it.
A header walk never consults the payload, so those ids still stand.

Halting on a torn **header** is unchanged and deliberate. `LocalBlockStore::with_options` fences a
torn tail on the active slab precisely because this scan stops early; returning an error instead
would reach the caller's `unwrap_or_default()` and reset the counter to 0.

## Tests

`next_page_id_at` had no test at all. It now has four, and one of them tells the two
implementations apart rather than passing either way:

- `a_corrupt_payload_does_not_lower_the_next_page_id` — a record whose header is intact and whose
  payload is not. Decoding the payloads answers 5; reading the headers answers 10.
- `a_torn_header_halts_the_walk` — the fencing behaviour the active-slab truncation depends on.
- `the_next_page_id_clears_every_slab` — the highest id across slabs, not merely the last written.
- `an_empty_root_starts_at_zero`.

### DIFF
--- a/crates/temporalstore-rust/src/block_store.rs
+++ b/crates/temporalstore-rust/src/block_store.rs
@@ -30,7 +30,8 @@ use paths::{
 use record::{
     decode_page_record, default_page_record_compression_enabled,
     default_page_record_compression_level, default_page_record_compression_min_bytes,
-    encode_page_record, inspect_slab, logical_range_from_slab, sha256_hex, summarize_slab,
+    encode_page_record, inspect_slab, logical_range_from_slab, max_page_id_in_slab_file,
+    sha256_hex, summarize_slab,
     PageRecordCompression,
 };
 use self::band_manifest::*;
--- a/crates/temporalstore-rust/src/block_store/record.rs
+++ b/crates/temporalstore-rust/src/block_store/record.rs
@@ -2,7 +2,8 @@
 // Copyright 2026 MatrixArkAI
 
 use std::collections::BTreeSet;
-use std::io::Cursor;
+use std::fs::File;
+use std::io::{BufReader, Cursor, Read, Seek};
 
 use sha2::{Digest, Sha256};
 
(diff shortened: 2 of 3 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Opening a block store ran a whole-store scan to find the next page id: it decoded every page record (3,434,580 of them on the live store) with a zstd decompression, a checksum verify and a second SHA-256 each, building a ~655 MB report it then discarded but for one field. The page id is in the header, so it now walks headers through a BufReader and never touches a payload. It also stops a corrupt payload from lowering next_page_id, which was page-id reuse.

PR #1166, opened 2026-09-06
Stop re-hashing every page record for a field nobody reads

bjmeetsfo:
`inspect_slab` runs on every engine open. For each page record it verifies the stored checksum
inside `decode_page_record` — and then hashes the payload a **second** time and hex-encodes it into a
64-character `String`, to fill `BlockStoreBlockIndexReport::checksum`.

Nothing reads that field. `block_index_entries` appears only in its own struct definition and in
tests, no test asserts the checksum, and the field is `skip_serializing_if = "Option::is_none"`, so
omitting it is wire-compatible.

It is now behind `TS_BLOCK_INDEX_CHECKSUMS=1`, default off. Kept rather than deleted because it is a
genuine diagnostic for hand-inspecting a slab — it just should not be computed for every record of
every open when no caller wants it.

### On the performance claim: I could not measure one

The motivation was cold start. Slab verification on a live-store copy runs at about **13.5 MB/s**
where sha256 on this hardware does hundreds, and this is one of two full passes per record.

But this box cannot resolve the effect. Three alternated pairs on the same copy:

| | runs | mean | spread |
|---|---|---|---|
| base | 116.2 / 44.6 / 38.5 s | 66.5 s | **3.0x** |
| with this change | 62.2 / 99.3 / 32.2 s | 64.6 s | **3.1x** |

Both arms swing 3x from page-cache state alone and the means differ by 3%, which is noise. An
earlier attempt at the same problem — parallelising the slab fan-out — produced an apparent 1.5x
that dissolved under the same test, so this is stated plainly rather than dressed up.

**So this is a cleanup on the merits of the code, not a measured speedup:** provably redundant work
removed from a hot path, with the diagnostic still available on request. If cold start is attacked
again, the target is the per-record verification rate itself.

### Tests

`cargo test --release -p temporalstore-rust --lib block_store` — 64 passed.

### DIFF
--- a/crates/temporalstore-rust/src/block_store/record.rs
+++ b/crates/temporalstore-rust/src/block_store/record.rs
@@ -823,6 +823,22 @@ pub(super) fn max_page_id_in_slab_file(
     Ok(max_page_id)
 }
 
+/// TS_BLOCK_INDEX_CHECKSUMS: recompute and record a hex digest per page record while inspecting.
+///
+/// Default OFF. `inspect_slab` runs at every engine open, and this hashes each payload a second
+/// time -- `decode_page_record` has already verified the stored checksum -- then allocates a
+/// 64-character String for it. Measured on a live-store copy, slab verification ran at 13.5 MB/s
+/// against hundreds of MB/s for sha256 alone. Nothing in the crate reads the field; it is kept for
+/// hand-inspecting a slab.
+fn block_index_checksums_enabled() -> bool {
+    std::env::var("TS_BLOCK_INDEX_CHECKSUMS")
+        .map(|value| {
+            let value = value.trim().to_ascii_lowercase();
+            value == "1" || value == "true" || value == "yes"
+        })
+        .unwrap_or(false)
+}
+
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. inspect_slab hashed every page record a second time, after decode_page_record had already verified it, to fill a report field nothing reads; now opt-in behind TS_BLOCK_INDEX_CHECKSUMS=1. Stated in the PR as a code-merits cleanup rather than a measured speedup - this box swings 3x on page-cache state, so the cold-start effect was not resolvable.

bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. inspect_slab hashed every page record a second time (after decode_page_record had already verified it) to fill a report field nothing reads; now opt-in behind TS_BLOCK_INDEX_CHECKSUMS=1. Complementary to #1163, which removed the page-id scan caller - the band-manifest reconcile still calls inspect_slab at every open. Stated in the PR as a code-merits cleanup, not a measured speedup: this box swings 3x on page-cache state so the effect was not resolvable.

PR #1172, opened 2026-09-06
A test method was pasted twice, so 33 lines never ran

bjmeetsfo:
`test_there_is_one_packing_sort_key.py` defined `test_only_one_module_defines_the_predicate_the_penalty_depends_on` at line 155 and again at line 188. The two copies are byte for byte identical, so the later one shadowed the earlier and 33 lines sat there reading as live code while never running.

Removed the shadowed copy. The file still runs five tests, because the shadowed one was never one of them.

This is exactly what `test_no_module_defines_the_same_top_level_name_twice` exists to catch, and it is the failure currently red on `main` — so the ratchet gate is red on every branch until this lands.


bjmeetsfo:
Superseded: this fix rides in the combined gate-unblock PR, because the flag-inventory defect on main kept this branch's ratchet red on its own.

PR #1173, opened 2026-09-06
One definition of the packing-predicate test

bjmeetsfo:
`test_there_is_one_packing_sort_key.py` defines
`test_only_one_module_defines_the_predicate_the_penalty_depends_on` twice, at lines 155 and 188.
Python keeps the second and discards the first, so one of them never ran -- and
`test_no_module_defines_the_same_top_level_name_twice` is failing on main because of it, which means
every pull request inherits the failure and the ratchet will not merge.

The two definitions are byte-identical. The duplicate is removed and the surviving one is untouched,
so nothing that was running stops running and nothing that was asserted stops being asserted.

The guard that caught this arrived in #1159, whose subject is "One name meant two things, and the
guard for it was never committed". It found a duplicate in the file it was added to, which is the
guard working rather than the guard being wrong.

### DIFF
--- a/tools/test_there_is_one_packing_sort_key.py
+++ b/tools/test_there_is_one_packing_sort_key.py
@@ -185,39 +185,6 @@ class ThereIsOnePackingSortKeyTest(unittest.TestCase):
             "agrees wherever the caller has already established ref_type == 'event', which is "
             "most call sites, so nothing routine catches the difference")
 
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Superseded: main already carries this. The duplicate definition of the packing-predicate test is gone from main, and this branch now rebases to zero files against it, so there is nothing left to merge.

PR #1176, opened 2026-09-07
Nine functions written out twice, now written once

bjmeetsfo:
`tools/` carries 206 groups of byte-identical function bodies — about 3,139 lines of the same code
written out more than once. The shape is an extraction that was started and never finished:
modules were split out of `matrixark_mcp_core.py`, and the copy left behind stayed live on its own
import path. So both copies are reachable, and which one a caller gets is decided by its import
line rather than by anything deliberate.

This collapses nine of them. The body moves nowhere: the module that keeps it is unchanged, and the
other module imports it.

## Why each copy chosen is the safe one

Four checks, each excluding a way this could go wrong quietly:

- **Signatures compared separately from bodies.** A body hash ignores parameters, so two functions
  whose defaults differ hash identically; merging those would change the answer at every call site
  relying on a default. Groups whose interfaces differ are left alone.
- **Module-level definitions only.** Collected from the module body rather than by walking the
  whole tree, so nothing inside a class is touched — a method replaced by a module-level import
  would leave its class without the attribute. Anything taking `self`/`cls`, or carrying a
  decorator, is refused too.
- **Direction derived from the module graph.** The importing module is always one that ALREADY
  reaches the module keeping the body, so the graph gains no edge and no cycle can appear. Pairs
  that reach each other both ways are left for a later change that gives the body a third home.
- **Counting a use includes calls inside the defining module.** An earlier version of the
  analysis called a private helper dead because nothing imports it, when it is called ten lines
  below itself.

## How it is verified

By **object identity**, not by reading the code: after the change, `owner.f is importer.f` for
every consolidated function. Comparing sources, or calling both and comparing answers, would pass
just as well with two copies still present.

Each module's `__all__` still resolves — the import binds the name — which matters because
`__all__` and string literals are what keep several of these copies reachable at all.

## One pair deliberately left alone

`build_shared_context_policy` stays duplicated. Three tests encode that particular pair on purpose:
two assert the copies are two DISTINCT objects, because the test above them loops over both to
check each honours an environment ceiling — collapse them and that loop silently covers one module
twice while continuing to pass. The third lists the module as one that reads a setting through
`live_int`, which stops being true once the body moves.

No coverage is actually lost by consolidating it — the module keeping the body still holds the read
and is still scanned — but changing what three tests assert deserves its own review rather than a
line inside a bulk cleanup.

## Result

Nine functions, 482 lines. The whole `tools/` suite (381 files) ran before and after, the before run
against a full worktree at the same commit — a tools-only tree fails dozens of files on missing
`crates/`, `docs/` and `config/` directories, and comparing against that would let a real regression
hide among phantom failures.

341 pass / 40 fail before, 342 pass / 39 fail after, with **zero tests failing only after**. All 39
remaining failures fail identically on both trees. The single difference is one test that failed
before and passes now, which reproduces as flaky when run on its own.

### DIFF
--- a/tools/matrixark_mcp_budget_pack.py
+++ b/tools/matrixark_mcp_budget_pack.py
@@ -325,75 +325,10 @@ def candidate_memory_layer_name(candidate: Json) -> str:
     return context_class or ref_type or "unknown"
 
 
-def local_context_budget(args: Json) -> Json:
-    raw_items = args.get("local_context", [])
-    if raw_items is None:
-        raw_items = []
-    if not isinstance(raw_items, list):
-        raise MatrixArkError("local_context must be an array")
-    items: list[Json] = []
-    text_hashes: set[int] = set()
-    token_total = 0
-    for index, item in enumerate(raw_items):
-        if isinstance(item, str):
-            text = item
-            source = f"local:{index}"
-            ref_type = "local_context"
-        elif isinstance(item, dict):
-            text = str(item.get("text") or item.get("content") or "")
-            source = str(item.get("source") or item.get("ref") or f"local:{index}")
-            ref_type = str(item.get("ref_type") or "local_context")
-        else:
-            raise MatrixArkError("local_context items must be strings or objects")
-        text = clip_context_text(text)
-        if not text:
-            continue
-        item_tokens = token_count(text)
-        token_total += item_tokens
-        text_hashes.update(context_text_hashes(text))
-        items.append(
-            {
-                "ref_type": ref_type,
-                "source": source,
-                "text": text,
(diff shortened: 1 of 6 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Nine functions that existed as two byte-identical copies each are now one shared object, 482 lines removed; the module keeping the body is unchanged and the other imports it. Direction was derived from the module graph so no import edge is added. One pair, build_shared_context_policy, is deliberately left duplicated because three tests assert those two copies are distinct objects - changing what they assert wants its own review.

PR #1182, opened 2026-09-07
Thirty-seven more functions written out twice, now written once

bjmeetsfo:
Thirty-seven more functions that existed as two byte-identical copies are now one. The body stays
where it is; the other module imports it.

## Identical bodies were not a sufficient test, and that is the interesting part

Seventy-three pairs passed every structural check — same signature, module level, undecorated, and
an import direction that adds no edge to the module graph. **Thirty-six of them still had to be
refused**, because a function's free names resolve in whichever module holds it. The same source
text calls a different helper, or raises a different class, depending on where it lives:

```
optional_object: global 'MatrixArkError' differs
    matrixark_mcp_validation    -> matrixark_mcp_errors.MatrixArkError
    matrixark_mcp_core_identity -> matrixark_mcp_core_identity.MatrixArkError
```

Consolidating that one changed which exception `normalize_envelope` raises. The payload was still
correctly rejected, but `except` compares by identity, so the handler stopped matching and a
passing test began reporting an unhandled error. `test_ingest_envelope_schema` states the invariant
in its own import comment: *"Import both from the same module so assertRaises matches the raised
type."* The rest of the refusals are the same shape — each module carries its own `stable_hash`,
`safe_identifier`, `clamp01`, so collapsing would silently switch which one runs.

The check compares what each module actually binds every free name to, at runtime:

- **annotations are skipped.** Two modules each writing `Json = dict[str, Any]` bind distinct
  objects of identical meaning, and under `from __future__ import annotations` they are never
  evaluated. Counting them refused two thirds of the work for no behavioural reason.
- **classes and functions compare by identity**, because `except` does, and two functions are
  interchangeable only if they are the same object.
- **everything else compares by equality**, since equal values behave identically.

Six more pairs were refused because a test guards that duplication on purpose — asserting the two
copies are distinct objects, so that a loop covering both cannot silently collapse to covering one
twice.

## Verification

By **object identity**: `owner.f is importer.f` for all 46 consolidations now in the tree, with
every module's `__all__` still resolving. Comparing sources, or calling both and comparing answers,
would pass equally well with two copies still present.

The whole `tools/` suite (381 files) ran before and after at the same commit: **346 pass / 39 fail
on both, with zero tests failing only after**. All 39 remaining failures fail identically on both
trees.

### DIFF
--- a/tools/matrixark_agent_hook.py
+++ b/tools/matrixark_agent_hook.py
@@ -464,29 +464,10 @@ def agent_context_from_payload(payload: Json, *, agent: str, event: str, session
     }
 
 
-def payload_resource_uri(payload: Json) -> str:
-    return first_string_at(
-        payload,
-        [
-            ["raw_uri"],
-            ["rawUri"],
-            ["uri"],
-            ["url"],
-            ["path"],
-            ["file_path"],
-            ["filePath"],
-            ["resource_path"],
-            ["resourcePath"],
-            ["document_path"],
-            ["documentPath"],
-            ["params", "raw_uri"],
-            ["params", "uri"],
-            ["params", "path"],
-            ["metadata", "raw_uri"],
-            ["metadata", "uri"],
-            ["metadata", "path"],
-        ],
-    )
+try:  # the implementation lives in matrixark_codex_hook; this module re-exports it
+    from .matrixark_codex_hook import payload_resource_uri
+except ImportError:  # Direct script execution from tools/.
+    from matrixark_codex_hook import payload_resource_uri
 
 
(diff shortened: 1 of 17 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. 37 more duplicated function bodies collapsed to one shared object each, 436 lines. The notable part is what was REFUSED: of 73 pairs that were byte-identical and passed every structural check, 36 had to be left alone because their free names resolve per module - optional_object binds MatrixArkError to a different class in each of its two homes, so consolidating it changed which exception normalize_envelope raises and a test stopped catching it.

PR #1183, opened 2026-09-07
Inspect page slabs in parallel at engine open, a batch at a time

bjmeetsfo:
Reconciling the band manifest at engine open reads each page slab and re-verifies the sha256 of
every record in it — one slab at a time, on one core, with the rest of the machine idle. It is on
the critical path of every start.

The per-slab work is independent and pure, so it fans out across a scoped thread pool. The band-map
update loop is untouched and still walks slabs in `live_slab_ids` order, so it makes the same
decisions in the same sequence and produces the same manifest.

### Batched, because the obvious version costs memory

Fanning out over the whole slab list first is the same speed but holds **every** slab's report until
the update loop runs, where the serial version consumed each immediately:

| | mean cold start | peak RSS |
|---|---|---|
| serial (main) | 20.7 s | 384 MB |
| parallel, whole list | 14.8 s | **964 MB** |
| **parallel, batched** | **14.0 s** | **527 MB** |

Lowering the worker count did not help the memory (8 workers and 3 workers both sat near 1.15 GB),
which is the tell: the cost is the retained reports, not concurrent slab buffers. Inspecting one
batch, applying its updates, then dropping it keeps the speed and most of the memory.

### Measurement

The page cache is dropped before **every** run. Without that this is unmeasurable here — the same
engine on the same store measured 116 s, 45 s and 39 s, a 3x swing, and an earlier version of this
change was wrongly discarded as "inside the noise" on exactly that basis.

With the cache controlled, arms alternated:

```
old  22943ms  peak=384MB      new  13882ms  peak=527MB
new  14140ms  peak=523MB      old  18374ms  peak=385MB
```

**1.47x** on the means. The batched arm is also the most consistent measured all day (1.02x spread
against 1.25x for serial), which matters on a path every start pays.

### Tests

`cargo test --release -p temporalstore-rust --lib block_store` — 68 passed.

### DIFF
--- a/crates/temporalstore-rust/src/block_store/band_manifest.rs
+++ b/crates/temporalstore-rust/src/block_store/band_manifest.rs
@@ -113,24 +113,59 @@ pub(super) fn reconcile_band_manifest_with_disk(
         .copied()
         .unwrap_or_default();
 
-    for page_slab_id in &live_slab_ids {
-        let path = slab_path(root, *page_slab_id);
-        let bytes = fs::read(&path)?;
-        let report = inspect_slab(&bytes, *page_slab_id);
-        let desired_state = if *page_slab_id == latest {
+    // Read and hash the slabs in parallel, then apply the band updates in the original order.
+    //
+    // Inspecting a slab re-verifies the sha256 of every page record in it, and this runs at EVERY
+    // engine open. Measured on a live store: ~950 MB of slabs took about 70 s of a cold start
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Engine open re-verified every page record one slab at a time on a single core; it now fans out a batch at a time. 1.47x on cold start (20.7s -> 14.0s) for +143 MB peak, measured with the page cache dropped before every run - without that control the same engine swung 116/45/39s on one store, which is why an earlier version of this was wrongly discarded as noise.

PR #1185, opened 2026-09-07
The shared-flag floor follows the consolidation it counts

bjmeetsfo:
`test_the_scan_still_finds_shared_flags` asserts the scan finds at least EXPECTED_SHARED_FLOOR flags
read from more than one production site. It finds 14 and the floor says 15, so it fails on main --
and because the suite runs against a recorded baseline, it then fails on every open branch as a new
failure on work that never touched it.

Nothing is broken. The number counts flags read from MORE THAN ONE site, so consolidating a
duplicate reader lowers it by design, and two merges just did:

  * #1171 removed four time-index builders nothing called. One held the second read of
    MATRIXARK_CONTEXT_EVENT_TIME_INDEX_FULL_PAYLOAD.
  * #1182 removed thirty-seven functions written out twice. One held the second read of
    MATRIXARK_SEGMENT_PROVIDER_FALLBACK.

Both flags are still read. Each is now read once, which is what both merges were for.

Lowering the floor is what this guard expects rather than a weakening of it: the comment beside it
already records twenty-one when it was written and fifteen afterwards. This makes it fourteen and
writes down the part that was missing -- that a change removing a duplicate reader should lower it in
the same change, because leaving it stale does not fail that change, it fails everybody else's.

The guard keeps doing its job. It exists so a scan that quietly stops matching fails loudly instead
of reporting that every remaining reader agrees, and `test_no_two_readers_disagree` still passes on
the fourteen it finds.

### DIFF
--- a/tools/test_flag_readers_agree.py
+++ b/tools/test_flag_readers_agree.py
@@ -45,9 +45,21 @@ _ENV_BOOL = re.compile(
     r'env_bool\(\s*["\']([A-Z][A-Z0-9_]{3,})["\']\s*,\s*(True|False)\s*\)')
 ONE_VOCABULARY = ("<env_bool>",)
 
-# Twenty-one when this was written. Asserted so a scan that stops matching fails rather than
-# reporting that every reader agrees.
-EXPECTED_SHARED_FLOOR = 15
+# Twenty-one when this was written, fifteen a while later, fourteen now. Asserted so a scan that
+# stops matching fails rather than reporting that every reader agrees.
+#
+# This counts flags read from MORE THAN ONE production site, so it falls whenever a duplicate reader
+# is consolidated away -- which is work being done deliberately, not damage. The last two both came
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Superseded, and by a better version than this one. main already sets the floor to 14, and its note names the cause precisely -- mx#1171 removing `matrixark_mcp_event_keys.context_event_time_index_entries`, the second production reader of MATRIXARK_CONTEXT_EVENT_TIME_INDEX_FULL_PAYLOAD -- along with the rule that each step down needs a named cause. That is what this branch was for, so it has nothing to add.

One measured detail its record does not yet cover, in case it is useful: the set also lost MATRIXARK_SEGMENT_PROVIDER_FALLBACK, whose second reader was `matrixark_mcp_segments.py:55`, removed by mx#1182. Scanning at main~8 finds 16 shared flags, at main~4 finds 15, and now 14 -- two steps, not one.

PR #1193, opened 2026-09-07
A datanode URL with no scheme is discarded, host and port

bjmeetsfo:
`urlparse("datanode:9000")` reads `datanode` as the **scheme** and leaves the netloc empty, so both `hostname` and `port` come back empty and the defaults win:

```
configured  MATRIXARK_DATANODE_URL=datanode:9000
the gateway talks to  127.0.0.1:17102
```

The configured host **and** port are discarded, and nothing said so.

It is not a failure anybody sees. `127.0.0.1:17102` is a plausible address: on a box where something answers there, the deployment appears to work against the wrong datanode; where nothing answers, the readiness checklist reports *"The gateway could not connect to the datanode"* and tells the reader to check that their datanode process is running — which it is, at the address they configured and the gateway never used.

## What changes

- the parse is one named function, `datanode_target()`. It was three lines inside `GatewayConfig.__init__`, and the snapshot needs the same answer — a second copy would be a second thing to be wrong, which is how the connection and the report come to disagree.
- the model snapshot carries `datanode_address`: what was **configured**, what is **effective**, and whether the URL was usable.
- a warning fires only when a URL was set and yielded no host, naming the value, the address calls are actually going to, and the remedy.

| configured | effective | warned |
|---|---|---|
| `http://datanode:9000` | `http://datanode:9000` | no |
| `datanode:9000` | `datanode://127.0.0.1:17102` | **yes** |
| *(unset)* | `http://127.0.0.1:17102` | no — that is the documented default |

**The fallback is deliberately unchanged.** Pointing the connection at the host the operator wrote would move live traffic on the strength of a string that has never parsed successfully. Saying where the calls are actually going is the part that is unambiguously right, and it makes the misconfiguration obvious in one line. Redirecting can be a separate decision.

## Mutations

Eleven tests. Each of these reddens at least one:

| mutation | reddens |
|---|---|
| the snapshot stops reporting the address | `test_a_good_url_reports_itself` |
| it always calls the URL usable | `test_a_url_with_no_scheme_yields_no_host` |
| the warning never fires | `test_the_warning_names_the_value_the_address_and_the_remedy` |
| it warns on a good URL too | `test_a_usable_url_is_not_warned_about` |
| the warning stops naming the remedy | `test_the_warning_names_the_value_the_address_and_the_remedy` |
| the warning stops naming where calls go | `test_the_warning_names_the_value_the_address_and_the_remedy` |
| the config object parses it separately again | `test_the_config_object_uses_it` |
| a malformed port raises again | `test_a_malformed_port_does_not_raise` |

That last one is a real edge the inline version had: `urlparse().port` **raises** `ValueError` on a bad port rather than returning `None`, and this runs while a request is being served.

### DIFF
--- a/tools/matrixark_v1_gateway.py
+++ b/tools/matrixark_v1_gateway.py
@@ -412,10 +412,8 @@ class GatewayConfig:
         # Enforced-mode hashed keystore: {api_key_hash -> {tenant_id, account_id}}.
         self.hashed_keys: dict[str, Json] = fields.get("hashed_keys", {})
         # Datanode blob target, parsed for the http.client proxy.
-        parsed = urlparse(str(self.datanode_url))
-        self.blob_scheme = parsed.scheme or "http"
-        self.blob_host = parsed.hostname or "127.0.0.1"
-        self.blob_port = parsed.port or (443 if self.blob_scheme == "https" else 17102)
+        self.blob_scheme, self.blob_host, self.blob_port, self.datanode_url_usable = \
+            datanode_target(self.datanode_url)
         # Injectable so tests can proxy without a network; production builds an http.client conn.
         self.blob_connection_factory: Callable[["GatewayConfig"], Any] = fields.get(
             "blob_connection_factory", _default_blob_connection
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. A datanode URL with no scheme is discarded, host and port

PR #1197, opened 2026-09-07
The third model role is alertable

bjmeetsfo:
Two of the three model roles are alertable. The one called most was not.

`matrixark_gateway_extraction_model_active` and `matrixark_gateway_embedding_semantic` are gauges with panels and alerts, for the reason `config_health_lines` gives in its own docstring: a misconfigured deployment is indistinguishable from a healthy one at the API surface, so a dashboard charting only rate and latency shows it as perfectly healthy for months.

Summaries had **no metric, no panel and no alert** — and they are the role called most: extraction runs once per ingest, and *every* context node gets a summary that retrieval then walks.

## Why it is a third series and not a label on an existing one

| extraction provider | `extraction_model_active` | `summary_model_active` |
|---|---|---|
| `openai_compatible` | 1 | 1 |
| **`anthropic`** | **1** | **0** |
| `deterministic` | 0 | 0 |

An Anthropic deployment is a real model for extraction *and* returns rule-written summaries. On the existing metric it reads as fully model-backed while every node summary is written by rules.

The set of names `config_health_lines` uses for the other two **cannot express that** — the same provider name means 1 on one line and 0 on this one — so the value is taken from the snapshot's own `summary.writes`, decided by the one classifier that mirrors the summariser. A snapshot without a summary block reports `0`: a metric that invents a healthy answer is worse than one that is missing, because it is charted as fact.

## Layout

The three roles now sit on one row (`Retrieval mode`, `Extraction`, `Summaries`), with `Config warnings` moved right and `Requests in flight` taking the free slot on the datanode row. Three separate answers to "is this deployment using a model" are only useful side by side.

## Mutations

Fourteen tests. Each of these reddens at least one:

| mutation | reddens |
|---|---|
| the series is not emitted | `test_a_model_deployment_reports_one` |
| it copies the extraction answer | `test_anthropic_is_a_model_for_extraction_and_rules_for_summaries` |
| a missing snapshot reports healthy | `test_no_snapshot_reports_zero_rather_than_one` |
| the alert is removed | `test_an_alert_fires_on_it` |
| the alert watches the wrong series | `test_an_alert_fires_on_it` |
| the panel is removed | `test_a_panel_charts_it` |
| the three roles are split across rows | `test_the_three_roles_sit_together` |
| the new panel lands on another one | `test_no_two_panels_overlap` |

`test_it_does_not_simply_track_extraction` is the floor for the whole change: the two series must disagree somewhere, or one of them is noise.

Only the gateway dashboard and the gateway alert rules are touched — the engine dashboard is untouched.

### DIFF
--- a/docs/ops/matrixark-gateway-dashboard.json
+++ b/docs/ops/matrixark-gateway-dashboard.json
@@ -239,6 +239,56 @@
       ]
     },
     {
+      "id": 18,
+      "type": "stat",
+      "title": "Summaries",
+      "description": "Whether node summaries come from a model. Separate from Extraction because they can differ: an Anthropic extraction provider returns rule-written summaries.",
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${DS_PROMETHEUS}"
+      },
+      "gridPos": {
+        "h": 4,
+        "w": 6,
+        "x": 12,
+        "y": 4
+      },
+      "fieldConfig": {
+        "defaults": {
+          "mappings": [
+            {
+              "type": "value",
+              "options": {
+                "0": {
+                  "text": "rules only",
+                  "color": "red",
+                  "index": 0
+                }
+              }
+            },
+            {
+              "type": "value",
+              "options": {
+                "1": {
+                  "text": "model",
+                  "color": "green",
+                  "index": 1
+                }
+              }
+            }
+          ]
+        },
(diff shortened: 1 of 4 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The third model role is alertable

PR #1199, opened 2026-09-07
Twenty-six more functions written out twice, now written once

bjmeetsfo:
Twenty-six more functions that existed as several byte-identical copies are now one. The body stays
where it is and the other modules import it.

This round extends the analysis in three ways, each because something went wrong without it.

## Groups of more than two

The rules were never specific to pairs, but the tool only handled them, so `stable_hash` — three
copies — sat in a skip bucket **while blocking three other groups**, which could not consolidate
while a helper they both call resolved differently in each module. One copy owns the body and every
other imports it. Generalising unlocked a cascade, and the work now runs to a fixpoint: refusals
are transitive, so consolidating a helper unblocks its callers on the next pass. Three passes
converged.

## Cycles are checked cumulatively

Each group was checked against the import graph as it was at the start of the pass. Two groups can
each be cycle-free alone and together close one — which is how `matrixark_mcp_core` ended up
partially initialised, a circular import that no single decision was responsible for. Every
accepted edge now joins the working graph before the next group is considered.

## A consolidation must not change what production reaches

Choosing an owner inside a module production does not reach makes the importer depend on it, which
turns a dead island live. That is an architectural change and not something to do as a side effect
of removing duplication. `test_a_module_only_tests_reach_is_not_live` records that set — 7 islands,
43 modules — and it noticed:

```
NO LONGER UNREACHABLE (wired up or deleted -- take them out of the list):
  ['matrixark_mcp_session_runtime']
```

The tempting response is to edit the recorded list. The right one is to leave the island alone.
That guard's own record is now what the analysis consults; approximating it as "some non-test
module imports it" is a different question, because a module imported only by other unreachable
modules still has inbound imports — and a second module was wired up before that became clear.

## Still refused

Twenty-six groups remain, and the reasons are worth stating because they are not oversights:

- **11 blocked by `MatrixArkError`**, which is two classes in different modules rather than a
  duplicated function. Collapsing exception classes changes what `except` clauses catch and is its
  own change.
- the rest read a helper — `safe_identifier`, `now_ms`, `tokens` — that each module binds
  differently, so consolidating would switch which one runs. Some of those helpers are themselves
  genuinely different functions that happen to share a name.

## Verification

By **object identity**: `owner.f is importer.f` for all 72 consolidations in the tree, every
module's `__all__` still resolving, and every file still parsing.

The whole `tools/` suite (381 files) ran before and after, with the control **pinned to this exact
commit** so main cannot move underneath it: **345 pass / 40 fail on both, zero tests failing only
after.** All 40 remaining failures fail identically on both trees.

### DIFF
--- a/tools/convert_external_baseline_csv_to_benchmark_report.py
+++ b/tools/convert_external_baseline_csv_to_benchmark_report.py
@@ -389,12 +389,10 @@ def to_float(value: str | None) -> float | None:
         return None
 
 
-def percentile(values: list[float], pct: float) -> float | None:
-    if not values:
-        return None
-    ordered = sorted(values)
-    index = int((len(ordered) - 1) * pct / 100.0)
-    return round(ordered[index], 4)
+try:  # the implementation lives in summarize_oss_memory_benchmark_artifacts; this module re-exports it
+    from .summarize_oss_memory_benchmark_artifacts import percentile
+except ImportError:  # Direct script execution from tools/.
+    from summarize_oss_memory_benchmark_artifacts import percentile
 
 
 def safe_div(numerator: float, denominator: float) -> float:
--- a/tools/matrixark_mcp_core_compact.py
+++ b/tools/matrixark_mcp_core_compact.py
@@ -252,29 +252,10 @@ def attach_storage_route(record: Json) -> Json:
     return record
 
 
-def context_placement_key(record: Json, *, scope_key: str = "", node_hash: Any = None) -> str:
-    explicit = str(record.get("placement_key") or "")
-    if explicit:
-        return explicit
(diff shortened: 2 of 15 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. 26 more duplicated bodies collapsed, 309 lines. Three checks were added because each caught a real fault: cycles are now checked cumulatively within a pass (two groups can each be cycle-free alone and together close one, which left matrixark_mcp_core partially initialised); an owner is never chosen inside a module production does not reach, taken from the guard that records that set rather than approximated; and groups of more than two copies are handled, which unblocked a cascade that stable_hash was holding up.

PR #1201, opened 2026-09-07
The diagnostics bundle dates itself against the browser

bjmeetsfo:
The diagnostics bundle is the file that leaves the building. It is downloaded and sent to somebody who was not at the keyboard, cannot ask which machine collected it, and cannot ask whether the page was still receiving anything at the time. It carried neither answer.

## It dated itself with the browser's clock

`collected_at` came from `new Date()`, while `recent_failures` **inside the same file** carries gateway timestamps — so a reader lining the two up was comparing two machines, with nothing in the document saying so. It now carries the gateway's own time for the same instant and the difference between them.

## It did not say how fresh its live half was

`recent_failures` and `datanode` come off the stream; everything else is fetched when the button is pressed. On a stalled stream the file holds one current half and one arbitrarily old half, presented as one document. It now records whether the stream was connected, how long since anything arrived, and whether that counts as stalled.

| state when collected | `gateway_time` | `clock_skew_ms` | `stream` |
|---|---|---|---|
| live, clocks agree | same as `collected_at` | `0` | connected, not stalled |
| live, browser 60s off | differs | `60000` | connected, not stalled |
| stalled 8s | present | present | `since_last_block_ms: 8000`, `stalled: true` |
| no frame ever arrived | **`null`** | **`null`** | `connected: false`, `since_last_block_ms: null` |

**Both clock answers are absent rather than guessed** when no frame has arrived. Repeating the browser's clock under a second name would be worse than saying nothing, because it would read as corroboration.

## The probe

`bundle_clock_harness.js`, not `bundle_harness.js`. The latter already existed, drives the whole overview page and reports its own `ok`/`FAIL` lines — and writing over it took four of its tests with it. That was caught by running the **full ratchet** on the branch rather than the suites this change touches, which is also how the `event: hello` mistake in #1189 surfaced.

## Mutations

Eleven tests, driven through the shipped `bundle()` with a real stream in front of it. Each of these reddens at least one:

| mutation | reddens |
|---|---|
| the gateway's time is dropped | `test_it_carries_the_gateways_time_as_well_as_the_browsers` |
| the skew is dropped | `test_a_skewed_browser_is_recorded_not_hidden` |
| with no frame it repeats the browser's clock | `test_with_no_frame_it_says_nothing_rather_than_guessing` |
| the gateway time is silently the browser's | `test_a_skewed_browser_is_recorded_not_hidden` |
| the stream block is dropped | `test_a_stalled_stream_is_recorded_as_stalled` |
| a stalled stream is recorded as fine | `test_a_stalled_stream_is_recorded_as_stalled` |
| never-connected reads the same as quiet | `test_never_connected_is_not_the_same_as_quiet` |
| the failures off the frame stop being carried | `test_the_failures_off_the_frame_still_arrive` |

That last one is the floor: everything the bundle already carried has to still be in it.

### DIFF
--- a/tools/portal/build_portal_pages.py
+++ b/tools/portal/build_portal_pages.py
@@ -734,6 +734,12 @@ function sinceLastBlockMs() {
   return liveBlockAtMs ? (Date.now() - liveBlockAtMs) : 0;
 }
 
+/* Whether a stream has ever delivered anything to this page. Read by code that runs on a button
+   press rather than on a frame, which has no other way to ask. */
+function streamConnected() {
+  return liveBlockAtMs > 0;
+}
+
 /* Three ticks: one missed tick is scheduling, two is a slow one, three is not arriving. Derived
    from the cadence the gateway sent rather than a number written here, because a page holding its
    own copy of the server's interval is a second place for it to be wrong. Unknown cadence means an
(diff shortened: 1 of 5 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The diagnostics bundle dates itself against the browser

PR #1203, opened 2026-09-07
One MatrixArkError, not two

bjmeetsfo:
`class MatrixArkError(ValueError): pass` is written out twice — in `matrixark_mcp_errors` and
again in `matrixark_mcp_core_identity`. The bodies are identical, but they are two different
classes, and `except` compares by identity: a handler bound to one does not catch an exception
raised through the other.

That split is reachable rather than merely untidy. Both sides have raise sites AND except sites:

| class in use | modules | raise sites | except sites |
|---|---|---|---|
| via `matrixark_mcp_core` → `core_identity` | 31 | 107 | 7 |
| `matrixark_mcp_errors` | 16 | 55 | 6 |
| `core_identity` directly | 2 | 8 | 0 |

Nothing prevents an except site in one group from receiving a raise from the other; it just
depends on which call path runs. The failure mode is not hypothetical either — moving a validation
helper between these two modules produced exactly it, and a test that expects a payload to be
rejected instead reported an unhandled error, because the rejection arrived as the other class.

`matrixark_mcp_errors` owns it. It has the most importers, and it is a leaf — it imports only
`__future__` and `typing`, so nothing can cycle through it. `core_identity` already reaches it, so
this adds no edge to the module graph. The two subclasses defined there,
`MatrixArkInvalidRequestError` and `MatrixArkNotFoundError`, rebase onto the unified class, which
is the point.

Unifying only ever makes an `except` catch more, never less: both classes derive from `ValueError`
and both bodies are `pass`. No test depends on the two being distinct — checked for the shapes that
would (comparing identity, or importing the name from both sources).

## The rest of the change

With one class, several duplicated functions that read `MatrixArkError` as a free name became
consolidatable — their bodies were byte-identical but resolved that name to different classes
depending on which module held them, which is a behaviour difference rather than duplication. Five
of those are now one shared object each.

## Verification

By object identity: the class is one object through every path that reaches it
(`core is errors`, `validation is errors`), where before `core is validation` was false. All 78
function consolidations in the tree resolve to one shared object, every module's `__all__` still
resolves, and every file still parses.

The whole `tools/` suite (381 files) ran before and after with the control pinned to this exact
commit.

### DIFF
--- a/tools/matrixark_mcp_core.py
+++ b/tools/matrixark_mcp_core.py
@@ -881,18 +881,10 @@ def build_segment_prompt(messages: list[Json]) -> str:
     )
 
 
-def parse_first_json_object(text: str) -> Json:
-    decoder = json.JSONDecoder()
-    for index, char in enumerate(text):
-        if char != "{":
-            continue
-        try:
-            value, _end = decoder.raw_decode(text[index:])
-        except json.JSONDecodeError:
-            continue
-        if isinstance(value, dict):
-            return value
-    raise MatrixArkError("model response did not contain a JSON object")
+try:  # the implementation lives in matrixark_mcp_extraction_provider; this module re-exports it
+    from .matrixark_mcp_extraction_provider import parse_first_json_object
+except ImportError:  # Direct script execution from tools/.
+    from matrixark_mcp_extraction_provider import parse_first_json_object
 
 
 def oss_model_memory_segments(messages: list[Json], *, model: str, model_path: str = "", max_new_tokens: int = 512, local_only: bool = False) -> Json:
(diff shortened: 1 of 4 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. MatrixArkError was defined twice with identical bodies, and except compares by identity, so a handler bound to one could not catch an exception raised through the other. Both sides had raise AND except sites (107/7 and 55/6), so the split was reachable rather than untidy. matrixark_mcp_errors owns it now; it is a leaf module so nothing can cycle through it, and unifying only ever makes an except catch more.

PR #1204, opened 2026-09-07
The strip notices a quiet stream too

bjmeetsfo:
#1189 taught the page-side stream client that a connection which has stopped delivering is not a live one. **The strip runs its own client** — five of the seven portal pages never load the other one — so on those five the dot stayed green however long the gateway had been silent.

The strip is the one element on *every* page. It is the thing somebody glances at to answer "is this deployment alive", and it was answering "the socket is open".

## The rule, and why silence alone is not it

When nothing has changed the gateway sends `: keepalive` instead of a frame, so a quiet stream on an idle deployment is **healthy**. What is not normal is nothing at all — no frame and no keepalive — for three ticks of the gateway's own cadence, which every frame carries since #1189.

| | dot | title |
|---|---|---|
| quiet 4s at a 2s cadence | `live-dot` | live |
| quiet 8s at a 2s cadence | `live-dot stale` | `no update for 8s` |
| then anything arrives | `live-dot` | live |
| quiet 8s at a **10s** cadence | `live-dot` | live |
| gateway states no cadence | `live-dot` | live — nothing claimed |

`stale` is already in the strip's vocabulary and already amber; it simply never reached it while connected.

## Two clients, one rule

The strip cannot use the shared client — it has to work on pages that never load it — so the rule is now written twice, which is a thing that drifts. `TheTwoClientsAgreeTest` is the answer: both must read `tick_s` off a frame, both must wait `tickMs * 3` with the interval **taken from the gateway rather than written down**, and the strip must stamp on *every block* rather than every frame, since a keepalive is what says an idle stream is alive.

## Mutations

Fifteen tests, driven through the shipped strip client with a held clock. Each of these reddens at least one:

| mutation | reddens |
|---|---|
| the strip never stamps a block | `test_four_ticks_of_quiet_turns_it` |
| it stamps only on status frames | `test_anything_arriving_turns_it_back` |
| the watchdog never starts | `test_four_ticks_of_quiet_turns_it` |
| the rule writes the interval down | `test_a_slower_gateway_gets_a_longer_leash` |
| one tick counts as a stall | `test_both_wait_three_ticks` |
| the strip stops reading the cadence | `test_both_read_the_cadence_off_a_frame` |
| the dot never comes back | `test_anything_arriving_turns_it_back` |

Three of the agreement tests were wrong before they were right — they sliced fixed character windows that missed the code they meant to check, and one banned four-digit literals and caught `Math.round(quiet / 1000)`. They now cut by matching braces from a named anchor, and `test_both_bodies_are_really_the_clients` is the floor that stops an empty region passing everything.

### DIFF
--- a/tools/portal/api_key_portal.html
+++ b/tools/portal/api_key_portal.html
@@ -1274,6 +1274,22 @@
   if (window.__matrixarkLive) { return; }
   window.__matrixarkLive = "strip";
 
+  /* Nothing arrives to announce that nothing is arriving, so noticing needs a timer. Three ticks
+     of the gateway's OWN cadence: one missed tick is scheduling, two is a slow one, three is not
+     arriving. The number is not written here -- it is whatever the gateway last said. */
+  function watchQuiet() {
+    if (!tickMs || !blockAt || !controller) { return; }
+    var quiet = Date.now() - blockAt;
+    if (quiet > tickMs * 3 && dot.className.indexOf("stale") < 0) {
+      dot.className = "live-dot stale";
+      dot.title = "no update for " + Math.round(quiet / 1000) + "s";
+    }
+  }
+  quietTimer = setInterval(watchQuiet, 1000);
+  window.addEventListener("pagehide", function () {
+    if (quietTimer) { clearInterval(quietTimer); quietTimer = null; }
+  });
+
   function key() {
     var el = document.getElementById("key");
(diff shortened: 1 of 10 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The strip notices a quiet stream too

PR #1208, opened 2026-09-07
Eighteen short helpers written out several times, now written once

bjmeetsfo:
Eighteen more functions that existed as several byte-identical copies are now one. This round is
mostly short helpers, and it does not save lines — a two-line helper replaced by a four-line import
block adds two. What it buys is one definition instead of several, and the larger groups those
helpers were blocking.

## Why short functions were excluded before, and what actually distinguishes them

The analysis ignored functions under four lines, as a guard against trivial collisions. That is a
heuristic rather than a safety rule, and it was hiding real helpers: `now_ms` is two lines,
`safe_identifier` and `stable_hash` three. While those differed between modules, every larger
function that calls them was unconsolidatable — their bodies were identical but resolved that call
to a different implementation depending on which module held them.

Simply lowering the threshold is wrong in the other direction. It admits `normalize`, `compact` and
`content_hash`: two-line helpers with identical bodies in scripts that have nothing to do with each
other, where an import would couple unrelated modules to save two lines.

So the rule is not about length. **A short function may only be consolidated where the importer
already reaches the owner**, so no new import edge appears. That refuses ten coincidental matches
and admits the helpers that sit in modules already connected. The cascade then pays for itself:
fourteen consolidations in the first pass released four more in the second.

## Copies of one body can have different names

Groups are keyed by body, so `_now_ms` and `now_ms` land in the same group. Taking the name from
one copy and importing it into another leaves the importer's own name unbound, and every caller
gets a `NameError`. Every copy in a group must now share the name. Importing under an alias would
work, but a private `_now_ms` beside a public `now_ms` is a distinction someone made, and erasing
it is not this change's job.

## Verification

By object identity: `owner.f is importer.f` for all 96 consolidations in the tree, every module's
`__all__` still resolving, and every file still parsing.

The whole `tools/` suite (381 files) ran before and after with the control pinned to this exact
commit: **357 pass / 36 fail on both, zero tests failing only after.** All 36 remaining failures
fail identically on both trees.

## What stays refused

Eighteen groups remain, and the refusals are answers rather than omissions. They read helpers — `tokens`
among them — that are genuinely different functions sharing a name, so consolidating would change
which one runs.

### DIFF
--- a/tools/matrixark_agent_hook.py
+++ b/tools/matrixark_agent_hook.py
@@ -778,8 +778,10 @@ def normalized_session_buffer_from_ingest(ingest: Json | None) -> Json:
     return summary
 
 
-def hook_storage_options() -> Json:
-    return {"route": os.environ.get("MATRIXARK_HOOK_STORAGE_ROUTE", "shared_store_async")}
+try:  # the implementation lives in matrixark_codex_hook; this module re-exports it
+    from .matrixark_codex_hook import hook_storage_options
+except ImportError:  # Direct script execution from tools/.
+    from matrixark_codex_hook import hook_storage_options
 
 
 def require_retrieval_memory_coverage(value: Any) -> bool:
--- a/tools/matrixark_mcp_core.py
+++ b/tools/matrixark_mcp_core.py
@@ -3620,8 +3620,10 @@ except ImportError:  # Direct script execution from tools/.
     from matrixark_mcp_scoring import clamp01
 
 
-def normalized_dense_score(value: float) -> float:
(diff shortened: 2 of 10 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Short helpers this time, so it saves no lines - a two-line helper replaced by a four-line import block adds two. The point is one definition instead of several, plus the larger groups those helpers were blocking. The rule that made it safe is connectivity rather than length: a short function is only consolidated where the importer already reaches the owner, which refuses ten coincidental matches like normalize and compact in unrelated scripts.

PR #1212, opened 2026-09-07
Every append copies its outcome payload before the encoder copies it again

bjmeetsfo:
`record_parts` builds the proto tail with `items: record.outcomes.iter().map(item_to_proto).collect()`.
Each outcome carries an owned `value: Option<Vec<u8>>`, and `item_to_proto` copies it into the proto
message. The encoder then copies it again, into the frame buffer that goes to disk.

That is the shape `staged_blocks` was taken out of this same struct for. The comment two lines below
still records why: filling it "copied every one of them before the encoder copied them again".

Measured with the allocation probe, driving the borrowing writer the way the append path does:

    outcome value      allocated per append      overhead
        4,096 B                4,489 B              393 B
       16,384 B               16,777 B              393 B

The rows differ by 12,288 bytes, which is exactly the difference in value size. So the copy is one
for one: a fixed 393 bytes of conversion, plus a whole second copy of whatever the outcome carries.
On a page write that is the page.

This change does not fix it. Fixing it means what fixing `staged_blocks` meant -- a hand-written
encoder and a hand-written length for the field, checked byte for byte against the ordinary path by
`the_borrowing_encoder_writes_the_same_bytes`. That was tractable for a staged block, which is two
scalars. An outcome item has ten fields and a nested message, and proto3 default omission has to be
reproduced exactly or records already on disk stop decoding. It deserves its own change rather than
being tacked onto a measurement.

What lands here is the measurement and a note at the field, so the next person finds a number
instead of rediscovering it: `does_the_outcome_value_get_copied`, behind the `alloc-probe` feature.

### DIFF
--- a/crates/temporalstore-rust/src/wal_proto.rs
+++ b/crates/temporalstore-rust/src/wal_proto.rs
@@ -816,6 +816,15 @@ mod tests {
                 .map(metadata_to_proto)
                 .transpose()
                 .unwrap(),
+            // `items` is still filled the ordinary way, and it has the shape `staged_blocks`
+            // below was taken out of this struct for: every outcome carries an owned
+            // `value: Option<Vec<u8>>`, and `item_to_proto` copies it here before the encoder
+            // copies it again into the frame. Measured by `does_the_outcome_value_get_copied`, the
+            // bytes allocated per append track the outcome value one for one -- 4,489 B for a
+            // 4,096 B value and 16,777 B for a 16,384 B one, a fixed 393 B either side of a whole
+            // extra copy. Fixing it means what fixing `staged_blocks` meant: a hand-written encoder
+            // and a hand-written length for this field, which is a larger job here because an item
+            // has ten fields and a nested message rather than two scalars.
             items: record.outcomes.iter().map(item_to_proto).collect(),
             staged_blocks: record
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
One dead end checked, so the next person does not have to.

The cheap version of this fix does not work: build the proto item with `value: None`, encode it, and append the value field by hand. That would be byte-identical only if `value` were the highest field number in `EngineWalItem`, because prost emits fields in ascending order. It is not -- `value` is 13, and `kind_name` = 14, `routing_bucket` = 15 and `kind_code` = 19 come after it. Appending would write those three before field 13.

Protobuf allows fields in any order and prost decodes them either way, so that version would work. It would also change the byte layout of every new record for a reason nobody reading the log later could infer, and it would break `the_borrowing_encoder_writes_the_same_bytes`, which exists to catch exactly that kind of drift.

So the fix is the whole ordered encoder for an item, the way `put_staged_block` is written for a staged page: every field in ascending number order, with a matching length function, checked byte for byte against `item_to_proto` across `cases()`. The derivation in `item_to_proto` has to come with it -- `numeric_component` decides three of the fields between them, the address dedup drops `object_id` when it repeats the item's, and `object_deleted` is derived rather than stored.

PR #1214, opened 2026-09-07
A context pack that finished late was deleted instead of served

bjmeetsfo:
A retrieve that completed was thrown away because it was slow, and the agent was told it had no
history.

`dispatch_matrixark_tool` runs the retrieve, gets a full ContextPack back, and then — if the call
took longer than the request deadline — discards that pack and substitutes
`_retrieve_timeout_fallback`:

```python
result = server.adapter.retrieve(args)          # completes; refs computed
timeout = request_deadline_ms > 0 and elapsed_ms >= request_deadline_ms
if timeout and not (result.get("partial_context_pack") or result.get("partial")):
    result = server._retrieve_timeout_fallback(...)   # replaces it
```

On a native backend that replacement builds its pack from `records = []`, so a complete answer
becomes an empty one. Nothing raises, and the empty pack carries a `context_pack_id` like any
other, so every downstream check reads it as a legitimate answer.

A deadline bounds how long a caller **waits**. By the time this branch runs the wait is already
over and the full cost is already paid, so overrunning it makes the answer *late* — not wrong, and
certainly not worth deleting. This keeps the pack, marks it late, and lets it through.

The comment above the deadline constant describes this same failure happening once before, at the
old 5000 ms ceiling: *"made the server discard the real ContextPack for an empty
deadline_fallback_pack (silent recall collapse: refs computed but dropped)"*. It was fixed by
raising the ceiling to 30000 ms, on the stated assumption that *"actual warm/cold retrieve latency
stays well under this ceiling"*. That assumption expires as a store grows. Measured on a 16,717-record
store at four secondary-index budgets:

| tenant / scope budget | retrieve | outcome |
|---|---|---|
| 1024 / 128 (default) | 32.1 s | over deadline, pack discarded |
| 4096 / 512 | 57.3 s | over deadline, pack discarded |
| 8192 / 1024 | 75.7 s | over deadline, pack discarded |
| 32768 / 4096 | 72.3 s | over deadline, pack discarded |

Every setting overruns the 30 s deadline, including the default — so no tuning of that budget could
avoid it, and raising the ceiling again would only postpone the same collapse behind a larger
number. The behaviour is what needed fixing.

### Why it stayed silent

`compact_context_pack_for_serving` renames `quality_warnings` to `warnings`, and what reaches a hook
is the compact pack — but both hooks read only `quality_warnings`. Every retrieval warning therefore
read as an empty list on every live turn: the deadline warning, the embedding-model-conflict warning,
and the underfill warning alike. The signal was produced, compacted under another name, and read for
by nobody.

`retrieval_warnings()` now accepts either spelling, and both hooks use it. Both spellings rather than
just the compact one, because the full pack is what comes back with `include_retrieval_debug`, and a
diagnostic that only works in production is the same trap facing the other way.

### Scope

The raising path is deliberately untouched: when `adapter.retrieve` raises there is no pack, and the
fallback is the only thing available to serve. `test_a_retrieve_that_never_finished_still_falls_back`
pins that boundary and passes both with and without this change, so a "fix" that merely deleted the
timeout branch would not get past it.

### Verified on a live box

One agent turn against the 16,717-record store, before and after:

| | context served | refs | warnings surfaced |
|---|---|---|---|
| before | 3 bytes | 0 | `[]` |
| after | 7,886 bytes | 24 | `["request_deadline_after_retrieve"]` |

`context_pack_source` is unset on the served turn, so this is a freshly retrieved pack rather than
the previous-pack fallback covering for it. The warning string had been produced on every turn for
hours and read as an empty list the whole time.

### Tests

`tools/test_a_late_context_pack_is_served_not_discarded.py`. Against unmodified code the central
test fails with `a completed pack was replaced by the empty deadline fallback`; the boundary test
passes on both arms. The stub adapter sleeps past the deadline on purpose — with an instant adapter
the timeout branch never runs and all of these pass against unmodified code, which is no coverage at
all.

Suites touching the changed files were run on both arms and are unchanged:
`test_codex_pipeline_part4`, `test_python_module_boundaries_part3` and `test_matrixark_codex_hook_pipeline`
carry identical pre-existing failure counts before and after; `test_a_setting_declares_the_default_the_code_applies`,
`test_a_retrieve_that_raises_is_recorded_not_fatal` and `test_the_callers_deadline_reaches_the_socket`
pass.

### DIFF
--- a/tools/matrixark_agent_hook.py
+++ b/tools/matrixark_agent_hook.py
@@ -545,8 +545,7 @@ def retrieval_session_identity_from_retrieve(pack: Json | None, *, session_id_so
 def retrieval_quality_warnings_from_retrieve(pack: Json | None) -> list[Any]:
     if not isinstance(pack, dict):
         return []
-    warnings = pack.get("quality_warnings")
-    return warnings if isinstance(warnings, list) else []
+    return _pack_cache.retrieval_warnings(pack)
 
 
 HOOK_MESSAGE_ROLE_ALIASES = {
--- a/tools/matrixark_codex_hook.py
+++ b/tools/matrixark_codex_hook.py
@@ -1770,7 +1770,7 @@ def additional_context_from_retrieve(
         return ""
     refs = [ref for ref in _selected_refs_from_retrieve(pack) if not _ref_is_codex_hook_heartbeat(ref)]
     context_text = sanitized_rendered_context_from_retrieve(pack)
-    quality_warnings = pack.get("quality_warnings")
+    quality_warnings = _pack_cache.retrieval_warnings(pack)
     retrieval_metrics = pack.get("retrieval_metrics")
     budget = retrieval_budget_summary_from_retrieve(pack)
     budget_pressure = retrieval_budget_pressure_from_retrieve(pack)
--- a/tools/matrixark_hook_pack_cache.py
+++ b/tools/matrixark_hook_pack_cache.py
@@ -150,6 +150,28 @@ def store_answered(retrieve: object) -> bool:
                ("pack_id", "context_pack_id", "context", "refs", "selected_refs"))
 
 
+def retrieval_warnings(pack: object) -> list:
+    """The retrieval warnings on a pack, under either name it can arrive under.
+
+    `compact_context_pack_for_serving` renames `quality_warnings` to `warnings`, and what reaches a
+    hook is the COMPACT pack -- so a reader that knows only `quality_warnings` reports an empty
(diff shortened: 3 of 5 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. A retrieve that completed was being deleted for running past its 30s deadline and replaced with an empty pack, so agents were served no context at all; the pack is now kept and labelled late, and a warning-field rename that hid it from both hooks is fixed alongside.

PR #1216, opened 2026-09-07
Nine more functions written out twice, now written once

bjmeetsfo:
Nine more functions that existed as two byte-identical copies are now one, and the interesting part
is which rule had been stopping them.

## The owner was being chosen by filename

Several remaining pairs are one live copy and one sitting in a module production cannot reach —
`apply_entity_patch` in `matrixark_mcp_core`, and again in the unreachable
`matrixark_mcp_entity_ops`. Neither copy is imported across modules, so their weights tie, and the
tie broke on filename, which picked the unreachable module as the owner. A separate rule then
correctly refused the whole group for owning a body production cannot reach.

Two rules fighting, and the only visible symptom was "refused", which reads as a real constraint
rather than a bad tie-break. Reachability now leads the ordering, so the live copy owns the body
and the island imports it — adding no edge, since the island already reaches core.

That released a chain across three passes: `best_span_by_edit_distance`, then `apply_entity_patch`,
then `apply_entity_patches`, each unblocked by the one before it.

## Functions that read an environment flag are left alone

`test_flag_readers_agree` holds a floor on how many flags are read from more than one production
site, and checks that the readers of each agree. Collapsing two readers into one drops that count:

```
found 13 flags read from more than one production site, expected at least 14
-- if the read shape changed, this file is looking for something that no longer exists
   and the assertion below passes on an empty set
```

The floor exists so the agreement assertion cannot pass on an empty set, so lowering it to match
would defeat the guard rather than satisfy it. Flag readers are now refused, derived from the body
— `getenv`, `environ`, `env_bool`, `live_int` — rather than from a list of names.

## Verification

By object identity: `owner.f is importer.f` for all 105 consolidations in the tree, every module's
`__all__` still resolving, and every file still parsing.

The whole `tools/` suite ran before and after with the control pinned to this exact commit:
**393 of 393 files on both sides, 357 pass / 36 fail on each, zero tests failing only after.** All
36 remaining failures fail identically on both trees.

### DIFF
--- a/tools/matrixark_mcp_entity_ops.py
+++ b/tools/matrixark_mcp_entity_ops.py
@@ -34,87 +34,19 @@ except ImportError:  # Direct script execution from tools/.
     from matrixark_mcp_core import edit_distance
 
 
-def best_span_by_edit_distance(text: str, search: str) -> tuple[int, int, float]:
-    if not text or not search:
-        return -1, -1, 1.0
-    lower_text = text.lower()
-    lower_search = search.lower()
-    exact = lower_text.find(lower_search)
-    if exact >= 0:
-        return exact, exact + len(search), 0.0
-    search_len = len(search)
-    best = (-1, -1, 1.0)
-    min_len = max(1, int(search_len * 0.7))
-    max_len = min(len(text), max(search_len + 8, int(search_len * 1.3)))
-    step = max(1, search_len // 8)
-    for start in range(0, len(text), step):
-        for span_len in range(min_len, max_len + 1, step):
-            end = min(len(text), start + span_len)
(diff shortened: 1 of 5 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Nine more, and the block turned out to be a bad tie-break rather than a real constraint: several pairs are one live copy plus one in an unreachable island, their weights tie, and the tie broke on filename - picking the island as owner, which another rule then correctly refused. Reachability now leads the ordering. Also: functions that read an environment flag are now left alone, because test_flag_readers_agree holds a floor on how many flags have more than one reader.

PR #1221, opened 2026-09-07
One unreadable record made every read fetch the whole store a row at a time

bjmeetsfo:
`_load_records_by_count` asks `batch_hget` for every record location in one call. One value in the
store cannot be decoded, so that single call raises — and the handler turns the failure into an
empty list, which the next line cannot tell apart from an empty store:

```python
except Exception as exc:
    if is_retryable_temporalstore_error(exc):
        raise
    read_records = []      # a non-retryable failure becomes "no records"
...
if records or count == 0:
    return records         # empty is falsy -> fall through to the per-record loop
```

So every record is refetched one at a time, and nothing says so.

### Measured on a 16,717-record store

| batch size | result |
|---|---|
| 100 / 1000 / 4000 | complete, ≤ 1.0 s |
| 8000 | 5,146 of 8,000 returned, **no error** |
| 16000 / whole store | **raises** `stored value is not UTF-8: invalid utf-8 sequence of 1 bytes from index 0` |

The store holds exactly one such value — record #13526, `...records:000052` field `...214` — and both
`batch_hget` and `hget` raise on it. Because the batch covers everything, that one value costs the
whole read. Profiling one live retrieve:

```
33,434 hget calls  == 2 x 16,717      (the record load runs twice per retrieve)
33,633 socket connects, 89,081 recv_into, 64.5 s in recv_into alone
~167 s per load, against 3.5-10.3 s for the same work in chunks
```

### The change

Read in chunks of 1,000, and degrade **only the chunk that failed** to per-record reads. The other
sixteen chunks keep the fast path, so an unreadable value costs its own chunk instead of the store.
The fallback now also logs, because a silent 20x slowdown is how this ran unnoticed for as long as
one bad value sat there.

A constant rather than a new environment variable: 1,000 is an order of magnitude below the 8,000
where truncation began, and this tree already carries variables with more than one default.

### End to end on a live box

| | before | after |
|---|---|---|
| hook turn | 214,867 ms | **82,527 ms** |
| context served | 7,886 bytes | 7,886 bytes |
| refs | 24 | 24 |

2.6x, and the pack is byte-identical — this changes cost, not content. `context_pack_source` is
unset, so it is a freshly retrieved pack rather than the previous-pack fallback covering for it.

Verified separately on the live store that `batch_hget` returns the same multiset as `hget`, in the
requested order (1000/1000 exact on a dense window), which is what licenses reading in chunks at all.

### Tests

`tools/test_one_bad_record_costs_one_chunk_not_the_store.py`. Against unmodified code two tests fail
on behaviour — `4000 not less than 4000: one bad value still cost a whole-store per-record refetch`
and `the whole store was still requested in a single batch_hget` — while three pass on both arms,
pinning behaviour the change must preserve: readable records still come back in order, a healthy
store never falls back at all, and a **retryable** failure still propagates rather than being
degraded into a slow success.

Suites touching the changed module were run on both arms and are unchanged:
`test_delete_membership_index`, `test_matrixark_context_backfill`, `test_matrixark_rust_serve_readiness`
and `test_mem0_native_read_path` pass; `test_matrixark_the_placement_head_keeps_the_long_form`,
`test_side_index_batch_read` and `test_side_index_merge` carry identical pre-existing failures before
and after.

### Left alone deliberately

The unreadable record itself. Repairing store contents is a separate decision from making a reader
survive them, and the reader should not depend on the data being clean.

### DIFF
--- a/tools/matrixark_temporal_direct_retrieve.py
+++ b/tools/matrixark_temporal_direct_retrieve.py
@@ -33,6 +33,29 @@ except ImportError:
 )
 
 
+#: How many record locations to ask for in one `batch_hget`.
+#:
+#: Measured against the live 16,717-record store: 100/1000/4000 complete in under a second; 8000
+#: came back with 5,146 of 8,000 and no error; the whole store raised. 1000 sits an order of
+#: magnitude under where truncation began.
+#:
+#: Deliberately a constant and not an environment variable. The failure this guards against is a
+#: silent one, and a knob with a second default is its own silent failure.
+BATCH_HGET_CHUNK = 1000
+
+
+def _batch_hget_degraded_log(chunks: int, chunk_size: int) -> None:
+    """Report that a batched read fell back to per-record reads, and roughly what it cost."""
+    try:  # package path
+        from tools.matrixark_mcp_core import _mcp_debug_log
+    except ImportError:  # Direct script execution from tools/.
+        from matrixark_mcp_core import _mcp_debug_log
+    _mcp_debug_log(
+        f"matrixark batch_hget degraded to per-record reads for {chunks} "
+        f"chunk(s) of {chunk_size}: about {chunks * chunk_size} records read one at a time"
+    )
+
+
 class _TemporalDirectRetrieveMixin:
     def retrieve(self, args: Json) -> Json:
         self._ensure_backend_metric_fields()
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. One record in the store cannot be decoded, and because the record load asked for every location in a single batch, that one value made the whole read fail silently and refetch all 16,717 records one at a time; reads are now chunked and only the failing chunk degrades, taking a live turn from 215s to 83s with a byte-identical pack.

PR #1230, opened 2026-09-07
A proxy that came back stopped belonging to the group an operator put it in

bjmeetsfo:
A proxy's group decides which namespace it serves. An operator sets it, and the loop that could set it again is off unless asked for -- `TS_META_PROXY_CALIBRATION` defaults false, deliberately, because reassigning what a proxy serves is intrusive.

Registering again rebuilt the proxy's record and wrote the group away. So a proxy that restarted came back unattached, and nothing put it back.

## Why it did not look broken

The heartbeat still answers. An unattached proxy falls back to the namespace it declared for itself, so it keeps serving something and reports no change -- while the group an operator put it in has quietly stopped being the thing in force.

That is why the test checks what the proxy is told to serve, not just what the record says. The group's namespace and the proxy's own are deliberately different in it, so a proxy serving its own namespace fails even if the field looks right.

## What changed

The group is kept across a re-registration, for the reason the join time beside it is already kept: a proxy saying it is here again is not an operator saying to forget where it belongs.

## Testing

An operator attaches a proxy to a group whose namespace differs from the one the proxy declares. The proxy registers again, and the test asks both what the record says and what the next heartbeat tells the proxy to serve.

Against the tree before this change it fails, saying registering again threw away the group an operator put it in.

A proxy nobody attached still comes back unattached, so keeping a group cannot be confused with inventing one.

## Where this came from

The same shape has now been found five times: a record rebuilt by a struct literal, dropping a field added to it later. Reassigning a shard, registering one, and finishing a load onto one all threw away its pin; this is the proxy's version.

`register_server` is the one that does it right, and shows what the difference looks like: it re-declares the node's hardware and clears its reboot verdict on purpose, and says so in a comment at each. The sites that lost a field had no comment -- nothing was decided there, the field was simply written away.

### DIFF
--- a/crates/temporalstore-rust/src/meta.rs
+++ b/crates/temporalstore-rust/src/meta.rs
@@ -9022,6 +9022,97 @@ fn counting_resources_agrees_with_listing_them_and_counting_those() {
     }
 
     #[test]
+    fn a_proxy_that_comes_back_still_belongs_to_its_group() {
+        let meta = SingleNodeMeta::default();
+        assert!(meta
+            .put_proxy_group(PutProxyGroupRequest {
+                group: "g-1".to_string(),
+                namespace: "served".to_string(),
+                location: String::new(),
+                instance_num: 1,
+                drop_percent: 0,
+            })
+            .status
+            .ok);
+        // The proxy declares a different namespace for itself, so what it ends
+        // up serving shows which of the two is in force.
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. A proxy that came back stopped belonging to the group an operator put it in

PR #1245, opened 2026-09-07
A string delete makes the shard refuse to load, and nothing checked it

bjmeetsfo:
Nothing in the tree both deletes a string and reloads. `StringDelete` appears only in lib tests,
and no integration test covers the pair -- so the ordinary question "does a delete survive a
restart" had no answer.

It does not. **A string delete makes the shard refuse to load:**

    wal_replay_outcome_refused: WAL replay could not install a recorded string outcome at
    sequence 3; refusing load rather than serving a shard missing it
    (address UNRESOLVED, component missing)

`StringDelete` stages through the address-less path --
`stage_meta_outcome(shard_id, "string", &key, .., None, None, true)` -- and the `"string"` arm of
`apply_outcome_item` opens by demanding an address. That is the same defect as mx#1244, and this
makes it five kinds rather than four. It is also the worst of them: `list` and `zset` need a pop
or a rescore, and this needs a string delete.

## The first version of this test passed, and it was wrong

Worth writing down, because the difference is one line and the wrong answer was the comfortable
one.

The first version called `unload_shard` before reopening. That flushes the index, so the reopened
engine read a base that already had the delete and never replayed the tail -- the test passed
without touching the path it is about. `list_recovery` and `zset_recovery` DROP the engine
instead, which is why they reach it.

On the strength of that pass I had already struck the prediction from mx#1244 as measured-wrong.
It was the test that was wrong. The prediction was right.

## Ignored, not red

`#[ignore]`, because a failing test trips the ratchet on every later pull request. The attribute
carries its reason, and removing it is the verification for the fix:

    #[ignore = "fails on main: a StringDelete outcome cannot be installed; see the module comment"]

Run locally against a release build: fails without the attribute, `1 ignored` with it.

### DIFF
--- a/crates/temporalstore-rust/tests/string_delete_recovery.rs
+++ b/crates/temporalstore-rust/tests/string_delete_recovery.rs
@@ -13,14 +13,20 @@
 //! is refused. Nothing exercised it for strings: `StringDelete` appears only in lib tests, and no
 //! test both deletes a string and reloads.
 //!
-//! **It passes.** Reading the arms predicted a fifth failing kind and the prediction was wrong,
-//! which is why this file exists as a test rather than as a sentence in a report. Whatever saves
-//! the string path does not save `list` and `zset`, whose equivalents fail on main today, and the
-//! difference is worth knowing to whoever fixes those: something here already does the right
-//! thing.
+//! **It fails**, and `StringDelete` is a fifth kind:
 //!
-//! Kept because the coverage was genuinely missing, not to prove a point. A string that is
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Superseded by mx#1247, which fixes the defect and carries this test passing rather than ignored.

The finding stands and is worth keeping in the history: nothing checked that a deleted string survives a restart, and it did not -- the load was refused. The correction inside this pull request also stands, because it is the trap: the first version called unload_shard before reopening, which flushes the index so the reload never replays the tail, and it passed while the defect was live.

PR #1246, opened 2026-09-07
Every typed removal refuses the load, hash and set included

bjmeetsfo:
Completes the scope of mx#1244. `hash` and `set` are the two kinds whose typed removal goes
through `mark_bucket_index_page_deleted` and which had **no restart coverage at all** -- so
nothing said whether they shared the defect. They do:

    hash outcome at sequence 3 ... (address UNRESOLVED, component present, 5 chars)
    set  outcome at sequence 3 ... (address UNRESOLVED, component present, 10 chars)

The hash field `alpha`, and the same member hex-encoded for the set.

## So every typed removal refuses the load

| kind | the operation that does it | confirmed by |
|---|---|---|
| `list` | a pop | `list_recovery`, red on main |
| `zset` | a rescore | `zset_recovery`, red on main |
| `string` | a delete | mx#1245 |
| `hash` | deleting a field | here |
| `set` | removing a member | here |

Deleting a hash field, removing a set member, popping a list element, rescoring a zset member and
deleting a string key are all ordinary operations, and each one makes the shard refuse to load
when the tail replays.

## Dropped, not unloaded

Both tests drop the engine rather than calling `unload_shard`, which is the difference between
reaching this path and not. `unload_shard` flushes the index, so the reopened engine reads a base
that already has the removal and never replays the tail -- a test written that way passes without
touching the path it is about. That is exactly how the string case was briefly and wrongly
declared healthy in mx#1245 before the test was corrected.

## Ignored, not red

`#[ignore]` with the reason inline, because a failing test trips the ratchet on every later pull
request. Removing the two attributes is the verification for the fix on the two kinds that have
no other coverage. Run locally against a release build: both fail without the attributes,
`2 ignored` with them.

### DIFF
--- a/crates/temporalstore-rust/tests/hash_set_delete_recovery.rs
+++ b/crates/temporalstore-rust/tests/hash_set_delete_recovery.rs
@@ -1,0 +1,126 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 MatrixArkAI
+
+//! Do a deleted hash field and a deleted set member survive a restart?
+//!
+//! `hash` and `set` are two of the five kinds whose typed removal goes through
+//! `mark_bucket_index_page_deleted`, which stages `address: None, deleted: true`, while their arms
+//! in `apply_outcome_item` open by demanding an address. `list`, `zset` and `string` are confirmed
+//! to refuse the load because of it; these two share the arm shape and had no coverage at all.
+//!
+//! **Both fail**, which completes the set at five:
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Superseded by mx#1247, which fixes the defect and carries both tests passing rather than ignored.

The result stands: hash and set share the defect, which made it five kinds rather than three, and those two had no restart coverage at all before this.

PR #1250, opened 2026-09-07
A portal timestamp says which clock it is

bjmeetsfo:
A timestamp on the portal did not say which clock it came from.

Four places showed one — when the configuration was last written, each entry in its history, the catalog's *updated* column, and a stored memory's time — all through `toLocaleString()`, and **none named the zone**. The records they describe carry the *gateway's* time: a configuration write, an audit entry, a stored memory. So correlating a portal timestamp with a server log was guesswork, and on a browser whose clock is off — which the setup page can now measure and say, since #1181 — it was wrong as well as ambiguous.

The four now go through one helper in the shared nav, which every page has, including the two that are hand-maintained and only get the nav injected.

```
1760000000000  ->  10/9/2025, 1:53:20 AM PDT
```

## Scope, and a count I got wrong

This is **not** a sweep of every `toLocaleString()` in the tree. A first pass counted twenty-four sites; listing them showed most are `Number(...).toLocaleString()` putting separators in a count. There are four timestamps.

## Two things it fixes on the way

- **A missing time read `Invalid Date`.** Two of the four sites had no guard. It is now an em-dash.
- **`0` still means "no timestamp", not 1970.** The catalog site this replaces read `ms ? … : dash`, and my first version rendered the epoch — caught by its own test before it went anywhere.

## Mutations

Thirteen tests, run against the shipped helper with `TZ` pinned so they read the same on any machine. Each of these reddens at least one:

| mutation | reddens |
|---|---|
| the zone is dropped again | `test_a_rendered_time_carries_its_zone` |
| a missing time is the epoch again | `test_nothing_renders_as_a_dash` |
| garbage renders as `Invalid Date` | `test_nothing_ever_reads_invalid_date` |
| the catalog formats its own again | `test_no_timestamp_site_formats_its_own` |
| the config history formats its own again | `test_the_four_sites_go_through_it` |
| the helper is not defined at all | `test_every_page_carries_the_helper` |

`test_the_zone_is_the_browsers_own` is the floor: it renders the same instant under two zones and requires them to differ, so a helper printing a constant would fail rather than pass.

### DIFF
--- a/tools/portal/api_key_portal.html
+++ b/tools/portal/api_key_portal.html
@@ -1096,6 +1096,27 @@
 
      `detail` still wins where there is one: it names the setting and the reason. The scope is
      appended to that rather than replacing it. */
+  /* One way to render an absolute time, and it names the clock.
+   *
+   * Four places showed a timestamp with toLocaleString() and none said which zone it was in, while
+   * the records they describe -- a configuration write, an audit entry, a stored memory -- carry
+   * the GATEWAY's time. Correlating a portal timestamp with a server log was guesswork, and on a
+   * browser whose clock is off (which the setup page can now measure and say) it was wrong as well
+   * as ambiguous.
+   *
+   * `timeZoneName: "short"` rather than the IANA name: "GMT+1" beside the time reads at a glance
+   * where "Europe/London" pushes the useful part off the end of a table cell. Falls back to the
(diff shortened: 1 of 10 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. A portal timestamp says which clock it is

PR #1251, opened 2026-09-07
The durable log is served from memory between writes

bjmeetsfo:
`_read_raw_records` re-read and re-parsed every line of every retained shard, on every call, and
then walked the result again to expand interned metadata. Nothing cached it, while the compacted
view beside it has been cached on the shard signature all along.

Sixteen call sites pay it, including `history`, the raw tombstone count that gates auto-purge, and
the surviving-source-id scan.

| api, 600 messages | before | after |
|---|---|---|
| `history` | 960.7 ms | **5.2 ms** |
| `retrieve` | 204.5 ms | **44.7 ms** |
| `get_memory` | 3.31 ms | 2.06 ms |
| `list_users` | 5.25 ms | 4.38 ms |

`history` is a lookup by memory id; it was reading and parsing the whole durable log to answer, at
761 ms against a store of only 150 messages.

It is now served from memory while the shards are unchanged, keyed on `(total size, newest mtime)`
across the retained shards -- the same validity key `_read_all_compacted` already trusts for itself.
The log is append-only, so a write is the only thing that changes what this returns, and a write
always moves the signature.

## Test

Two cases, both about staleness on a durable path rather than speed:

* a record written after a raw read must appear in the next one;
* a tombstone written after a raw read must be counted -- a stale count feeds the auto-purge
  decision.

Both verified to FAIL with the invalidation removed: "41 not greater than 41: a record written
after a raw read must appear in the next one" and "0 not greater than 0: a tombstone written after
a raw read must be counted".

The suites that exercise the raw log pass: TTL (8), mem0 completion (27), delete/forget (21),
delete-membership index (14), native read path (11), and interned metadata.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -7477,13 +7477,32 @@ class MatrixArkLocalAdapter(_LocalAdapterRetrieveMixin, _LocalAdapterIngestMixin
         raw: list[Json] = []
         if not self._local_jsonl_enabled:
             return raw
+        # Served from memory while the shards are unchanged.
+        #
+        # Every call re-read and re-parsed every line of every retained shard, then expanded the
+        # interned metadata over the result -- two full passes over the durable log to answer, for
+        # instance, one memory's history. Measured at 761 ms against a 150-message store, which is
+        # the whole of that call. The compacted view beside this one has been cached on the shard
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The durable log is served from memory between writes

PR #1263, opened 2026-09-07
One window in the browse harness

bjmeetsfo:
`SyntaxError: Identifier 'win' has already been declared` — every test in
`test_matrixark_a_failed_list_is_not_an_empty_scope` is red on main, from a merge
that had nothing to report.

## What happened

Two changes fixed the same broken tests at the same time, and both gave the browse
harness's sandbox a `window`:

* one lifted the timestamp formatter out of the page — `const win = { __matrixarkWhen: … }`;
* the other, mine in #1261, ran the page's whole shared-helper block — `const win = {}`.

They touched **different lines**, so git had no conflict to report and the merge was
clean. The file that came out declared `win` twice, which is a parse error, so every
test the harness serves failed with a stack trace rather than an assertion.

Neither change was wrong on its own. That is the part worth keeping: a conflict a diff
cannot see needs something that reads the result.

## The fix

The block that runs the whole shared script stays, because it subsumes the other —
`__matrixarkWhen` is defined *in* that block, along with the three helpers beside it,
so running it covers the formatter and them, with one `win` for all of it. The reason
the narrower one was written is folded into the comment on the one that remains.

`sliceBlock` stays too: it is what slices `loadBrowse` itself.

## The guard

`node --check` over every harness in `tools/portal`. It parses without running, costs
a few milliseconds each, and answers before any test that depends on a harness has to.

A second assertion names the mistake rather than leaving the parser to describe it: no
harness declares the same **top-level** name twice. (Column zero only — the same name
inside two different functions is ordinary JavaScript, and a first draft that allowed
any indentation reported those.)

Checked both ways: green on this branch, and against main's harness it fails both
assertions, the second one naming `win`.

### DIFF
--- a/tools/portal/browse_failure_harness.js
+++ b/tools/portal/browse_failure_harness.js
@@ -33,7 +33,9 @@ const loadBrowseSrc = sliceBlock(start);
 /* The page's shared helpers are a script block of their own, emitted before the page's own
  * script. The sandbox below models what the browse region can see, so it needs them too --
  * and it RUNS the shipped block rather than stubbing it, because a stub would let the helper
- * change without a single test noticing. */
+ * change without a single test noticing. That block is where `__matrixarkWhen` is defined, which
+ * is what loadBrowse formats each row's timestamp through -- so running it covers the formatter
+ * and the three helpers beside it, and there is exactly one `win` for all of them. */
 const sharedAt = page.indexOf("Helpers every page may call");
 if (sharedAt < 0) { console.log("FAIL the shared helper block is not on this page"); process.exit(2); }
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
Heads up: main already carries the harness half of this. #1262 removed the same `const win` block ~10 minutes ago and is merged at 548075b8c, so that hunk will conflict on rebase and can be dropped.

The other half is the part main still does not have, and it is the better half: `test_matrixark_every_harness_parses.py` is what would have caught this in the first place. I have deliberately not written my own version of it -- this one should land.

Same root cause we both found, and the fact that two of us fixed it within minutes without either noticing is itself the argument for the test.

PR #1275, opened 2026-09-07
Regenerate the engine flag inventory

bjmeetsfo:
The generated inventory says the engine has 287 flags read by 88 functions. It
has 286, read by 87 -- a flag was removed and the document was not regenerated,
so `test_regenerating_produces_the_same_document` has been failing.

It fails on **main**, which means it fails on every open pull request's ratchet
as well: the one new failure each of them reports is this, not theirs. That is
the cost of a stale generated document -- it does not merely go out of date, it
spends everyone else's gate.

Regenerated with `python3 tools/build_engine_flag_inventory.py .` and nothing
else. Verified stable across two runs, because a generated document that is not
deterministic swaps one permanent failure for an intermittent one.

### DIFF
--- a/docs/ops/temporalstore-engine-flags.md
+++ b/docs/ops/temporalstore-engine-flags.md
@@ -8,7 +8,7 @@ within a week and its staleness is silent.
 
 ## Why this exists
 
-There are 287 of them, read by 88 functions.
+There are 286 of them, read by 87 functions.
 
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Reopening under a branch named for what the change is: a flag was removed, not added, and the branch name survives in the merge commit.

PR #1311, opened 2026-09-07
The one-box retrieval profile is visible to monitoring

bjmeetsfo:
`MATRIXARK_ONEBOX_EMBEDDING_FIRST` decides whether a candidate's score is its vector similarity alone or a 0.72/0.28 blend with a lexical match. It is **on by default**, it decides every result the deployment returns, and **nothing published it** — so "why did the answers change?" had no answer in any series the gateway emits, and two deployments answering differently looked identical on every dashboard.

Three gauges, because they answer different questions: the profile says how a score is computed, `return_all_candidates` says whether ranking may drop anything, and the threshold says at what size that stops applying. A panel showing the first without the others explains half of a changed answer. All three are emitted **at their defaults**, because a gauge that appears only once somebody changes something cannot be alerted on before they do.

## The first version of this gauge lied

It read the profile by importing the retrieval module — which is a **circular import**, the adapter imports the retrieval mixin and the mixin imports the adapter — so the import raised, an `except` caught it, and the gauge published `0`.

A gauge saying "blended scoring" about a deployment running the profile **on** is worse than no gauge: every dashboard reading it describes the opposite of what happened.

It reads the environment now. The default it falls back to is a second copy of the retrieval module's own, and is asserted against it — two copies of a default agree until one is edited.

Seven tests, including the case almost every deployment is in (nothing set, profile on) which is exactly the case the first version got backwards, and the positive control that the series reaches the scrape at all.

### DIFF
--- a/docs/ops/matrixark-gateway-dashboard.json
+++ b/docs/ops/matrixark-gateway-dashboard.json
@@ -738,6 +738,46 @@
           "legendFormat": "{{reason}}"
         }
       ]
+    },
+    {
+      "id": 20,
+      "type": "timeseries",
+      "title": "One-box retrieval profile",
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
Rebased onto main now that #1310 has landed.

The two failures were a merge-order artifact, not a defect: this branch's `onebox_lines()` imports `return_all_candidates_enabled` and `return_all_candidate_threshold`, so the wiring guards counted it as a caller — while the branch was still based on a main where those knobs were listed as unwired and read-by-nothing. #1310 struck them off; both guards pass on the rebase.

PR #1312, opened 2026-09-07
Nine flag readers spell the vocabulary; now they call the one that owns it

bjmeetsfo:
#1308 gave the crate one reader for boolean flags and a guard that no reader may decide by
case. Nine accessors were already correct — they just each restated the vocabulary instead of
calling it:

```rust
pub(crate) fn trim_enabled() -> bool {
    !matches!(
        std::env::var("TS_MALLOC_TRIM")
            .unwrap_or_default()
            .trim()
            .to_ascii_lowercase()
            .as_str(),
        "0" | "false" | "no" | "off"
    )
}
```

becomes

```rust
pub(crate) fn trim_enabled() -> bool {
    crate::env_flag::env_bool("TS_MALLOC_TRIM", true)
}
```

Nine bodies of nine lines become nine calls: `TS_MALLOC_TRIM`, `TS_WAL_BINARY_FRAME`,
`TS_WAL_COMPRESS_RECORDS`, `TS_WAL_OUTCOME_ITEMS`, `TS_VECTOR_INT8`, `TS_VECTOR_SCALED`,
`TS_NODE_SUMMARY_VECTOR`, `MATRIXARK_BULK_INGEST` and `MATRIXARK_EAGER_CACHE_WARM_ON_LOAD`.

## Why this is the same behaviour, not merely similar

Both inline shapes are `env_bool` with a fixed default, for every input:

| written | `!matches!(…, "0"\|"false"\|"no"\|"off")` | `env_bool(n, true)` |
|---|---|---|
| unset | `unwrap_or_default()` gives `""`, in neither list → `true` | default → `true` |
| `off` | in the list → `false` | `Some(false)` → `false` |
| `yes` | not in the list → `true` | `Some(true)` → `true` |
| `wat` | not in the list → `true` | unreadable → default → `true` |

and the mirror for the `"1" | "true" | "yes" | "on"` shape with `env_bool(n, false)`, where an
unset variable gives `""`, matches nothing, and comes back `false`. Trim and lowercase were
already being applied in both, by the same two calls the shared reader makes.

That equivalence is what makes this a rewrite rather than a change: the one case where an inline
deny-list and `env_bool` could differ is an unreadable value, and for these the flag's default
*is* the value the deny-list produced.

## Checks

The tests that select **both** positions of these flags pass unchanged — which is the point,
since a semantics change would land there first:

- `memory_trim::the_trim_switch_is_read_every_time`, which drives `"0"`, `"false"`, `"NO"`,
  `"Off"`, `" 0 "` and `"1"` — the case and whitespace spellings.
- `wal_proto::compression_is_on_when_nothing_says_otherwise`, and its sibling that drives every
  off spelling including `"OFF"`.
- The `log_framing` frame tests and the vector-encoding tests for the three `types.rs` knobs.

20 targeted tests pass. `cargo check --release --lib --bins --tests` clean. Full
`cargo test --release --lib --tests --no-fail-fast -- --test-threads=1`: no failure that main
does not already have.

Net −63 lines, and `crate::env_flag` is now the only place in the crate that spells the words.

### DIFF
--- a/crates/temporalstore-rust/src/engine.rs
+++ b/crates/temporalstore-rust/src/engine.rs
@@ -1873,14 +1873,7 @@ pub(crate) fn timestamp_range_bounds(
 /// themselves and gate their own half of the same decision; a copy that drifted would leave one
 /// subsystem in bulk mode and the rest on the live path.
 pub(crate) fn bulk_ingest_mode() -> bool {
-    matches!(
-        std::env::var("MATRIXARK_BULK_INGEST")
-            .unwrap_or_default()
-            .trim()
-            .to_ascii_lowercase()
-            .as_str(),
-        "1" | "true" | "yes" | "on"
-    )
+    crate::env_flag::env_bool("MATRIXARK_BULK_INGEST", false)
 }
 
 /// Whether the per-command model-map -> bucket-index promotion and first-index
@@ -1900,14 +1893,7 @@ fn defer_bucket_index_reconstruct() -> bool {
 /// store after reconstructing the index (disk->memory promotion on restart).
 /// Defaults ON; set MATRIXARK_EAGER_CACHE_WARM_ON_LOAD to 0/false/off/no to disable.
 pub(crate) fn eager_cache_warm_on_load() -> bool {
-    !matches!(
-        std::env::var("MATRIXARK_EAGER_CACHE_WARM_ON_LOAD")
-            .unwrap_or_default()
-            .trim()
-            .to_ascii_lowercase()
-            .as_str(),
-        "0" | "false" | "no" | "off"
-    )
+    crate::env_flag::env_bool("MATRIXARK_EAGER_CACHE_WARM_ON_LOAD", true)
 }
 
 /// Current on-disk shape of a shard index.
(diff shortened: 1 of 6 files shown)


bjmeetsfo:
Withdrawing this. The nine readers it rewrites were not duplication for its own sake: the literal `env::var("TS_X")` in a reader body is the anchor a family of derivations keys on, and replacing it with `env_flag::env_bool("TS_X", true)` takes that anchor away.

Measured rather than assumed. Three separate derivations degrade:

* `test_matrixark_the_wal_compresses_a_record_by_default` reads the body to decide which way the flag defaults.
* `engine_defaults()` in `test_matrixark_engine_settings_match_the_engine` scans for `env::var("TS_...")` to derive the default the portal must show.
* A second scan in the same file derives which knobs the engine reads at all -- and TS_WAL_COMPRESS_RECORDS, TS_VECTOR_SCALED and TS_MALLOC_TRIM started reporting as offered-but-never-read.

I taught the first two and the third surfaced, which is the signal: eight files under tools/ key on that anchor, and build_engine_flag_inventory.py already documents having been extended once to cope with helper-mediated reads. The benefit was 63 lines of cosmetics; the cost is the machine-readability of defaults that this repo deliberately relies on.

What the change was protecting against is already fixed and guarded by #1308: crate::env_flag owns the vocabulary, and test_no_flag_reader_decides_by_case stops any reader deciding by case. The nine readers here are the canonical shape already, so nothing about them is a defect.

PR #1318, opened 2026-09-08
A 32-wide vector did not come from a 3072-wide encoder

bjmeetsfo:
Measured on the live store: **9,098 vectors carrying `text-embedding-3-large` at 32 dimensions.**

That model is 3072 wide. 32 is the width of the deterministic fallback, and every encoder in the catalogue is 384 or more — so those vectors were produced by the fallback and recorded under whatever name was configured. There is exactly one such pair in that store, so every embedded record is in this state, and the count is still growing.

## Why it was invisible

`embedding_status` computed each record's model and its width **in the same pass**, then put them in two separate tallies — one of names, one of widths. The store could say *"these models appear"* and *"these widths appear"*, and never *"this model wrote this width"*, which is the only form in which the contradiction exists.

Pairing them is three lines. The check falls out of it.

## Why it is not cosmetic

Two vector spaces sit in one store under one name, so nothing downstream can separate them. The **model hash cannot** — both were hashed from the same configured string — and a backfill that trusted the name would skip exactly the records that need redoing.

## It reports; it does not repair

Rewriting the labels would make the store self-consistent and still wrong. Recording the encoder that actually ran is a change to the **write path**, with its own migration, and belongs in its own change.

The one-box page shows the finding **above** the table rather than below it: a reader who stops at the first thing they see should stop at the one saying the names cannot be right.

## Checks

Five mutations, each reddening a distinct assertion — including removing the pairing in the adapter, which leaves the check with nothing to read and is the floor the rest stands on.

The threshold is the encoder's own `EMBEDDING_DIM`, not the literal 32, so a fallback that changes width cannot silently turn the check off. Verified against three shapes: the live pairing (1 finding), the fallback naming itself honestly (0), and a real encoder at its real width (0).

### DIFF
--- a/tools/matrixark_local_adapter_dashboard.py
+++ b/tools/matrixark_local_adapter_dashboard.py
@@ -728,6 +728,10 @@ class _LocalAdapterDashboardMixin:
         without_vector = 0
         models: dict[str, int] = {}
         dimensions: dict[int, int] = {}
+        # The pairing the other two throw away. Kept because the interesting question is not which
+        # names appear or which widths appear, but whether a name appears at a width it could not
+        # have written.
+        model_dimensions: dict[tuple, int] = {}
         oldest_pending_ms = 0
         newest_pending_ms = 0
         deferred_tasks = 0
@@ -790,6 +794,9 @@ class _LocalAdapterDashboardMixin:
                 dim = 0
             if dim:
                 dimensions[dim] = dimensions.get(dim, 0) + 1
+            if model and dim:
+                pair = (model, dim)
+                model_dimensions[pair] = model_dimensions.get(pair, 0) + 1
 
         return {
             "status": "ok",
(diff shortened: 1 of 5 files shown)


bjmeetsfo:
Note on the repair this deliberately leaves out, since the obvious next step does not work.

The truth looks like it is already available: `embedding_execution_mode_name()` returns `local_hash_embedding_fallback` when the fallback ran, while `embedding_model_name()` returns the configured name — and the write path records the second. So recording the first instead looks like a two-line fix.

It is not. `embedding_fallback_used()` reads `_EMBEDDING_FALLBACK_USED`, a module global set in three places and **never reset**. It is sticky for the life of the process, so it means "a fallback happened at some point", not "this vector used one". Labelling on it would mark every vector written after the first fallback — including ones the API produced correctly once it recovered — and would replace a wrong label with a differently wrong one.

A correct repair needs a **per-call** provenance signal: the encoder returning what produced each vector, rather than a process flag consulted afterwards. That is a change to the hottest write path and belongs in its own PR with its own measurement.

Which is why this check keys on the **width**. The width is per-record, already stored, and cannot be made sticky by a process global — it is the one signal that is true about the individual vector.

PR #1323, opened 2026-09-08
The flag inventory names the slab-reverify flag

bjmeetsfo:
`test_matrixark_engine_flag_inventory` is red on main, and has been since #1315 — which added `TS_REVERIFY_ALL_SLABS` to the engine without regenerating the inventory that lists it.

```
test_every_flag_named_in_the_engine_is_listed   [] != ['TS_REVERIFY_ALL_SLABS']
test_regenerating_produces_the_same_document    287 flags, 88 functions  !=  286, 87
```

Regenerated with the command the failure message itself names: `python3 tools/build_engine_flag_inventory.py .`

Found while diagnosing two unrelated PRs whose ratchets failed with exactly this pair. Neither touched the engine or any generated document, which is what made it worth checking main before touching either branch — the failures were shared with main, not caused by them.

### DIFF
--- a/docs/ops/temporalstore-engine-flags.md
+++ b/docs/ops/temporalstore-engine-flags.md
@@ -8,7 +8,7 @@ within a week and its staleness is silent.
 
 ## Why this exists
 
-There are 286 of them, read by 87 functions.
+There are 287 of them, read by 88 functions.
 
 Deleting unreachable code is not the lever. An earlier version of this document argued that
 by asserting every accessor had a caller -- true when it was hand-checked at 55, and carried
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Superseded by #1321, which landed while this was queued and did the job more completely: it regenerated the inventory AND recorded the decision the new flag needs. This one only regenerated, which is why it now fails `test_every_unselected_boolean_has_a_decision` — the very check #1321 satisfied.

Avoidable. A red test on main is a race, and the rule is to check OPEN PRs before fixing one. I checked main and not the queue.

PR #1329, opened 2026-09-08
Did this call fall back, not has anything ever

bjmeetsfo:
Every retrieve reports `embedding_fallback_used` and `embedding_execution_mode` in its context pack. Both read one module global, set `True` in three places and reset **nowhere** — the only assignment to `False` was the import-time initialiser.

So one fallback, anywhere, at any point in the process, made every later retrieve report that it had fallen back. Including the ones a recovered encoder served correctly. The pack states it as a fact about *that* retrieve, and after the first fallback it was never true again.

Measured on main, with the two probes this suite uses:

| | main | this change |
|---|---|---|
| after a call that fell back | `True` | `True` |
| after a call that did **not** | `True` — and for ever | `False` |
| execution mode | `local_hash_embedding_fallback` | `deterministic-token-hash` |

## Per call, and per thread

The gateway serves retrieves through `asyncio.to_thread`, so two requests can be inside the encoder at once. A single global reset at entry would let one request's reset erase the other's answer — an intermittent wrong answer, which is harder to notice than the permanent one being fixed. It is a thread-local.

## Checks

Five mutations, each reddening a distinct assertion, including *"it never resets"* — the exact behaviour this replaces.

The last one took three attempts, and both failures were worth having:

* **A batch delegates.** `embeddings_for_texts` calls `embedding_for_text` for anything it has to encode, and that resets — so removing the batch's own reset changes nothing on any path that encodes something. Only a batch served *entirely from cache* reaches the batch entry and nothing beneath it.
* **The cache is keyed on `(model, text)`.** A first attempt at that test switched provider between the two calls, which switches the model name and misses every cache entry — so it delegated after all. The working version uses the same provider both times.

The thread test is driven through the flag rather than the encoder: making each thread encode would mean each setting `MATRIXARK_EMBEDDING_PROVIDER`, and the environment is process-wide, so the two would stomp each other and the test would measure that race instead of thread isolation.

### DIFF
--- a/tools/matrixark_mcp_embeddings.py
+++ b/tools/matrixark_mcp_embeddings.py
@@ -9,6 +9,7 @@ import collections
 import hashlib
 import math
 import os
+import threading as _threading
 import threading
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
The four ratchet failures here are **not from this branch**. All four are the engine flag inventory, which is red on main again — a flag's category moved and the generated document was not regenerated with it.

Checked both ways before touching anything: no open PR was already fixing it, and the two suites fail identically on `c142c71ee` with nothing of mine applied. #1331 regenerates it.

This branch changes one module (`matrixark_mcp_embeddings`) and adds one test file; it touches no engine flag and no generated document. I will rebase onto main once #1331 lands so the gate runs against a green base.

PR #1331, opened 2026-09-08
The flag inventory catches up with the engine again

bjmeetsfo:
`test_matrixark_engine_flag_inventory` is red on main again — a flag's category moved (69 → 70 in one bucket) and the generated inventory was not regenerated with it.

Regenerated with the command the failure message names: `python3 tools/build_engine_flag_inventory.py .`

## This is the second time today

The same gate was red this morning for the same reason, fixed by #1321. It is red again a few hours later.

The gate itself works — `test_regenerating_produces_the_same_document` catches this precisely. What it cannot do is stop the engine change from landing without it, because the check runs on the PR that introduces the flag and the merge happens anyway. Every subsequent PR then inherits a red ratchet and has to prove its own innocence before it can merge.

Worth someone deciding whether the inventory should be regenerated by CI rather than by hand. I have not done that here: it changes how a generated document enters the tree, which is a policy call rather than a fix, and this PR is the unblock.

Found while diagnosing #1329, whose four ratchet failures were all this and none of them its own.

### DIFF
--- a/docs/ops/temporalstore-engine-flags.md
+++ b/docs/ops/temporalstore-engine-flags.md
@@ -62,7 +62,7 @@ Anything else is blank, and a blank means go and look.
 |---|---|
 | total | 287 |
 | booleans whose default this could read off the source | 47 |
-| numbers whose default this could read off the source | 70 |
+| numbers whose default this could read off the source | 69 |
 | **defaulting on, and set by nothing** | 5 |
 | offered on the portal | 26 |
 | **that nothing in this repository sets** | 158 |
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
Extended: the same rename left **three** checks behind, not one.

`TS_STORAGE_ZONE_SIZE` keeps the old word while the field it fills took the new one (`storage_band_size`), so the flag and its default no longer resemble each other — and all three checks joined them by resemblance.

1. **The inventory** — regenerated (the original commit here).
2. **The tuning validator** — asked the engine for `DEFAULT_STORAGE_ZONE_SIZE`, which the rename removed. Its own test names that shape: *"a constant that no longer exists reads as no default"*. One line.
3. **The settings-match-the-engine derivation** — finds a default by looking for `env::var("TS_X")` with an `unwrap_or` beside it. This knob is now read as `parse_u64(get(TS_STORAGE_ZONE_SIZE), defaults.storage_band_size)`, with the value three hops away through the defaults struct. It follows that now, which covers **seven** knobs in that shape rather than only the one that was failing.

The number the portal shows was correct throughout — 1073741824, which is `1 << 30`. What the rename cost was not a wrong value but everything that would notice if it became one.

PR #1338, opened 2026-09-08
The one-box write path, its background passes, and the reads that serve a request

bjmeetsfo:
Ten changes to the one-box write path, its background passes, and the reads that
serve a request. Every claim below is a measurement on an AWS one-box against a
production corpus, with e5-large embeddings and glm-4-flash extraction. Where a
change is not yet measured, it says so.

## The write path

**A write re-indexed every field of the object it touched.** A single-component
write re-synced the whole object's page list. Datanode CPU over the same run:
40.9 s -> 0.13 s.

**A deployment that ranks by vector alone can stop storing index postings.**
Opt-out for the inverted index on a one-box that retrieves by embedding.

**The index log can carry compressed payloads.** A new codec on the existing
container. Measured on an identical 60-message corpus, both arms on their own
fresh store:

| | index log | pages | durable | add p50 | frames |
|---|---|---|---|---|---|
| off | 1,620,116 | 1,313,686 | 4,024,597 | 0.065 s | 555 msgpack / 0 zstd |
| on | 430,910 | 1,205,839 | 2,727,615 | 0.064 s | 152 / **403 zstd** |

Index log **3.76x** smaller, total durable **1.48x**, add latency unchanged. In
the 2 h production run, 87.6% of frames were zstd.

## The background passes

**The background summary pass remembers the count it left, not the one it
found.** The skip compared the record count against the count taken BEFORE the
pass -- and a pass writes, so the next tick always saw a different count and the
skip could never fire again. It did a full read_all and wrote on every tick of
its 1 s timer, forever, on a store nobody was touching. 240 s with zero client
traffic:

| | idle bytes | rate | add p50 |
|---|---|---|---|
| before | 2,969,295 -> 9,377,677 | **26,701 B/s** (~2.3 GB/day) | 0.064 s |
| after | 1,992,908 -> 1,992,908 | **0 B/s** | 0.064 s |

Controlled separately for "does it still do the work": four facts written, then
polled for ten minutes. Both builds recall the fact at t=30 s; the fixed one
writes 55,663 bytes and stops, the old one writes 3,589,516 and is still
climbing at t=600 s.

**The refresh skip also asks whether a purge moved the store under it.** A purge
rewrites records in place without moving the count, so a count-only skip would
park forever on exactly the node that needs rebuilding.

## The proxy

**The remote path reads every spelling of "there is no metaserver".** It decided
with is_empty(), so only "" counted -- but the rule has one implementation
already (META_ADDR_SENTINELS / single_node), and a one-box sets
TS_META_ADDR=local. "local" reached the client as a literal socket address and
**every write failed**. Isolated: metaserver "local" -> ingest 500, "" -> 202.

**A leak this does NOT fix, documented rather than papered over.** Four caches
live for the life of the process with no cap. Same 312 MB store, two processes:
a fresh one answers a retrieve in 22.6 ms holding 13.1 MB; the aged one took
4,639 ms holding 4.6 GB, and is OOM-killed repeatedly (the gateway respawns it,
so only dmesg shows it). A control arm with shedding disabled reproduces it:
proxy RSS 174 -> 1,419 MB in 22 minutes, two more OOM kills.

An entry-count cap was written, measured against its own control, and REVERTED
in this branch because it does not work -- 970 MB vs 1,433 MB at t=750s, no
reduction. The entries are not what is large: one entry is a whole hash
snapshot, a whole scan result, or a candidate snapshot. Bounding this needs a
bound on bytes, or a cap small enough to bite. The commit and its revert are
both here so the measurement is not lost.

## The read path

**Authenticating stops costing the whole record log.** Seven lookups in the
access layer are newest-wins scans for a single record type, and each read the
entire log. find_active_api_key runs on every authenticated request, and
ensure_account_tenant_active and ensure_user_active add more, so an
authenticated call did three to five full-corpus reads before any work. Each
asks for its own type now, through the type it already names inline -- so no
central list of type names exists to fall out of date, which matters because the
metadata boundary is a prefix.

**An SSO login looks up its mapping, not the whole log.** The same shape on the
live login path.

## Gates

- `cargo test -p temporalstore-rust --lib -- --test-threads=1` on this branch
- `tools/test_an_authenticated_request_does_not_read_the_whole_log.py`: 4 tests,
  including a positive control that the narrow path is taken (read_all_calls ==
  0) and that an authoritative empty does not fall back. Disabling the fast path
  fails two of them, so they discriminate.
- `tools/test_matrixark_summary_worker.py`: 6 passed
- `test_matrixark_access_governance` has one pre-existing failure, verified
  identical on the unmodified file.

## Not claimed

The proxy's unbounded caches are diagnosed, not fixed -- see above. The
`orjson` lane-parser change is deliberately not in this branch: 68% of gateway
CPU is decoding lane JSON, but that change has no measured benefit yet and does
nothing where orjson is not installed.

### DIFF
--- a/crates/temporalstore-rust/src/context_workflow/embed_drainer.rs
+++ b/crates/temporalstore-rust/src/context_workflow/embed_drainer.rs
@@ -323,9 +323,19 @@ pub fn run_embed_drainer_loop<F: Fn() -> bool>(
 ) {
     while !should_stop() {
         let report = drain_embedding_dirty_once(engine, config);
-        // If the pass did real work and may have more queued, loop immediately;
-        // only back off when the pending set was empty this pass.
-        if report.pending_scanned == 0 {
+        // Back off unless the pass made PROGRESS -- not merely unless it found work.
+        //
+        // The old condition read "we scanned something" as "we did real work, so come straight
+        // back". Those differ whenever a node stays dirty: an unreachable or misconfigured
+        // encoder fails the batch, `failed` rises, the markers are deliberately left for a later
+        // pass, and the next scan returns the SAME nodes. The condition is then true forever and
+        // the loop never sleeps -- a hot spin that re-queries and rewrites the same records as
+        // fast as the CPU allows, for as long as the encoder is unavailable.
+        //
+        // `cleared` is the progress signal: markers actually retired this pass, whether by
+        // embedding them or by clearing text-less ones. A pass that cleared some and left more
+        // queued still comes straight back, which is what the original condition wanted.
+        if report.cleared == 0 {
             std::thread::sleep(config.interval);
         }
     }
--- a/crates/temporalstore-rust/src/engine/storage_bucket_internals.rs
+++ b/crates/temporalstore-rust/src/engine/storage_bucket_internals.rs
@@ -1108,6 +1108,67 @@ pub mod uncovered_maintenance {
     }
 }
 
+/// Sync only the components a command actually wrote.
+///
+/// [`sync_context_pages_for_object`] is given an object key and nothing else, so it re-upserts
+/// EVERY field that object holds. For a record hash carrying one field per record that is one
+/// upsert per stored record on every write, and each upsert removes and reinserts an entry in the
+/// object's component vector -- two `Vec` shifts whose tails are the whole object. The cost is
+/// quadratic in the object's field count.
+///
+/// Measured on a production-corpus one-box at a 260 MB store: a single-message ingest cost 40.1 s
+/// of datanode CPU out of a 52 s wall, and a DWARF-unwound profile put 60.9% of engine CPU in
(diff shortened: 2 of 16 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Ten changes to the one-box write path, its background passes and the request-serving reads: the summary refresher's skip could never fire (idle writes 26,701 B/s -> 0 with recall unchanged), the proxy's remote path read only one of five "no metaserver" spellings so a one-box failed every write (500 -> 202), index-log compression is 3.76x at flat add latency, and authenticating no longer costs a full-corpus read per request. Gate: cargo test -p temporalstore-rust --lib = 1,720 passed, 2 failed, both reproducing identically on unmodified main. One change (an entry-count cap on the proxy's unbounded caches) is committed AND reverted in the same branch because its own control showed no reduction - the leak it targets is real and still open.

PR #1364, opened 2026-09-09
The gateway talks to the proxy over HTTP, and the proxy gives its heap back

bjmeetsfo:
The proxy was the CPU bottleneck of the one-box and the reason it ran out of memory: measured over a production-corpus soak it burned 46.3% of a core to the gateway's 20.5%, and its resident set climbed monotonically to 2,941 MB while the datanode holding the actual store sat at 385 MB.

Four changes, each measured separately on AWS against the arm before it, on a fresh store, with zero request failures throughout.

## The gateway talks to the proxy directly, over HTTP

`--serve` reads one request per line from stdin, so something has to own that pipe. In the one-box that owner was a Python daemon: it accepted a unix socket, parsed each request, re-serialised it onto the proxy's stdin, read the answer back and re-serialised that too — four full JSON operations per request, on the largest payloads in the system, to move bytes between two file descriptors. It cost **21.9% of a core and 72 MB of RSS**, and it added a process to a box that was already OOM-killing. It also held one proxy behind one lock, so every caller serialised behind every other.

The proxy now serves HTTP itself, through the same `http.rs` server the datanode listens on, so the whole path is one pattern:

    gateway --HTTP--> proxy --HTTP--> datanode

The wire format is deliberately unchanged — the body is the same `RecordLogRequest` the pipe carried and the reply the same `RecordLogResponse` — so a client changes only *where* it sends, never what it sends. That is what makes the transports swappable under a running gateway and comparable in an A/B.

Requests are still serialised by default, because nothing in this engine has ever had two in flight: the pipe handled one at a time and the daemon's lock enforced it. Removing the middleman is a transport change and is worth having on its own; letting requests overlap is a concurrency change, so it sits behind `MATRIXARK_RUST_PROXY_HTTP_CONCURRENT` where it can be measured, and reverted, separately.

The gateway side reads `MATRIXARK_RUST_PROXY_HTTP` and keeps a per-thread keep-alive connection — per-thread because `HTTPConnection` is not safe to share, keep-alive because a TCP handshake per request would hand back a slice of exactly the overhead this removes. Its deadline formula is copied from `_call_socket_json` rather than reinvented, so the two transports remain comparable.

## The proxy hands freed pages back to the OS

`free` returns memory to the allocator, not to the kernel, so a process that encodes and decodes whole record batches keeps every peak it has ever reached.

This is the whole story of the proxy's resident set, and the diagnosis turned on one observation: **in the one-box the proxy runs in remote mode and holds no engine at all.** It reports root `proxy://127.0.0.1:17002/...` and forwards every operation to the datanode. So its 2,941 MB was never engine state, and never live data — it was freed chunks the allocator had not given back.

A background thread now trims once a byte ledger passes a threshold. It is off the serving path deliberately: a trim walks the free lists, so a request pays one relaxed atomic add and nothing else. Threshold is 8 MiB via `MATRIXARK_RUST_PROXY_TRIM_BYTES` (0 disables), and `TS_MALLOC_TRIM=0` disables the trim itself. Bytes below the threshold stay on the ledger rather than being discarded, because a trickle of small requests retains as much as one large one — it only takes longer to get there.

## Index maintenance stops building `Value` trees

The write path parsed every appended record into a `serde_json::Value` to read two things from it: the record type, and a scope key that may sit at any of five places. `IndexFacts` borrows exactly those out of the caller's buffer and lets serde skip the rest of the record without materialising it. A payload whose shape `IndexFacts` does not model falls back to the `Value` decode, so an unmodelled record costs a parse rather than being misfiled.

The snapshot cache also held `BTreeMap`s by value, which meant a deep copy of the whole snapshot twice — once to store it and once on every hit. For a record shard the values *are* the payloads, so `fetch_indexed_payload_values` copied entire shards to read the handful of fields its index named. It now holds `Arc`s, and the write side patches them through `Arc::make_mut`, which copies only while a reader holds one.

## The snapshot cache is bounded by bytes

It had no bound and no eviction of any kind. The budget is in **bytes**, deliberately: a cap on the *number* of entries says nothing about memory when an entry is a whole shard of variable-size records, and an entry-count cap tried on this proxy earlier moved the resident set by nothing at all. Default is RAM/16 held between 64 MiB and 512 MiB, via `MATRIXARK_PROXY_SNAPSHOT_CACHE_BYTES`.

**Reported honestly: this one measured inert.** Set to 384 MiB and verified applied by reading it back from `/proc/<pid>/environ`, resident memory followed the same curve (2,987 MB at 198 MB durable, against 2,202 MB at 156 MB unbounded). Removing the double *copy* was worth it; bounding the cache *contents* was not, because the contents were never the bytes. It is kept because an unbounded structure with no eviction is still worth bounding, but it earns no part of the numbers below.

## Measured

AWS one-box, c7i.xlarge, production corpus, e5-large encoder, fresh store per arm, 1,500 s, zero request failures. Compared at equal **message count**, not equal elapsed time — the faster arms do more work in the same wall clock and would otherwise be penalised for holding a larger store.

| | stock | this branch |
|---|---|---|
| add p50 | 1,862 ms | **1,262 ms (−32%)** |
| throughput | 1,005 msgs | **1,180 msgs (+17%)** |
| proxy CPU / message | ≥93.1 ticks | **72.3 ticks (≥−22%)** |
| proxy RSS | climbs to 2,987 MB | **sawtooth, 238–1,437 MB (−52% to −89%)** |
| processes | 2 | **1** |

The CPU figure is a floor, not a point estimate: the stock arm ran under the daemon, which respawns the proxy, so its `/proc` ticks covered only the last incarnation. The arms here have no daemon and never restart, so their counters are complete.

Retrieve p50 is **not** improved (2,348 → 2,579 ms). The likeliest cause is untouched by this branch: `type_index_payloads` collects every location of the requested types into a `BTreeSet` before applying a per-type cap, and the serving path passes `newest_by_type: None`, so no cap applies at all — an O(corpus) walk per request regardless of how many records the caller asked for.

## Tests

- The borrowed parse is checked against the `Value` path's own answer over 17 payloads: all five scope-key sources, their precedence, an empty earlier source falling through, bundles, and four shapes that must fall back. Reordering the sources turns it red.
- A companion test asserts an ordinary record does **not** fall back — without it the agreement test would pass vacuously, since the fallback arm trivially agrees with itself.
- The cache bound has seven tests, including a positive control that it is not simply evicting everything, which is how a bound can satisfy a bound-test while destroying what it bounds.
- Seven transport tests run the identical request down **both** transports and compare, so a silent divergence fails rather than passes. They use a namespace unique **per run**: the store root is derived from namespace/table, so a fixed one resolves to the same `/tmp` directory every time, and a store left behind by a build of another lineage makes the suite report a transport failure that is really a stale store — observed exactly that way. Cases: health, write-then-read, an application failure staying a 200 with `ok:false`, malformed JSON reported not dropped, `client_request_id` echoed, keep-alive across many requests on one connection, and eight concurrent callers each getting their own answer.

## Not included

The daemon's own relay was rewritten to forward bytes verbatim and is **left out of this PR**. It measured as a regression — an unbuffered binary pipe makes `readline()` read one byte per syscall, which took the daemon from 22% of a core to 86% and cut throughput 64%. The bug is understood and fixed locally, but it has not been re-measured since, and HTTP supersedes that path.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -456,6 +456,12 @@ fn default_true() -> bool {
 
 fn main() {
     let args: Vec<String> = env::args().collect();
+    // `--serve-http <addr>`, or the env var, so a deployment can switch transport without
+    // changing its command line. Checked BEFORE `--serve`, because "--serve-http" would
+    // otherwise never be reached if anything ever passes both.
+    if let Some(addr) = http_serve_addr(&args) {
+        std::process::exit(serve_http(&addr));
+    }
     if args.iter().any(|arg| arg == "--serve") {
         std::process::exit(serve());
     }
@@ -473,6 +479,21 @@ fn main() {
     if !response.ok {
         std::process::exit(1);
     }
+}
+
+/// The HTTP listen address, from `--serve-http <addr>` or `MATRIXARK_RUST_PROXY_HTTP_ADDR`.
+fn http_serve_addr(args: &[String]) -> Option<String> {
+    if let Some(index) = args.iter().position(|arg| arg == "--serve-http") {
+        if let Some(addr) = args.get(index + 1) {
+            if !addr.trim().is_empty() && !addr.starts_with("--") {
+                return Some(addr.trim().to_string());
+            }
+        }
+    }
+    std::env::var("MATRIXARK_RUST_PROXY_HTTP_ADDR")
+        .ok()
+        .map(|addr| addr.trim().to_string())
+        .filter(|addr| !addr.is_empty())
 }
 
 fn single_shot_debug_enabled(args: &[String]) -> bool {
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The proxy now serves HTTP itself so the gateway talks to it directly (no Python daemon: -1 process, -21.9% of a core, -72 MB) and hands freed pages back to the OS, which took its resident set from a climb to 2,941 MB down to a 238-1,437 MB sawtooth; add p50 -32%, throughput +17%, proxy CPU/message >=-22%, measured on AWS at equal message count. Retrieve latency is explicitly NOT improved and the cache byte-bound is reported as measured-inert - both are called out in the PR body rather than claimed.

PR #1366, opened 2026-09-09
Profile-driven: the lane can be binary, the scan can return less, and the engine can rank

bjmeetsfo:
Follow-up to #1364, from profiling the running one-box rather than from reading it.

Two profiles under load said where the time goes:

    gateway (py-spy, 4,628 samples)   53.5%  raw_decode -- the stdlib JSON decoder
    proxy   (perf, 60s)                8.4%  malloc
                                       3.1%  BTreeMap<String, Value>::insert
                                       2.9%  BTreeMap<String, Value>::insert_entry
                                       2.7%  serde_json Value::deserialize
                                       2.6%  _raw_spin_lock under handle_mm_fault

The proxy's JSON cost is **tree construction and destruction**, not parsing — which matters, because a binary codec that still builds a `Value` keeps most of it.

Attributing the lane calls by caller settled where the volume comes from: **`matrixark_scan_candidates` is 87% of all time spent inside lane calls**, at ~93 round trips and ~448 records read per user message.

## Everything here is additive and default-off

An existing deployment that upgrades and changes nothing behaves identically.

| change | default | effect if untouched |
|---|---|---|
| `record_fields` projection | absent = whole record | none |
| `query_vector` / dense ranking | absent = lexical, as today | none |
| msgpack lane | off unless `MATRIXARK_LANE_CODEC=msgpack` | none |
| gateway heap trim | on, off via `MATRIXARK_GATEWAY_TRIM_BYTES=0` | reclaims only unused pages |
| per-type freshness token | on | narrows *when* a cached scan is reused, never what it returns |

## Both lane transports use the fast parser

This module already selected orjson for lane payloads, measured earlier at −14.1% gateway CPU — but only the old lane path used it. The socket transport and the HTTP transport added in #1364 both called `json.loads` directly, so the busiest payloads in the process went through the slow decoder while the fast one sat twelve hundred lines above, unused. Found by profiling; reading the transport gives no hint it exists.

## The lane can stop being text

msgpack behind `MATRIXARK_LANE_CODEC=msgpack`, using the same `rmp_serde::to_vec_named` the stdio lane already trusts. No negotiation and no header: a msgpack map begins 0x80–0x8f, 0xde or 0xdf and a JSON object begins `{`, so the first byte settles it in both directions — the rule the stdio lane already uses. A client and a proxy that disagree about the setting still understand each other, and a reply that cannot be encoded falls back to text rather than being dropped.

**Measured**, 1,500 s arms, fresh store each, zero failures, the codec verified live in both processes:

| | text | msgpack |
|---|---|---|
| messages served | 981 | 1,164 (+18.7%) |
| gateway CPU / message | 37.7 ticks | 27.9 (−26.1%) |
| proxy CPU / message | 93.2 ticks | 78.4 (−15.8%) |
| add p50 at equal message count | 1,656 ms | 1,086 ms (−34.4%) |

Retrieve p50 was ~9% **worse**, which is why it stays behind the flag. Plausibly `rmp_serde::to_vec_named` costing more than `serde_json::to_string` on the very large retrieve responses — a hypothesis, not a finding.

## The scan can return only the fields the caller reads

A record averages 673 bytes of which the text is 2.4%; `embedding_meta` is 26.9%, `envelope` 11.9%, `scope` 10.7%, `vector` 10.8%. Projecting the **same 2,954 records** took a response from 11,008,921 B / 8,979 ms to 2,159,313 B / 384 ms — 5.1× smaller, 23.4× faster, with the record count identical, which is the built-in correctness check: it drops fields, never records.

The caller names the fields; the engine never guesses, because a wrong guess is a missing field rather than a slow response. Filtering happens **before** projection, so a scan still filters on fields it does not return.

The audited list is included but **off by default** (`MATRIXARK_SCAN_RECORD_FIELDS`). It keeps 47 of 59 fields — about 80% of bytes — while the engine pays a set lookup per field per record, and the first soak with it enabled came out slower than the arm without. The mechanism is right; that list is too conservative to pay for itself. The aggressive projection it exists for becomes available once ranking moves into the engine.

## The scan cache stops missing on every write

Its key held the storage prefix's **total record count**. Correct, and maximally coarse: every append changed it, so every scan of every type missed. At 229 records written per user message the hit rate under ingest was effectively zero, and each miss re-fetched the whole corpus of the requested types.

A scan that names its types can only be changed by those types, so it now keys on their versions plus a delete epoch; a scan naming no types keeps the total count. The version rides the same durable batch as the data and the index entries, so it can never lag a committed append, and the types are already computed there for the type index.

Deletes get an epoch rather than a per-type version deliberately: removing the last records of a type whose version key does not exist yet would leave a per-type token unchanged and the next scan would still contain them. Coarse, and the right trade because deletes are rare and appends are not.

## Ranking can now happen in the engine

It could not before, and the reason was not plumbing. The native packer assembles a complete pack and is wired and enabled — but it scored lexical substring matching and reported `ranking_uses_vectors: false`, while the caller ranks densely. Ranking there would have silently turned semantic retrieval into keyword matching, where a paraphrase scores exactly 0.

The missing piece was that the query embedding never reached the engine. The engine has had `cosine_similarity` all along. So the request carries `query_vector`, candidates carry their own vector through the cached snapshot, and scoring is cosine when a query vector arrives and lexical when it does not. `ranking_uses_vectors` now **reports** which scorer ran instead of being a constant.

Three judgements, each a silent quality bug if taken the other way: a mismatched vector length scores 0 rather than comparing a prefix (this store has held 32-dimension vectors labelled as a 1024-dimension model — they are not less relevant, they are not comparable); a candidate with no usable vector falls back to its lexical score rather than 0, because it is a record we could not compare rather than one we know to be irrelevant; a zero vector scores 0 without dividing by zero.

**Dense ranking is not enabled by default.** It changes which records come back, and no latency measurement would catch that going wrong. It needs a recall comparison against the current ranking first.

## Tests

83 proxy tests and 14 transport tests.

- The borrowed index parse is checked against the `Value` path's own answer over 17 payloads and goes red if the scope-key sources are reordered; a companion test asserts an ordinary record does **not** fall back, without which that agreement would be vacuous.
- The cache bound has a positive control that it is not simply evicting everything — how a bound can satisfy a bound-test while destroying what it bounds.
- Projection: only the named fields survive; **no** projection returns the whole record byte for byte; a non-object is left alone; projection does not change which records match; a projected scan cannot share a cache entry with a whole-record one in **either** direction.
- Scorer: identical/orthogonal/opposed ordering with the exact 1.0/0.5/0.0 mapping, scale invariance, mismatched lengths, zero vectors not producing NaN, a partial vector rejected rather than half-read, and the reported flag following the scorer that actually ran.
- Transport: seven cases run the identical request down **both** codecs and compare against the pipe's answer, on a namespace unique per run — a fixed one resolves to the same `/tmp` directory every time, and a store left by a build of another lineage makes the suite report a transport failure that is really a stale store.

## Reported honestly

- The **byte bound on the snapshot cache measured inert**. Set to 384 MiB and verified applied by reading `/proc/<pid>/environ`, resident memory followed the same curve. Removing the double *copy* was worth it; bounding the *contents* was not, because the contents were never the bytes. Kept because an unbounded structure with no eviction is worth bounding, but it earns none of the numbers above.
- msgpack and the freshness token each produced ~+18% throughput and ~−21% p50, within 1% of each other. The box is ~84% idle at 4 cores and load 0.9, so throughput here is bounded by **serialization**, not CPU: under strict serialization throughput is 1/latency, and both changes cut latency by about a fifth. They are two real latency wins measuring the same ceiling, not two independent throughput wins.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -100,6 +100,24 @@ struct RecordLogRequest {
     /// cap on the union would drop the very records that make deleted memories stay deleted.
     #[serde(default)]
     newest_by_type: Option<BTreeMap<String, usize>>,
+    /// Return only these top-level fields of each record.
+    ///
+    /// A scan returns whole records, and a caller uses a handful of fields. Measured on a
+    /// production store, the average record is 673 bytes of which the TEXT is 2.4%:
+    /// `storage_options` is 17.5%, `envelope` 11.9%, `scope` 10.7%, `embedding_meta` 9.6% and
+    /// `vector` 4.4% -- and the packer that asks for these records reads none of those five, since
+    /// it ranks lexically and reports `ranking_uses_vectors: false`. About nine tenths of every
+    /// record crossing the lane is never looked at, and the gateway spends 53.5% of its CPU
+    /// decoding it.
+    ///
+    /// The CALLER names the fields rather than the engine guessing them: the engine cannot know
+    /// what a caller will read, and a projection that guesses wrong is a missing field rather than
+    /// a slow response. Absent means the whole record, so nothing that does not ask is affected.
+    ///
+    /// Filtering happens BEFORE projection, so a scan can still filter on a field it does not
+    /// return -- otherwise asking for less data would silently change which records match.
+    #[serde(default)]
+    record_fields: Option<Vec<String>>,
     /// Identity ids to remove, for `matrixark_delete_records`. Sent by the caller, which owns the
     /// decision about what a delete covers; the engine only matches and removes.
     #[serde(default)]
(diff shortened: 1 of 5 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Profile-driven follow-up to #1364: the gateway was 53.5% stdlib JSON decode and the proxy's cost is Value tree building, so the lane can now speak msgpack (gateway CPU/message -26%, +18.7% throughput, add p50 -34% at equal message count), the scan can return only the fields a caller names (same 2,954 records: 11.0 MB/8,979 ms -> 2.16 MB/384 ms), and the scan cache stops missing on every write. Everything is additive and default-off, so an existing deployment that upgrades and changes nothing behaves identically. Two results are reported as negative rather than claimed: the snapshot-cache byte bound measured inert, and the audited 47-field allowlist ships OFF because it was slower than not projecting.

PR #1372, opened 2026-09-09
Rank in the engine on the retrieve path the gateway actually uses

bjmeetsfo:
## The flag never reached the path it was meant to change

`MATRIXARK_NATIVE_DENSE_RANKING` attached the query vector and the ranking policy in
`_try_native_context_pack`, which belongs to one `retrieve` implementation. The one-box gateway
resolves `retrieve()` to the local adapter's, which builds its **own** native-pack request and
carried neither field. So the engine received a pack request with nothing to rank densely with,
the flag changed nothing, and an A/B across it compared two identical arms.

This was confirmed at runtime rather than inferred. Instrumenting both branches on a copy of the
tools tree, with the flag on and 10/10 retrieves served, **neither** the query-vector helper nor
the pack call site was entered. The probe was not blind: the gateway's `PYTHONPATH` and cwd both
pointed at the patched copy and both modules had fresh `.pyc` files.

After the fix the same probe shows the flag flipping behaviour — ten query vectors computed with
it on, none with it off.

## The flag stays off by default

At equal elapsed time the arms return the same refs per retrieve (21.2 against 21.3) and the
engine arm's retrieve p50 is worse (541 ms against 617 ms) on a single sample. There is no
measurement here that justifies changing a default, so this only makes the flag *work* — it does
not turn it on.

Worth recording separately: the query embedding handed to the engine is **32-dimensional** while
the encoder service answers with 1024, so this ranking is cosine over hash vectors. That is
pre-existing and shared with the Python packer — the same vector is requested moments later — but
it caps what any ranking change can achieve on this box.

## A hazard the fix exposes, now held by tests

The caller sends `ranking_weights` on **every** request but sends `query_vector` only when the flag
is on, and on a one-box those weights are dense 1.00 / sparse 0.00. Blending an absent dense term
would compute `1.00 * 0 + 0.00 * lexical = 0` for every candidate and rank the whole corpus flat —
silently, with no error and a full-looking pack.

The engine already fell through to lexical instead. `candidate_score` now states that in one place
and four tests hold it, including a positive control that the blend genuinely runs when both
vectors are present, so the fallback tests are not asserting a branch nothing else would take. The
index hint is a lazy closure so the lexical path does not pay for a lookup it discards, and one
test asserts that laziness.

## Gate

`cargo test -p temporalstore-rust --bin matrixark_rust_proxy` — **94 passed, 0 failed**.

That specific target matters: this file is reached through `src/bin/`, so `cargo test --lib`
contains none of its tests and reports `ok` while running zero of them. `--bins` also misses it,
because cargo stops after the first failing test binary. Seven Python test files covering the
native-pack request also pass.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -81,6 +81,17 @@ struct RecordLogRequest {
     /// unaffected.
     #[serde(default)]
     query_vector: Option<Vec<f32>>,
+    /// How to combine the dense and lexical scores, and what an index hint is worth.
+    ///
+    /// The CALLER owns ranking policy and the engine executes it. That is deliberate: the caller
+    /// already computes these weights from its own configuration -- on the one-box they are 1.00
+    /// dense and 0.00 lexical, elsewhere 0.72 and 0.28 -- and an engine that hardcoded its own
+    /// would silently rank differently from the caller that used to do it, which is the failure
+    /// this whole change has to avoid.
+    ///
+    /// Absent means dense-only (1.0 / 0.0 / 0.0), which is what this engine already did.
+    #[serde(default)]
+    ranking_weights: Option<RankingWeights>,
     /// Send record payloads as sub-documents instead of JSON strings (see `RecordPayload`).
     /// Absent means the historical string shape, so an older reader is unaffected.
     #[serde(default)]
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. MATRIXARK_NATIVE_DENSE_RANKING never reached the one-box gateway retrieve path, so every measurement across that flag compared identical code; this attaches the query vector and ranking policy to the request the live path actually builds, and adds a guard (with tests) so weights arriving without a vector cannot score every candidate zero. The flag stays OFF by default - the corrected A/B shows no ranking gain and a slightly worse retrieve p50.

PR #1373, opened 2026-09-09
Say when a scoped scan gives up and reads the whole store

bjmeetsfo:
## A degraded window did ten whole-corpus reads and reported nothing

Measured on a native one-box: with the proxy killed, one request produced **10** `read_all` calls
and wrote nothing anywhere. From outside it was indistinguishable from a request that was merely
slow.

Two causes, both fixed here:

1. `_note_full_read_fallback` returned early unless `MATRIXARK_MCP_DEBUG_LOG` named a file, which
   it does not by default.
2. The branch that actually fires when the backend cannot answer — `subset is None` — **never
   called the notifier at all**. All nine existing calls sit on the exception path beside it. Six
   such branches read every record in the store and said nothing.

## What changed

The count is now kept unconditionally: a dict increment against a caller about to read the entire
store, so it is not worth a flag to skip. `full_read_fallback_counts()` exposes it. The detailed
line still requires the debug log, so no deployment starts writing lines it did not ask for — and a
test asserts that, because "counts always" quietly becoming "writes always" is the obvious way this
change could go wrong.

`where` defaults to the calling function name so a branch with no exception can announce itself,
and `reason` is optional because "the scan could not answer" is not something anything raised.

## Why this, and not deleting read_all

`read_all` is already off the healthy serving path. Measured on a native one-box with the stack
verified healthy first — ingest 202, and ingest 500 with the proxy killed, which proves the gateway
was really on the backend under test rather than quietly serving from a local adapter:

| traffic | `read_all` calls |
|---|---|
| 30 ingests + 20 retrieves, healthy | **2** (one-time init, not per-request) |
| one request, proxy down | **10** |

So what is left to retire is the fallback, not the serving path. A fallback nobody can see firing
cannot be retired safely — this makes it visible first.

## Tests

`tools/test_a_scoped_scan_that_gives_up_says_so.py`, five checks, all pass. It fails on unpatched
code. Being precise about that: it fails there by `ImportError` on the new symbol, so the assertion
that genuinely discriminates is the source-level one — six silent branches must become zero.

### DIFF
--- a/tools/matrixark_mcp_temporal_adapters.py
+++ b/tools/matrixark_mcp_temporal_adapters.py
@@ -534,27 +534,59 @@ def _shadow_log_path(operation: str) -> str:
     return _os.environ.get("MATRIXARK_SHADOW_LOG", "/tmp/matrixark_%s_shadow.log" % operation)
 
 
-def _note_full_read_fallback(where: str, reason: BaseException) -> None:
-    """Say which scoped scan gave up and why, when the answer is to read the whole store.
+_FULL_READ_FALLBACKS: dict[str, int] = {}
+
+
+def full_read_fallback_counts() -> dict[str, int]:
+    """How many times each scoped scan gave up and read the whole store instead.
+
+    Counted in this process since start. Exposed because the fallback is otherwise undetectable
+    from outside: a measured degraded window did TEN whole-corpus reads and wrote nothing at all,
+    and from the caller it looked like a normal request that was merely slow.
+    """
+    return dict(_FULL_READ_FALLBACKS)
+
+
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. A degraded window was doing ten whole-corpus reads and reporting nothing: the notifier was gated behind a debug log that is unset by default, and the branch that fires when the backend cannot answer never called it at all. Counts are now unconditional and exposed via full_read_fallback_counts(); the detailed line still needs the debug log. No fallback was removed - read_all is already off the healthy serving path (2 calls across 50 requests), so making the fallback visible is the prerequisite to retiring it.

PR #1374, opened 2026-09-09
A full read that goes back to the refused backend is not a fallback

bjmeetsfo:
## Two different answers, treated as one

A scoped scan returns `None` for two unrelated reasons and every caller treated them identically:

- it could not be **asked** — this client has no scanner, so the full read is the correct path
- it was asked and **failed** — in which case the full read calls the very client that just refused

Measured on a native one-box: one request with the backend down did **10** whole-corpus reads and
returned 500 anyway. None of them could have answered. On a large store they would have been
expensive as well as useless.

## What changed

`_scan_records_of_types` now records *why* it gave up, and the six branches route through
`_read_all_after_scoped_scan`, which re-raises the scan's own error when the full read provably
cannot help. The request fails exactly as it did before, with the real reason — a refused
connection reads as a refused connection rather than as a slow request that eventually gave up.

"Provably cannot help" is deliberately narrow, and both escapes keep today's behaviour:

| escape | still does the full read |
|---|---|
| warm Python hot cache | yes — off by default on native backends (`python_hot_cache_allowed` returns `backend_label == "local"`), but honoured wherever it is on and holding records |
| configured disk fallback store | yes — `read_all` recovers from disk before reading |

Anything unexpected counts as "it can help": if the policy cannot be read, the full read happens as
before. Turning a served request into an error is the one outcome this must not cause.

The state is **thread-local**. One adapter serves every request, so an attribute on the instance
would let one thread's outage decide another thread's fallback — a test covers exactly that.

## Why this is what "retire read_all" comes down to

`read_all` is already off the healthy serving path (mx#1338/1340), confirmed by measurement with
the stack verified healthy first:

| traffic | `read_all` calls |
|---|---|
| 30 ingests + 20 retrieves, healthy | **2** — one-time init, not per-request |
| one request, backend down | **10** |

So the remaining calls were all on paths where it could not work. This removes those; it does not
remove the fallback where the fallback can genuinely serve.

## Tests

`tools/test_a_full_read_that_cannot_answer_is_not_a_fallback.py` — six checks, all pass, including
a positive control that an **empty** hot cache does not count as an offline source (treating it as
one would quietly restore the futile read) and the thread-locality check.

Also fixes a test shipped in the previous change: it asserted a literal source string, which broke
the moment those branches were routed through a helper. The invariant is that nothing reads
everything silently, not that a particular line appears six times.

Neighbouring suites: 25 pass. Four `backend_policy` suites fail by `SkipTest` at import and are red
on main too — verified by name-diff, not by comparing red counts.

### DIFF
--- a/tools/matrixark_mcp_temporal_adapters.py
+++ b/tools/matrixark_mcp_temporal_adapters.py
@@ -536,6 +536,13 @@ def _shadow_log_path(operation: str) -> str:
 
 _FULL_READ_FALLBACKS: dict[str, int] = {}
 
+# Why the most recent scoped scan in THIS thread gave up, or None if it simply could not be asked.
+#
+# Thread-local rather than an attribute on the adapter: one adapter serves every request, so an
+# instance attribute would let one thread's backend error decide another thread's fallback. Reset
+# at the top of every scan, so what it holds is always the scan the caller is reacting to.
+_SCAN_STATE = threading.local()
+
 
 def full_read_fallback_counts() -> dict[str, int]:
     """How many times each scoped scan gave up and read the whole store instead.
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. A scoped scan returning None meant both "no scanner here" and "the backend refused"; the second sent callers to a full read against the same refused client, measured at ten whole-corpus reads for one request that 500d anyway. Those are now skipped and the scan error re-raised instead, with the hot-cache and disk-fallback paths left intact so nothing that could serve stops serving.

PR #1392, opened 2026-09-10
The retrieve snapshot stops copying what it never reads

bjmeetsfo:
The retrieve candidate snapshot copied several things it never read, and did one pass it never
used. Each is paid per record, per rebuild — and the snapshot is invalidated on every write, so an
ingesting store rebuilds it on close to every retrieve.

### 1. Every record was deep-cloned out of its payload

`flatten_context_payload` did `records.push(item.clone())` for each record in a bundle.
`decode_matrixark_payload` — the same operation on the append path — was already fixed to
`std::mem::take` the bundle, and its note records why: sampling the proxy under sustained ingest
put `BTreeMap::clone_subtree` and `Vec<Value>::clone` among the hottest frames. The fix reached one
of the two, and was left in the worse of the two places: this one copies the whole corpus per
rebuild, embedding vectors included, where one 1024-dim vector is 1024 individually cloned
`Value`s. The parsed payload is local and owned, so moving the records out is free.

### 2. The whole shard payload map was cloned to read it

`hgetall_map` is `(*hgetall_shared(..)?).clone()`. The rebuild loop only iterates the values and
drops them, so `for payload in hgetall_map(engine, key)?.values()` copied every payload string in
the shard for nothing. `hgetall_shared` returns the `Arc` that exists so a reader does not have to.
Only this call site moves — the other `hgetall_map` callers consume the pairs by value, so handing
them an `Arc` would relocate the copy rather than remove it.

### 3. Each candidate carried a lowercased second copy of its text

`CachedRetrieveCandidate` held `lower_text` — `text.to_ascii_lowercase()` — purely so the lexical
scorer could call `contains`. That copy is *resident*: the snapshot lives behind an `Arc` in a
cache and is reused across requests, so it duplicated the text already in `selected_ref` for the
life of the entry. It also defeated the dense-only skip — the lexical *score* was made lazy
earlier, but the string it scores was still built eagerly for every candidate, so the one-box
default still paid.

The obvious fix — drop the copy and match case-insensitively over the original — was tried and
**measured worse**: the blended scoring pass went 7.05 ms to 25.63 ms over 4,000 candidates, 3.6x
slower, because `str::contains` is a SIMD-accelerated two-way search and a hand-rolled case-folding
scan is a byte loop. So the copy is *deferred* rather than removed: `OnceLock<String>`, never built
under the one-box weights, built exactly once per candidate under blended weights, with `contains`
unchanged. Strictly better than the eager field in both regimes rather than better in one and worse
in the other.

### 4. Each ref was built by cloning out of a record that was about to be dropped

`context_record_text` cloned the text out of the record and `selected_ref_from_record` cloned that
`String` again into the ref, so the largest field on a serving record was copied twice per
candidate. The ref also cloned eight collections — `source_roles`, `source_role_counts`,
`source_hook_types`, `source_hook_type_counts`, `source_codex_events`,
`source_codex_event_counts`, `source_session_ids`, `source_entity_hashes`. The memory inventory and
the index-term pass both run first and both only borrow, and the records are dropped as soon as the
candidates exist, so the build now consumes them and takes what it needs.

`take_record_field` is deliberately `get_mut(key).map(Value::take).unwrap_or_else(empty)` with no
null filter: that is exactly what `get(key).cloned().unwrap_or_else(empty)` returned. Absent and
explicitly `null` stay distinguishable, so a field written as `null` still arrives as `null`.
Repairing that here would be a data change wearing an optimization's clothes. The takes are
separate statements rather than a closure — a closure capturing the record mutably would still be
live during the `json!` block, which reads other fields of the same record immutably.

### 5. The index-term maps were built even when nothing would read them

Their only consumer is the secondary-group filter, which returns early on
`secondary_groups.is_empty()`. With no groups requested that was a whole extra pass over the
corpus, plus a `String` and a `HashSet` insert for every `context_index` record, to produce maps
nothing looks at.

### 6. The filter chain ran the expensive predicate first

`scope_matches_record` builds a candidate scope key, walks eight identity fields and calls
`record_scope_sources` — which allocates — for each one it enforces. `is_serving_context_record` is
one `get` and a `matches!` on a `&str`, and it is the *selective* one: it rejects every
`context_index` and `context_embedding` record, which is where most of the corpus lives. All three
predicates are pure and AND-ed, so the order changes the cost and cannot change the result.

### 7. The whole context pack was deep-cloned into the response

The top-level branch cloned every key and value out of `response`, and one of those values is
`context_pack` — the entire assembled pack, every selected ref and every ref's text. `response` is
local, the other branch is mutually exclusive, and nothing reads it afterwards, so it can be taken
apart.

### Measured

`what_the_copies_on_the_retrieve_path_cost` times each operation with both shapes written out, so
one binary compares them and each pair builds its fixture identically:

| operation | clone | move / take | saved |
|---|---:|---:|---:|
| extract 2,048 records from bundles | 47.9 ms | 0.23 ms | 99.5% |
| a ref's text + 8 `source_*` collections, ×2,000 | 3.28 ms | 2.03 ms | 38.1% |
| take a 1,000-ref response apart, ×20 | 5.52 ms | 0.02 ms | 99.6% |

**The same bench reports the number that matters more than any of those: parsing those 2,048
records costs 878 ms — about 429 µs per record, 18x the extract clone.** Neither shape avoids it.
On a snapshot rebuild the parse is the cost. These copies were worth removing and they are not the
lever; the lever is not re-parsing shards that cannot have changed, and that is separate work.

### On the end-to-end numbers: there are none, and that is the honest answer

A one-box A/B was run properly — fixed message count, ABBA ordering, total ticks over total
messages, two binaries verified by checksum and by the listening process's own cmdline — and it
**cannot resolve a change of this size on this hardware**. Three rounds spread 41% between
identical baseline arms and 350% between identical patched arms, with one arm four times the cost
of its twin. Every reported effect was smaller than that spread. Four separate harness defects were
found and fixed along the way (a time-boxed loop, a tick window that did not match its denominator,
a teardown pattern that missed two of three binary directories, and an encoder gate that could not
fail) and the variance did not come down. So the claims here are the bench's, scoped to the
operations it measures, and no end-to-end percentage is offered.

### Tests

`the_lowered_text_is_built_once_and_only_when_asked` pins both halves of the laziness — nothing is
lowered before something asks, and the second ask returns the same borrow rather than rebuilding.
`a_candidate_without_text_lowers_to_nothing` covers a record with no text, which must score rather
than panic. `the_lexical_score_is_the_share_of_terms_present` pins the scorer.
`moving_a_bundle_out_returns_what_cloning_it_returned` carries a nested vector in its fixture on
purpose: it is the largest thing a record holds and the part a shallow move would be most likely to
get wrong. `a_payload_that_is_not_a_bundle_is_taken_whole_or_not_at_all` covers a torn payload,
because one bad shard must not cost the whole retrieve.

`cargo test -p temporalstore-rust --bin matrixark_rust_proxy -- --test-threads=1`: 117 passed, 0
failed. That is the only command that reaches this file — `--lib` and `--bins` both run zero of its
tests and report ok.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -271,8 +271,20 @@ struct HashReadRecord {
 #[derive(Clone, Debug)]
 struct CachedRetrieveCandidate {
     selected_ref: Value,
-    lower_text: String,
     ref_type: String,
+    /// The lowercased text, built ON FIRST USE and never under a dense-only policy.
+    ///
+    /// This was an eager `String`. The snapshot is cached behind an `Arc` and reused across
+    /// requests, so that copy was RESIDENT for the life of the cache entry, duplicating the text
+    /// already in `selected_ref` -- the largest field a serving record has. It also defeated the
+    /// dense-only skip: mx#1390 made the lexical SCORE lazy, but the string it scores was still
+    /// built for every candidate, so the one-box default still paid for it.
+    ///
+    /// Searching the original text case-insensitively instead was tried and MEASURED SLOWER --
+    /// 7.05 ms to 25.63 ms over 4,000 candidates, because `str::contains` is a SIMD-accelerated
+    /// two-way search and a hand-rolled case-folding scan is not. Deferring the copy keeps that
+    /// search and still costs nothing when it is never needed.
+    lower_text: OnceLock<String>,
     /// The record's own embedding, kept so ranking can be dense.
     ///
     /// Held on the candidate rather than re-read from the record at scoring time because the
@@ -5646,7 +5658,9 @@ fn load_retrieve_candidate_snapshot(
     let mut records = Vec::new();
     for shard in 0..shard_count {
         let key = format!("{record_hash_key}:{shard:06}");
-        for payload in hgetall_map(engine, key)?.values() {
+        // `hgetall_shared`, not `hgetall_map`: the map form clones the shard's whole payload
+        // map, and this loop only reads it. The `Arc` temporary lives for the loop.
+        for payload in hgetall_shared(engine, key)?.values() {
             if payload.trim().is_empty() {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The retrieve candidate snapshot stops deep-cloning records out of bundles, stops cloning each shard payload map to read it once, defers the lowercased text copy behind a OnceLock, takes the ref fields instead of cloning them, skips the index-term pass when nothing reads it, and stops deep-cloning the context pack into the response. 117 proxy-bin tests pass. Numbers come from a bench over the operations rather than a soak: the one-box A/B spread 350% between identical arms, so no end-to-end percentage is claimed.

PR #1393, opened 2026-09-10
A shard is parsed once per write, not once per retrieve

bjmeetsfo:
The candidate snapshot is invalidated on every write, so an ingesting store rebuilds it on
essentially every retrieve — and the rebuild re-parses the entire corpus each time. Records are
appended, so every shard but the newest parses to exactly what it parsed the time before.

The bench added with the previous change measured what that costs: **parsing 2,048 records takes
878 ms, about 429 µs per record.** That is 18x the per-record copy that change removed. On a
rebuild, the parse *is* the cost.

### Where the parsed records live, and why it is there

Beside the payloads they came from, in the `SnapshotEntry` the payload cache already holds — not in
a cache of their own. That placement is the entire correctness argument: **the write path already
patches and drops these entries, so a stale parse is dropped by the same call that makes it
stale.**

A separate cache would need its own invalidation, and the cheap identity tests available to one are
all wrong here. `Arc::ptr_eq` cannot see a `patch`, because `SnapshotCache::patch` goes through
`Arc::make_mut`, which mutates in place at refcount one — and deliberately keeping the `Arc` alive
elsewhere to force copy-on-write would make every write clone a whole shard payload map, which is a
far worse trade on the write path than the parsing this removes.

### The decodes get their own budget

Charging parsed records to the payload budget would have been the obvious thing and is a trap. A
`Value` tree is several times the text it came from, so the payload cache would hold roughly a
seventh as many shards, and **every shard that stopped fitting would be re-read from the engine as
well as re-parsed** — a change meant to remove parsing causing more work than it saves, showing up
only on the large stores that need it most.

So `decoded_bytes` is budgeted separately (half the payload budget), with its own LRU eviction that
drops a parse and leaves its payloads in place. The two are worth keeping on different terms:
evicting a decode costs a parse, evicting a payload costs a read as well. The payload cache's
behaviour is unchanged, and this is a bounded addition on top of it — bounded in **bytes**, because
this cache's own note records that an entry-count cap was tried here before and moved the resident
set by nothing, the count never having been what was large.

`DECODED_BYTES_PER_PAYLOAD_BYTE` is a deliberate over-estimate. The real factor varies by record
shape, and the cost of guessing low is an unbudgeted cache — the failure this type exists to
prevent.

### What it is worth, and how that number was arrived at

Derived, not measured end to end — stated that way on purpose. From the bench in
`what_the_copies_on_the_retrieve_path_cost`, for 2,048 records:

| | cost |
|---|---:|
| parse (what an unchanged shard paid on every retrieve) | 878 ms |
| clone the cached records (what it pays now) | 48 ms |

So an unchanged shard costs roughly **a eighteenth** of what it did. The clone remains because the
candidate build takes fields out of records it owns; it costs ~23 µs per record against ~429 µs to
parse one, so it buys the reduction and gives up nothing. Removing it as well means handing the
snapshot shared candidates, which is a larger change to how the snapshot is shaped.

No end-to-end percentage is offered. The one-box A/B spreads 350% between identical arms on this
hardware, which is documented on the previous change and has not improved.

### Tests

`patching_a_shard_drops_what_was_parsed_from_it` is the one that matters: it patches a cached shard
and asserts the parse is gone and no longer charged. `a_parse_is_only_kept_for_a_shard_the_cache
_still_has` covers the entry disappearing while a parse is in flight — the result then describes
nothing and must be discarded rather than resurrect the key.
`a_parse_is_evicted_without_taking_its_payloads_with_it` pins the separate budget by asserting that
an over-budget parse is dropped while its payloads stay cached and still serve.

`cargo test -p temporalstore-rust --bin matrixark_rust_proxy -- --test-threads=1`: 120 passed, 0
failed.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -1379,7 +1379,16 @@ fn required_option(value: Option<String>, name: &str) -> Result<String, String> 
 /// One cached hash, with what it costs and when it was last wanted.
 struct SnapshotEntry {
     map: Arc<BTreeMap<String, String>>,
+    /// The shard's records, already parsed, or `None` if nothing has asked for them yet.
+    ///
+    /// Kept HERE rather than in a cache of its own so that it is dropped by the same call that
+    /// makes it stale: the write path patches and removes these entries, and a patched payload
+    /// map with a stale decode beside it is the one shape that must not exist.
+    decoded: Option<Arc<Vec<Value>>>,
+    /// The payload bytes. `decoded` is charged separately, so this stays comparable to `weigh`.
     bytes: usize,
+    /// What the decoded records are charged at, or 0 when there are none.
+    decoded_bytes: usize,
     used: u64,
 }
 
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. A record shard is now parsed once per write instead of once per retrieve: the parsed records sit in the SnapshotEntry that already holds the payloads, so the write path drops a stale parse in the same call that makes it stale, and they carry their own byte budget so the payload cache is not squeezed. The bench puts a parse at 429us per record against 23us to copy one, so an unchanged shard costs about an eighteenth of what it did. 120 proxy-bin tests pass.

PR #1394, opened 2026-09-10
A retrieve says where its time went, and the rebuild stops copying records

bjmeetsfo:
A retrieve could report the scoring pass and a total, and nothing in between, so a slow one could
not be attributed without attaching a profiler to a running proxy. It can now, and the first thing
it said was that **the snapshot rebuild is 1134.9 ms of an 1167 ms retrieve — 97.2%.** The scoring
pass, which has had most of the attention, is 4.4 ms.

### The retrieve says where its time went

`snapshot_ms` sits beside `score_ms` in the pack's metrics, and a retrieve past a threshold logs one
line with both, the records scanned and the refs selected. The threshold is a single constant with
an env override, and 0 silences it.

### The cached parse changed which shape is cheaper, and this puts it back

The previous change made the ref builder *take* the eight `source_*` collections out of the record
rather than copy them, worth 38% of that step. That was right for its context: the records were
freshly parsed, owned, and dropped immediately, so taking was free.

Caching the parse removed that context. The records now belong to the shard cache, so the only way
to take from one is to **clone the whole record first** — embedding included — to save the copies
the take was avoiding. Measured on the same bench, per 2,000 refs:

| | |
|---:|---:|
| own the record, then build the ref | 96.5 ms |
| borrow it, then build the ref | 34.5 ms |
| | **64.3% saved** |

So the builder borrows again and the copies are back. It is one caller, so the fix is to change the
builder rather than add a borrowing sibling beside it — a second ref builder is how this file
already ended up with two packers and a fix that reached one of them.

`native_retrieval_memory_inventory` takes borrowed records for the same reason; the other packer
owns its records and goes on using them, so it passes a borrow view rather than changing shape.

### Measured on the retrieve itself

Same code path, before and after, taken from the proxy's own phase log. The record counts differ
between runs, so the per-record figure is the comparable one:

| | records | snapshot | per record |
|---|---:|---:|---:|
| before | 25,670 | 1134.9 ms | 44.2 µs |
| after | 22,687 | 672.0 ms | **29.6 µs** |

A third off the rebuild, which is 97% of a retrieve after a write. A retrieve with the snapshot
still cached is 229 ms against 919 ms after a write, so this is the half that matters.

No end-to-end percentage is offered: the one-box A/B spreads 350% between identical arms on this
hardware. Every number here comes from a bench over the operation or from a single request's own
phase timings, both of which are immune to that.

### What is left

The rebuild is still O(corpus) — every record is filtered, counted into the inventory and turned
into a ref on every rebuild, at ~29.6 µs each. Making it O(changed shard) means caching the
candidates per shard the way the parse now is, which is a larger change: the candidates depend on
the query scope, and the inventory would have to be counted per shard and merged.

### Tests

`cargo test -p temporalstore-rust --bin matrixark_rust_proxy -- --test-threads=1`: 120 passed, 0
failed. The bench's ref pair was re-pointed at the trade that now exists — measuring take against
copy would keep reporting a win for a shape the code no longer uses.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -5769,14 +5769,15 @@ fn load_retrieve_candidate_snapshot(
     } else {
         (count + DIRECT_RECORD_LOG_SHARD_SIZE - 1) / DIRECT_RECORD_LOG_SHARD_SIZE
     };
-    let mut records = Vec::new();
+    // Parsed at most once per shard per write, and then borrowed from the shard cache rather
+    // than copied out of it: copying a record to own it costs ~23 us against the ~0.6 us that
+    // owning it saves the ref build.
+    let mut shards = Vec::with_capacity(shard_count);
     for shard in 0..shard_count {
         let key = format!("{record_hash_key}:{shard:06}");
-        // Parsed at most once per shard per write. The clone that follows is what lets the
-        // candidate build keep taking fields out of records it owns; it costs ~23 us per record
-        // against ~429 us to parse one, so it buys the 18x and gives up nothing.
-        records.extend(hgetall_decoded(engine, key)?.iter().cloned());
+        shards.push(hgetall_decoded(engine, key)?);
     }
+    let records: Vec<&Value> = shards.iter().flat_map(|shard| shard.iter()).collect();
 
     // Built only when something is going to read them. The single consumer is the secondary-group
     // filter below, which returns early on an empty group list -- so with no groups this was a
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. A retrieve now reports its snapshot rebuild alongside the scoring pass, which showed the rebuild is 97.2% of a retrieve after a write. The rebuild also stops cloning records: caching the parse made owning a record cost more than the field-taking it enabled, so the ref builder borrows again. Rebuild cost per record 44.2us to 29.6us, from the proxy own phase log. 120 proxy-bin tests pass.

PR #1396, opened 2026-09-10
The pack redundancy filter stops comparing every item with every other

bjmeetsfo:
With the proxy's retrieve down to ~14 ms on a warm snapshot, the gateway became the whole of a
retrieve. Sampling it under load — a sampling profiler, not cProfile, which misattributes a
function called a million times, and this is exactly that shape — put **one function at 76.8% of
gateway on-CPU time**, in four adjacent lines:

```
  26.6%  drop_redundant_pack_items (matrixark_mcp_context_pack.py:1395)   id(other) in redundant
  24.2%  drop_redundant_pack_items (matrixark_mcp_context_pack.py:1397)   len(other_text) <= len(text)
  13.8%  drop_redundant_pack_items (matrixark_mcp_context_pack.py:1394)   the loop itself
  12.2%  drop_redundant_pack_items (matrixark_mcp_context_pack.py:1399)   text in other_text
```

It compared **every pack item against every other**, and a pack carries up to
`max_selected_refs` items — 1,000 by default, so a million pair tests per retrieve. The line split
is worth reading: the substring test is only 12%. Most of the cost was the loop and its two guards.

Worth noting separately: JSON decode of the proxy lane is now **0.9%** of gateway CPU. The old
finding that it was ~68% is superseded — the orjson lane change fixed it, and the bottleneck moved.

### The change

Only a **strictly longer** item can contain another, so the indices are ordered once by length and
each item's scan covers exactly the entries whose length exceeds its own. The scan runs over
parallel lists by index, so a visit costs an index and a compare rather than a tuple unpack, an
`id()` and a set lookup.

### Why the order is free to change

The old scan skipped a container that was itself already marked redundant, which makes the outcome
look order-dependent — and the pruning above changes the order, so this has to be settled rather
than assumed.

It is not order-dependent. **Containment is transitive, and it carries the length condition with
it:** if X is inside Y and Y is inside Z then X is inside Z, and `len(Z) > len(Y) > len(X)`. So
"contained in some longer item that is not itself redundant" and "contained in some longer item"
select exactly the same items. That check only ever saved work, never changed the answer.

### Measured

| | |
|---|---:|
| the function, on a 1,000-ref pack | **98.4 ms → 27.4 ms** (3.6x) |
| warm retrieve, end to end | 224 ms → 119 ms |
| of which the proxy reports | 14 ms |

So the gateway's share went from roughly 205 ms to roughly 107 ms. The end-to-end figures are from
a single request's own timings on each side, not a run-to-run comparison — this hardware spreads
350% between identical arms, which is why nothing here is quoted as a soak percentage.

### Tests

`test_a_chain_of_containments_keeps_only_the_longest` pins the transitivity argument directly: X
inside Y inside Z leaves only Z, and it asserts that for three different arrival orders.

`test_it_agrees_with_the_plain_definition_on_random_packs` compares the sweep against the plain
definition — "contained in some longer item" — over 60 randomised packs, rather than against
hand-written expectations. A hand-written table tests the cases the author thought of, and the risk
in a pruning change is the case they did not.

The five existing cases are unchanged and still pass: 9 tests, all green.

### DIFF
--- a/tools/matrixark_mcp_context_pack.py
+++ b/tools/matrixark_mcp_context_pack.py
@@ -11,6 +11,7 @@ except ImportError:  # Direct script execution from tools/.
     from matrixark_mcp_env import env_bool
 
 
+import bisect
 import os as _os
 
 import os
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Sampling the gateway put drop_redundant_pack_items at 76.8% of its CPU: it compared every pack item against every other, and a pack carries up to 1,000 refs. It now scans only strictly longer items, which is sound because containment is transitive. Function 98.4ms to 27.4ms; warm retrieve 224ms to 119ms end to end. Also worth knowing: JSON decode of the proxy lane is now 0.9% of gateway CPU, so the old 68% figure is superseded. 9 tests green.

PR #1397, opened 2026-09-10
The redundancy sweep searches a few joined haystacks, not hundreds of texts

bjmeetsfo:
The previous change pruned the sweep to strictly longer items and took it from 98.4 ms to 26.4 ms
on a 1,000-ref pack. Re-profiling the gateway moved the cost onto the substring test itself --
41.8% of gateway on-CPU time on one line -- so what is left is the NUMBER of searches, about 500
per item, each its own call.

Joining the longer texts into a handful of large haystacks does the same byte scanning in a handful
of calls. On the same 1,000-ref pack: **26.4 ms to 8.4 ms**, and 98.4 ms to 8.4 ms against where
this started.

### Why it can be this simple

An item is redundant iff some LONGER item contains it, whether or not that container is itself
redundant -- containment is transitive and carries the length condition with it. So the haystack is
exactly "every text longer than this one", with no per-item bookkeeping to keep in step.

Two things it has to get right, both structural:

- **A match must not span two texts.** They are joined with a NUL, which normalized record text --
  whitespace-collapsed lowercase prose -- never contains. A pack that somehow carries one takes the
  pairwise path instead of the risk.
- **Equal-length texts must not count as containers.** A whole length group is tested before any of
  that group joins the haystack.

The pending haystack is joined once per length group rather than once per item; joining per item
would reintroduce a cost proportional to the haystack on every candidate.

### Measured end to end

A warm retrieve, with the proxy reporting its own time separately:

| | wall | of which proxy | gateway |
|---|---:|---:|---:|
| before the sweep changes | 224 ms | 14 ms | ~205 ms |
| after the length pruning | 119 ms | 14 ms | ~105 ms |
| after this | **85 ms** | 16 ms | **~68 ms** |

Single-request timings on each side, not a run-to-run comparison: this hardware spreads 350%
between identical arms, so no soak percentage is claimed.

### Tests

 covers the fallback, which is
otherwise unreachable -- the guard prevents its own hazard, so nothing in normal data will ever
exercise it. The randomised agreement test from the previous change still compares the sweep
against the plain definition over 60 packs. 10 tests, all green.

### DIFF
--- a/tools/matrixark_mcp_context_pack.py
+++ b/tools/matrixark_mcp_context_pack.py
@@ -1367,6 +1367,60 @@ def _normalized_item_text(item: Json) -> str:
     return " ".join(text.split()).strip().lower().rstrip(".")
 
 
+# Joined between pack texts so a match cannot span two of them. Normalized record text is
+# whitespace-collapsed lowercase prose and never contains a NUL.
+_PACK_TEXT_SEPARATOR = "\x00"
+
+# How much text to seal into one haystack. Larger means fewer searches per item and more bytes
+# rescanned by each; 64 KiB keeps a full 1,000-ref pack to a handful of searches per item.
+_PACK_HAYSTACK_BYTES = 65536
+
+
+def _drop_redundant_pairwise(
+    groups: list[Json],
+    items: list[Json],
+    texts: list[str],
+    lengths: list[int],
+    order: list[int],
+) -> list[Json]:
+    """The one-text-at-a-time sweep, for a pack that cannot use a joined haystack.
+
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Second half of the pack redundancy work: the sweep now searches a few joined haystacks instead of hundreds of separate texts. 26.4ms to 8.4ms on a 1000-ref pack, 98.4ms to 8.4ms overall. A warm retrieve is now 85ms end to end against 224ms, with the proxy reporting 16ms of it. 10 tests green.

PR #1400, opened 2026-09-10
A rename cannot quietly drop a durable name

bjmeetsfo:
The vocabulary migration — page to block, zone to band, slot to bucket — is deliberate, documented
and only half done. What keeps it safe is that every historical spelling survives as a serde alias,
so a store written by an older build still loads. `BlockAddressWire` states the rule in its own
note: *"Every short name carries every spelling this field has ever had as an alias, so anything
already written still loads."*

**Nothing checked that.** A rename that changed a field and its alias together would compile, pass
every other test, and make existing data unreadable — and the symptom would surface at load, on
someone else's machine, later. That is the worst shape a defect can have here.

### What the guard does

It walks the crate's sources at test time and asserts that every durable name carrying the old
vocabulary — 260 of them, from `serde(rename = ...)` and `serde(alias = ...)` — is still written
somewhere.

It checks **one direction only**. Adding a durable name is free, which is what keeps it quiet while
the migration proceeds; removing or respelling one fails with the name in the message and says what
to do about it: keep the old spelling as an alias, or, if the name is genuinely dead, delete it from
the list in the same change so the decision is written down rather than inferred.

The guard also refuses to pass vacuously: if the source walk finds almost nothing it fails rather
than reporting that all 260 names are fine.

### Why this list, and why now

These 260 names are the compatibility surface of a rename that is still in progress. Measured from
the crate today:

| rename | occurrences | files | struct fields | Python side |
|---|---:|---:|---:|---:|
| `page` → `block` | 7,302 | 112 | 642 | 2,468 |
| `slot` → `bucket` | 910 | 64 | 3 | 217 |
| `zone` → `band` | 323 | 30 | 10 | 120 |

`page` also appears in `proto/temporalstore/v1/temporalstore.proto`. With 642 field declarations
carrying the word, the remaining tranches need something standing behind them, and this is it.

Two things the survey settled that belong next to the list:

- **`zone` was renamed to `band`, not to `slab`** — `page_store_bands`, `BlockStoreBandSummary`, and
  `legacy_page_zone_aliases_ready` holding the aliases. `slab` is a *different* concept in the same
  file: `band_id_for_slab(slab_id)` derives one from the other, and the two coincide only because
  two defaults are both 1 GiB. Renaming the remaining `zone` mentions to `slab` would merge two
  concepts rather than finish a rename.
- **`slot` → `bucket` is already done internally** — `routing_bucket` is the field name. The
  surviving `routing_slot`, `first_routing_slot` and `routing_slot_count` are
  `serde(rename)`/`alias`, which is to say they are the on-disk format and must not move.

### Tests

`a_durable_name_is_never_quietly_dropped`, plus the 46 existing `block_store` tests, all pass.

### DIFF
--- a/crates/temporalstore-rust/src/block_store.rs
+++ b/crates/temporalstore-rust/src/block_store.rs
@@ -1848,6 +1848,329 @@ mod address_size_tests {
 #[cfg(test)]
 mod tests {
 
+    /// A rename may not quietly drop a durable name.
+    ///
+    /// The vocabulary migration -- page to block, zone to band, slot to bucket -- is deliberate and
+    /// only half done, and what keeps it safe is that every historical spelling survives as a serde
+    /// alias, so a store written by an older build still loads. `BlockAddressWire` states the rule
+    /// directly: every short name carries every spelling the field has ever had.
+    ///
+    /// Nothing checked it. A rename that changed a field AND its alias would compile, pass every
+    /// other test, and make existing data unreadable -- with the symptom appearing at load, on
+    /// someone else's machine, later.
+    ///
+    /// The check runs one way. ADDING a durable name is free, which is what keeps this quiet while
+    /// the migration proceeds; removing or respelling one fails here with the name in the message.
+    /// If a removal is genuinely intended -- a field that no store in the world still carries --
+    /// delete it from the list below in the same change, so the decision is written down.
+    #[test]
+    fn a_durable_name_is_never_quietly_dropped() {
+        // Every serde `rename`/`alias` in the crate that carries the old vocabulary, as of the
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Groundwork for finishing the page-to-block rename: a guard asserting that all 260 durable serde names carrying the old vocabulary are still written somewhere, so a rename cannot drop a compatibility alias and make old stores unreadable. One direction only, so adding names stays free. Two survey findings recorded with it: zone became band (not slab, which is a separate concept derived from it), and slot to bucket is already done internally with routing_slot surviving only as the on-disk name.

PR #1403, opened 2026-09-10
A page is written as bytes, not as JSON

bjmeetsfo:
The framing and compression work reached the WAL (`#tsb1`, zstd) and the index log, and never
reached the page store. A page stayed JSON — and a page's `value` is a `Vec<u8>`, which JSON writes
as an array of decimal numbers, roughly four characters per byte. It was the last place a payload
was still stored as text.

### Measured on the code that ships

`what_the_page_format_costs`, 20,000 pages of one 400 B point:

| | JSON | bytes | |
|---|---:|---:|---:|
| encode | 1,128.6 ms | 4.6 ms | **99.6%** (243x) |
| decode | 1,742.8 ms | 5.9 ms | **99.7%** (297x) |
| size | 1,452 B | 423 B | 3.4x |

Per page that is **87 µs to decode against 0.29 µs**.

### It is not subsumed by compression, which is what I expected and was wrong about

The page segment on disk is zstd-framed, so the reasonable prior is that the compressor was already
removing the JSON redundancy and this buys little. Compressing the same page both ways says
otherwise:

| payload | zstd JSON | zstd bytes | saving |
|---|---:|---:|---:|
| 400 B text | 199 B | 109 B | 45.2% |
| 2 KB vector | 527 B | 297 B | 43.6% |
| 64 B text | 189 B | 96 B | 49.2% |

Still ~45% smaller *after* compression. And the CPU half was never compressible at all: bytes the
encoder does not write are bytes the compressor does not read.

### The format

One version, little-endian:

```
magic          b"TSFPB1\n"        7
point count    u32                4
per point:
    timestamp  u64                8
    length     u32                4
    value      length bytes
```

Overhead is a fixed 23 bytes rather than a multiple of the payload.

**The timestamp stays, at 8 bytes rather than the 29 it cost in JSON.** It is what picks one point
out of a page holding several. Every context write passes exactly one point, so on those paths it
is redundant with the index key — but the feature-series path does not, and addressing a point by
ordinal instead would change `BlockAddress` and every index that stores one. That is worth doing
and is not this change.

A page written before this still reads. Nothing writes that shape now, and the arm goes when those
stores do.

### Reading is deliberately paranoid

Every declared length is checked against the bytes that remain **before** anything is allocated. A
torn page declares a length it does not have, and sizing a buffer from a declared length is what
turned a corrupt tail into an aborted process once already.

`a_torn_page_is_reported_not_trusted` feeds in every prefix of a real page — each must decode to
nothing or say it is corrupt, and none may panic — and then a page that lies about its point count,
which must be refused rather than reserved for.

### One existing test changed, and why

`recovery_validates_all_timestamped_kv_page_families` asserted
`packed_timestamped_pages >= 10`. That counts **pages**, and the number of pages a fixed set of
points occupies depends on how densely a page stores them. The same 19 points — the assertion
directly above still checks 19 — now occupy 6 pages instead of 10.

Nothing was lost; the assertion was reading the encoding through a proxy. It now asserts that every
page the index points at is a packed page and that no legacy value page survived, which is what the
test is for, does not move with the format, and is strictly stronger: ten legacy pages satisfied the
old assertion and satisfy none of the new ones.

### On end-to-end numbers: there are none here, and the reason is worth recording

The first end-to-end comparison was void. The byte-page arm ingested four times faster, which gave
the background writers four times less wall clock, so it measured a store of **18,267 records
against 23,925**. A smaller store is faster to retrieve from, cheaper in CPU and lighter in RSS for
reasons that have nothing to do with the page format — and under that confound the "19% smaller
store" was actually *more* bytes per record.

The harness now waits for the store to settle and prints what it settled at. Even settled, the two
arms differ by 11% in records, and this hardware spreads 350% between identical arms, so no
end-to-end percentage is claimed. Every number above is from a bench over the operation or from
compressing the same page twice.

### DIFF
--- a/crates/temporalstore-rust/src/engine/constants.rs
+++ b/crates/temporalstore-rust/src/engine/constants.rs
@@ -5,6 +5,10 @@ use std::sync::atomic::AtomicU64;
 
 pub(super) const FEATURE_ADD_HARD_MAX_SIZE: usize = 100_000;
 pub(super) const FEATURE_PAGE_MAGIC: &[u8] = b"TSFPG1\n";
+
+/// The byte format. Distinct from the JSON magic so a page written before it still identifies
+/// itself, rather than being read as a length that happens to parse.
+pub(super) const FEATURE_PAGE_BINARY_MAGIC: &[u8] = b"TSFPB1\n";
 #[cfg(test)]
 pub(super) const TIMESTAMPED_KV_PAGE_TARGET_BYTES: usize =
     crate::storage_config::DEFAULT_CONTEXT_PAGE_TARGET_BYTES;
--- a/crates/temporalstore-rust/src/engine/packed_pages.rs
+++ b/crates/temporalstore-rust/src/engine/packed_pages.rs
@@ -7,7 +7,7 @@ use crate::block_store::{BlockAddress, BlockStoreError, LocalBlockStore};
 use crate::types::{FeaturePoint, ShardId};
 use matrixcache::{CacheKey, MultiLayerCache};
 
-use super::constants::FEATURE_PAGE_MAGIC;
+use super::constants::{FEATURE_PAGE_BINARY_MAGIC, FEATURE_PAGE_MAGIC};
 use super::state::{PackedFeaturePage, PackedFeaturePageDecode};
 use super::{append_value, read_page_bytes, read_page_bytes_cold, stable_page_object_id};
 use crate::storage_config::context_page_target_bytes;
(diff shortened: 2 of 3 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The page store was the last log still keeping payloads as text: the framing and compression work reached the WAL and index log and stopped there. Measured on shipping code over 20,000 pages: encode 1,128.6ms to 4.6ms, decode 1,742.8ms to 5.9ms, size 1,452B to 423B. Compression does not subsume it - zstd leaves the byte page 43-49% smaller than the JSON one, and the CPU half was never compressible. Reads check every declared length before allocating, and a torn page is tested at every prefix. Lib suite: 1726 passed, 2 failed, both pre-existing reds on main.

PR #1404, opened 2026-09-10
A retrieve stops copying every ref it might not need

bjmeetsfo:
A retrieve copied the ref of every candidate in the snapshot — the whole corpus — and then asked
whether anything wanted them:

```rust
let all_candidate_refs: Vec<Value> = snapshot.candidates.iter()
    .map(|candidate| candidate.selected_ref.clone()).collect();
let (by_entity, by_hash) = if current_state_query { … } else { (empty, empty) };
```

Two things wrong, fixed together. It is built unconditionally for a conditional consumer, so it now
lives inside the branch. And that consumer only *reads* the refs, so it takes borrows — which also
removes the copy from the branch that does run. The snapshot is shared behind an `Arc` and outlives
the call, so borrowing from it is free.

### What it is worth, honestly

The operation is expensive in isolation. `what_the_copies_on_the_retrieve_path_cost` now prices it:
cloning 20,000 refs costs **91.5 ms against 0.08 ms to borrow them — 4.6 µs per ref**.

**And removing it did not move the retrieve on this hardware: 15 ms either way**, across two runs
whose stores matched to 1.4% in record count. Whole-run CPU came out 10.7% lower on the proxy and
7.9% lower on the gateway, and I am not claiming those — this box spreads 350% between identical
arms, and one pair cannot carry a 10% effect.

So this ships on the mechanism rather than on a measured end-to-end win. It strictly removes work —
an unconditional O(corpus) allocation per request, for a branch that usually does not run — and
there is no configuration in which doing it is cheaper. On a corpus larger than this one it matters
more, and the price above says by how much.

### Tests

The proxy binary suite: 120 passed, 0 failed. That is the only command that reaches this file; both
of the usual whole-crate forms run zero of its tests and report ok.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -6140,12 +6140,15 @@ fn retrieve_context_pack_output(
         question_type,
         "current_state" | "latest" | "profile_memory"
     );
-    let all_candidate_refs: Vec<Value> = snapshot
-        .candidates
-        .iter()
-        .map(|candidate| candidate.selected_ref.clone())
-        .collect();
+    // Built inside the branch, and borrowed rather than copied: the consumer only reads them, and
+    // only sometimes. This used to clone the `selected_ref` of every candidate in the snapshot --
+    // the whole corpus -- before asking whether the one caller wanted them.
     let (profile_by_entity, profile_by_source_entity_hash) = if current_state_query {
+        let all_candidate_refs: Vec<&Value> = snapshot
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. A retrieve cloned the ref of every candidate in the snapshot before asking whether the one consumer wanted them, and that consumer runs only on a current-state query. Priced on the bench at 4.6us per ref, 91.5ms per 20,000. Stated plainly in the PR: removing it did not move the retrieve on this box, so it ships on the mechanism rather than a measured win. 120 proxy-bin tests pass.

PR #1405, opened 2026-09-10
The rebuild says which stage it spent, and the parse cache stops starving

bjmeetsfo:
A retrieve reported the snapshot rebuild as one number, so a rebuild that took 700 ms could not say
which part took it. It now reports three: reading and parsing the shards, counting the inventory,
and turning records into candidates. Each is fixed by a different change.

The first thing it said was that the cache added two changes ago was **starving**.

### The split, and the bug it exposed

Cold rebuild over 26,756 records:

| stage | ms | share |
|---|---:|---:|
| read + parse shards | 367.5–540.5 | **55–75%** |
| candidates | 92.9–173.7 | 14–26% |
| inventory | 36.6 | 5% |

The read stage should have been near zero on a rebuild after a write: records are appended, so
every shard but the newest parses to what it parsed before, and that is exactly what the parse
cache exists for.

It was not, because of how I budgeted it. Each parse was charged at **six times** its payload bytes
against **half** the payload budget — so the parse cache could hold about a *twelfth* of the payload
volume and evicted nearly everything it was given. The rebuild kept re-parsing, and the cache
looked like it had not helped.

### Measured, nothing else changed

Same store, 27,186 records:

| | payload/2 budget, 6x charge | equal budget, 4x charge |
|---|---:|---:|
| rebuild **read** stage, after a write | 367–540 ms | **0.1 ms** |
| rebuild total, after a write | 662–720 ms | **153 ms** |
| retrieve after a write | 744 ms | **221 ms** |
| warm retrieve p50 | 78 ms | 56.5 ms |
| proxy RSS | 551 MB | 585 MB (**+6%**) |

The first rebuild on a cold process still pays the read — 618.9 ms with 501.7 ms of reading — and
always will; there is nothing cached yet. It is the rebuild *after a write* that this fixes, and
that is the one an ingesting store does on nearly every retrieve.

### Why the two budgets stay separate

Evicting a parse costs a re-parse. Evicting a payload costs a re-read *and* a re-parse. They are
not worth keeping on the same terms, so they keep separate budgets and separate LRUs rather than
sharing one pool — and the payload cache's behaviour is unchanged by this.

The charge is still a deliberate over-estimate, because the cost of guessing low is an unbudgeted
cache, which is the failure the whole type exists to prevent.

### Tests

`a_parse_that_fits_is_kept` asserts the *relationship*, not the constants: the decode budget may not
be smaller than the payload budget, a single small parse must survive, and a parse may not be
charged less than its payload. An edit that halves the budget again fails there rather than quietly
making a retrieve slow — which is precisely what happened here and what nothing caught.

The proxy binary suite: 121 passed, 0 failed.

The report line also lost the runs of spaces it was carrying: it had been written with line
continuations that ended up as literal whitespace in the output.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -293,6 +293,19 @@ struct CachedRetrieveCandidate {
     vector: Option<Vec<f32>>,
 }
 
+/// What building a snapshot cost, by stage, in milliseconds.
+///
+/// A retrieve after a write is nearly all rebuild, and the rebuild is three unrelated pieces of
+/// work: reading and parsing the shards, counting the inventory, and turning records into
+/// candidates. Each is fixed by a different change, so a total is not enough to act on. A snapshot
+/// served from cache reports the build it came from.
+#[derive(Clone, Copy, Debug, Default)]
+struct RetrieveBuildCost {
+    read_ms: f64,
+    inventory_ms: f64,
+    candidates_ms: f64,
+}
+
 #[derive(Clone, Debug)]
 struct RetrieveCandidateSnapshot {
     candidates: Vec<CachedRetrieveCandidate>,
@@ -300,6 +313,7 @@ struct RetrieveCandidateSnapshot {
     scanned_records: usize,
     placement_partitions_touched: usize,
     index_postings_read: usize,
+    build: RetrieveBuildCost,
 }
 
 struct NativeScoredCandidate {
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The retrieve now reports the rebuild in three stages, and the first thing it showed was that the parse cache I added two changes ago was starving: charged at 6x payload bytes against half the payload budget, so it held about a twelfth of the volume. Fixed. On a 27,186-record store: rebuild read stage 367-540ms to 0.1ms, rebuild after a write 662-720ms to 153ms, retrieve 744ms to 221ms, for 6% more proxy RSS. 121 proxy-bin tests pass.

PR #1407, opened 2026-09-10
A shard's candidates are built once per write, not once per retrieve

bjmeetsfo:
With the parse cache fixed, the rebuild after a write was read 0.1 / inventory 35.6 /
**candidates 91.1 ms** — the candidate build was all of it. And it had the same shape the parse
did: records are appended, so every shard but the newest produces exactly the candidates it
produced before, filter for filter and ref for ref.

So a shard's candidates are cached beside its payloads and its parse, in the entry the write path
already patches and removes — which is the whole correctness argument, the same one that made the
parse cache safe.

### Measured, in one run, on one store

| | ms |
|---|---:|
| candidate stage, building a shard's candidates | 61.9 |
| candidate stage, taking them from the cache | **2.4 – 4.4** |

Those are two rebuilds in the same run against the same 18,054-record store, so nothing about
store size or machine drift is in the comparison. The retrieve after a write came out at 50 ms and
121 ms on the two rebuilds that followed, against 221 ms before this.

### Cached per scope, and why that is enough

The filter is scope-dependent, so the cache is keyed by scope signature. One caller means one key.
A second scope pays its own build once per shard per write rather than on every retrieve — which is
the same bargain the parse cache makes, one level up.

### Segments, not one list

The snapshot holds the per-shard lists as `Arc` segments rather than concatenating them. A
candidate costs about **4.6 µs to clone** — the ref bench measured that — so copying ~18,000 of
them into one vector would hand back most of what caching them saves.

That changes what an ordinal means, and the ordinal is load-bearing: the scoring pass records it
and the budget closures resolve it later. If those two disagree a retrieve scores one candidate and
bills the budget for another, and **nothing fails** — it just serves the wrong refs.
`an_ordinal_means_the_same_candidate_whatever_the_segments` walks every ordinal against the flat
order over seven segment shapes, including empty segments at the front, in the middle and at the
end, because a shard whose records are all filtered out contributes exactly that.

### Secondary groups take the old path, deliberately

Those are matched against index terms gathered across **every** shard, so a shard cannot be
filtered on its own. A caller that asks for them gets the whole-corpus build the loader always did,
wrapped in a single segment so everything downstream sees one shape. `candidate_from_record` is
shared by both paths so they cannot drift on what a candidate *is* — only on which records they are
given.

### What is left

The rebuild is now read 0.1 / **inventory 22.8–26.9** / candidates 2.4. The inventory is the top
stage, and it is mergeable the same way: its counters are per-record and scope-independent, so
per-shard counts could be cached and summed. That is the next one.

### Tests

The proxy binary suite: 122 passed, 0 failed.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -308,12 +308,67 @@ struct RetrieveBuildCost {
 
 #[derive(Clone, Debug)]
 struct RetrieveCandidateSnapshot {
-    candidates: Vec<CachedRetrieveCandidate>,
+    /// The candidates, as one list per shard rather than one list.
+    ///
+    /// A shard's candidates are cached and shared, and a candidate costs about 4.6 us to clone --
+    /// concatenating them here would hand back most of what caching them saves. `first_of_segment`
+    /// holds each segment's starting ordinal so a candidate can still be addressed by one number,
+    /// which is what the scoring pass and the budget closures use.
+    segments: Vec<Arc<Vec<CachedRetrieveCandidate>>>,
+    first_of_segment: Vec<usize>,
     memory_inventory: Value,
     scanned_records: usize,
     placement_partitions_touched: usize,
     index_postings_read: usize,
     build: RetrieveBuildCost,
+}
+
+impl RetrieveCandidateSnapshot {
+    fn from_segments(
+        segments: Vec<Arc<Vec<CachedRetrieveCandidate>>>,
+        memory_inventory: Value,
+        scanned_records: usize,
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. A shard now caches its CANDIDATES beside its payloads and its parse, invalidated by the same write-path patch. Measured in one run on one 18,054-record store: the candidate stage is 61.9ms when it builds them and 2.4-4.4ms from cache. The snapshot holds per-shard Arc segments rather than concatenating, since a candidate costs ~4.6us to clone - which changes what an ordinal means, and an ordinal disagreement would serve the wrong refs without failing, so a test walks every ordinal over seven segment shapes. 122 proxy-bin tests pass.

PR #1408, opened 2026-09-10
A shard counts its inventory once per write, and the counts are summed

bjmeetsfo:
With the parse and the candidates cached per shard, the rebuild was read 0.1 / **inventory
22.8–26.9** / candidates 2.4 — the inventory was what was left, and it had the same shape
everything else did: a pass over every record on every rebuild, producing what it produced last
time for every shard but the newest.

It is cacheable because **the counters do not depend on the query scope.** Only three things do,
and all three are derived at the end from the counts: the `query_scope` block, `available_layers`,
and the `has_*` flags. So counting is per shard and scope-free; finishing is per request.

### The split

`native_retrieval_memory_inventory` becomes `empty_inventory_counts` + `count_records_into` +
`finish_inventory`, and still exists with the same signature and the same output for its other
caller. The loader sums per-shard counts and finishes once.

Unlike the candidates there is no scope key on this cache — the scope never reaches the counting.

### Measured, in one run on one store

| stage | building | from cache |
|---|---:|---:|
| inventory | 35.9 ms | **0.1 – 1.2 ms** |

The rebuild as a whole, on the same 22,628-record store: **read 18.0 / inventory 1.2 /
candidates 3.2 = 22.4 ms**, with the retrieve at 39 ms.

Across this run of work, the rebuild after a write on a store of this size:

| | rebuild |
|---|---:|
| before any of it | 1,134.9 ms |
| after the parse budget fix | 153 ms |
| after the candidate cache | 42–50 ms |
| after this | **22.4 ms** |

### One thing worth knowing, and it is a limitation

The candidate cache is keyed by scope signature. In this run one rebuild still shows
`candidates 70.6` — a signature it had not prepared before. If a deployment's scope varies per
request (a session id inside it, say), the candidate cache will miss on every retrieve while the
parse and the inventory still hit. That is worth measuring on a real workload before assuming the
candidate half helps there; the inventory half is unaffected, because it has no scope key at all.

### Tests

`summing_shard_counts_equals_counting_them_together` is the one that matters. A miscounted
inventory does not fail — it changes which memory layers a retrieve believes exist, since
`available_layers` and the `has_*` flags are derived from these numbers. It asserts summed counts
against counting everything at once, over records landing in all three buckets, including a shard
that contributes nothing (a shard of pure index records does exactly that), and then asserts the
*finished* inventories match too, which is what a retrieve actually reads.

`merge_inventory_counts` walks the fixed buckets in step rather than trying to be general: anything
that is not an integer under a known bucket is a shape change, and quietly ignoring it would hide
it.

The proxy binary suite: 123 passed, 0 failed.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -1462,6 +1462,11 @@ struct SnapshotEntry {
     /// Keyed by scope because the filter is: one caller means one key, and a second scope pays its
     /// own build once per shard per write rather than on every retrieve.
     prepared: BTreeMap<String, Arc<Vec<CachedRetrieveCandidate>>>,
+    /// The shard's inventory counts, or `None` until something asks.
+    ///
+    /// No scope key, unlike `prepared`: the counters are scope-free, and the scope only enters
+    /// when counts are finished into an inventory, once per rebuild rather than once per shard.
+    counts: Option<Value>,
     /// The payload bytes. `decoded` is charged separately, so this stays comparable to `weigh`.
     bytes: usize,
     /// What the decoded records and the prepared candidates are charged at, or 0 for none.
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Last of the per-shard caching: the inventory counters are scope-free, so a shard counts once per write and the counts are summed, with the scope entering only at the finishing step. Inventory stage 35.9ms to 0.1-1.2ms; the rebuild is now read 18.0 / inventory 1.2 / candidates 3.2 = 22.4ms, against 1,134.9ms when this run of work started. One limitation stated in the PR: the candidate half is keyed by scope signature, so a deployment whose scope varies per request will miss there while parse and inventory still hit. 123 proxy-bin tests pass.

PR #1410, opened 2026-09-10
The engine embeds its own query, from one encoder configuration

bjmeetsfo:
The engine has been able to call a real OpenAI-compatible embedding endpoint all along — it is what
the embed drainer and the backfill binary do. But only a caller holding a
`ContextModelProviderConfig` could reach it, and the proxy had none, so the query vector had to be
computed on the Python side and shipped over.

This gives the engine one, and points both paths at one configuration.

### The retrieve embeds its own query

When the caller sends no vector and the query is not empty, the retrieve embeds it through the same
provider path extraction uses. Three things it deliberately does not do:

- **It does not override a vector the caller sent.** A caller that embedded the query itself — with
  its own model, or from its own cache — keeps its vector. Which model ranked the store stays the
  caller's decision.
- **It does not fail a retrieve when the encoder is unreachable.** No vector means lexical ranking,
  which is what a retrieve without one already did. An encoder outage costs dense ranking for the
  duration, not retrieval.
- **It does not use a deterministic vector.** With no endpoint the provider is deterministic, and a
  deterministic vector is not a ranking signal — it scores every candidate against noise, which is
  worse than ranking lexically *and* looks like dense ranking is working.

`MATRIXARK_PROXY_EMBED_QUERY=0` turns it off.

### One encoder, one configuration

The same endpoint was configured by two different variables: the Python side and the proxy read
`MATRIXARK_EMBEDDING_API_BASE`, while the embed drainer inside the data node read
`MATRIXARK_EMBED_BASE_URL` and built its provider inline. **A deployment that set one and not the
other got real vectors on one path and deterministic ones on the other, with nothing saying so.**

`context_provider_from_env` is now the single constructor, honours both names (preferring the one
the rest of the deployment uses), and the data node stops building its own. The old name is kept
rather than dropped: a running deployment sets it, and removing it would silently move that
deployment onto deterministic vectors — the exact failure this consolidates to prevent.

### What this does NOT fix, stated plainly

I tried to exercise the new path end to end by pointing the gateway at a dead encoder so it would
send no vector. It still sent one, and the retrieve logged:

```
snapshot holds 194 candidates, 194 with vectors up to 32 dims; query embed 0.0 ms
```

**32 dims.** With its encoder unreachable the Python side does not fail — it falls back to a 32-dim
hash vector, for the stored vectors *and* the query. So the engine never sees a missing vector, the
new path does not trigger, and ranking runs on hash noise while looking healthy.

That is a pre-existing defect and this change does not fix it. What it does is make the engine able
to embed on its own, which is what fixing it will need.

### Tests

`an_unconfigured_encoder_yields_no_query_vector` pins the deterministic-vector refusal — the one
that would silently degrade ranking while appearing to work.
`an_unreachable_encoder_yields_no_query_vector` pins that an outage costs a vector, not a retrieve.
`an_empty_query_is_not_embedded` covers the trivial case. All three take the file's existing
`env_guard`, which recovers from poisoning rather than turning one failure into a cascade.

The proxy binary suite: 126 passed, 0 failed.

### DIFF
--- a/crates/temporalstore-rust/src/bin/server.rs
+++ b/crates/temporalstore-rust/src/bin/server.rs
@@ -10,7 +10,8 @@ use serde::{Deserialize, Serialize};
 use temporalstore_rust::env_flag::env_bool;
 use temporalstore_rust::context_workflow::{
     context_pipeline_manage_report, context_workflow_state_report, default_context_model_providers,
-    embed_drainer_config_from_env, embed_drainer_enabled, extract_context, ingest_extract_context,
+    context_provider_from_env, embed_drainer_config_from_env, embed_drainer_enabled,
+    extract_context, ingest_extract_context,
     inject_context, retrieve_context, run_embed_drainer_loop, ContextExtractRequest,
     ContextIngestExtractRequest, ContextInjectRequest, ContextModelProviderConfig,
     ContextRetrieveRequest,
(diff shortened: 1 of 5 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The engine now embeds its own query when the caller sends none, through the provider path extraction already used, and never overrides a caller vector, never fails a retrieve on an encoder outage, and never uses a deterministic vector. Also consolidates a real config split: one endpoint had two variables (MATRIXARK_EMBEDDING_API_BASE vs MATRIXARK_EMBED_BASE_URL), so a deployment setting one got real vectors on one path and deterministic ones on the other. Stated in the PR: this does NOT fix the 32-dim hash fallback - pointing the gateway at a dead encoder still produces a 32-dim hash vector rather than none, so ranking runs on noise while looking healthy. 126 proxy-bin tests pass.

PR #1411, opened 2026-09-10
The proxy's serving caches move into the folder the rest of it uses

bjmeetsfo:
`matrixark_rust_proxy_impl.rs` was 11,993 lines. Four pieces have already been split out of it into
`matrixark_rust_proxy_impl/` -- scope scoring, budget parsing, native serving, the pack -- and the
serving caches are the next coherent piece.

**460 lines moved**, leaving the parent at 11,535.

### What moved, and why these together

One shard snapshot cache holding payloads, their parses and their prepared candidates; the
record-count cache; the scan cache; and the candidate-snapshot cache.

They belong together because **they are invalidated together**. The write path patches or drops a
shard's entry and everything derived from it goes with it, which is what stops a derivative
outliving what it describes -- the property three separate changes now depend on.

### It is a move, and only a move

Textual `include!`, exactly like the four already there, so scope stays flat and nothing needs
re-importing. The script that did it verified both directions: every item it names is present in
the new file and no longer defined in the parent, and it refuses to run if the block it cut does
not contain all of them or runs past its end.

126 proxy-bin tests pass, unchanged.


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Pure move: 460 lines of serving caches out of the 11,993-line proxy impl into matrixark_rust_proxy_impl/, alongside the four pieces already split out there. They go together because they are invalidated together - the write path drops a shard entry and everything derived from it goes with it. Verified both directions by the move script, 126 proxy-bin tests unchanged.

PR #1414, opened 2026-09-10
The engine drops the refs another ref already carries

bjmeetsfo:
The engine selects the refs and Python then read every one of them again to decide which are
carried twice. Sampling the gateway had put that filter at 76.8% of its on-CPU time; two changes
took it from 98.4 ms to 8.4 ms on a 1,000-ref pack, and it was still the largest single item there.

It is also the last substantial piece of *retrieval* that was still done outside the engine, so it
moves.

### What it does

After the refs are selected, the engine drops any whose text a **longer** ref already carries,
comparing the entity's value half — the part after `=` — against the other's whole text. Same rule,
same threshold, same reasoning as the filter it replaces:

- only a **strictly longer** item can contain another, which is what makes the scan affordable;
- containment is **transitive** and carries the length condition, so "contained in some longer item
  that is not itself redundant" and "contained in some longer item" select the same items — which
  is what frees the visit order;
- text under eight characters is never redundant, because a label or a stub matches inside almost
  anything.

The pack says `redundant_items_dropped`, and the one caller that builds groups reads it and skips
its own copy. Default false everywhere, so a pack from an engine that did not sweep is still swept
by the caller — a rolling upgrade works in either direction.

`MATRIXARK_PACK_DROP_REDUNDANT_ITEMS=0` still turns it off, the same variable as before.

### Measured, and it is not what I expected

| | |
|---|---:|
| the sweep, in Python | 8.4 ms |
| the sweep, in the engine | **8.7 ms** |
| refs dropped from the pack | 190 of 1,029 |

**Rust is not faster at this.** Both spend their time in optimised C-level substring scanning, so
the work is memory-bandwidth-bound and the language does not enter into it. Anyone expecting a
port to be a speedup here should see this number first.

What the move actually buys:

- **The pack crossing the wire is 839 refs instead of 1,029** — about 16% less to serialise in the
  proxy, carry, and decode in the gateway. That saving is real and was not available while the
  filter ran after the transfer.
- The gateway stops doing it, which is the point of the direction this belongs to.

What it does not buy is CPU: the work is the same work, done in a different process. The retrieve
wall time did not move outside noise, and I am not claiming it did.

### Tests

The cases the Python suite pins are ported rather than rewritten, so both implementations answer to
the same set: an entity projection of a kept event is dropped, an entity that adds content
survives, a short fragment is never redundant, equal-length duplicates are both kept, nothing
redundant leaves the refs untouched, and a chain of containments leaves only the longest **whatever
order it arrives in** — which is the transitivity argument, asserted directly.

`the_sweep_agrees_with_the_plain_definition` is the one that catches what a table does not: it
compares against "contained in some longer item" over 60 generated packs. The length ordering and
the joined haystacks are both optimisations, and neither may change which refs survive.

133 proxy-bin tests and 10 Python pack tests pass.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -6033,6 +6033,19 @@ fn retrieve_context_pack_output(
         }
         selected_refs.push(selected_ref.clone());
     }
+    // Dropped here, where the refs are chosen, rather than by the caller reading them all again:
+    // this was the last substantial piece of retrieval still done in Python, and the profiler put
+    // it at about a third of the gateway's CPU.
+    let swept_redundant_items = pack_drop_redundant_items();
+    let sweep_started = Instant::now();
+    let before_sweep = selected_refs.len();
+    let selected_refs = if swept_redundant_items {
+        drop_redundant_selected_refs(selected_refs)
+    } else {
+        selected_refs
+    };
+    let sweep_ms = sweep_started.elapsed().as_secs_f64() * 1000.0;
+    let swept_away = before_sweep.saturating_sub(selected_refs.len());
     let selected_count = selected_refs.len();
     let mut memory_inventory = snapshot.memory_inventory.clone();
     let selected_profile_ref_count = selected_refs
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The redundancy sweep moves from Python into the engine, which is the last substantial piece of retrieval that ran outside it. Stated plainly in the PR: Rust is NOT faster at this - 8.4ms in Python, 8.7ms in the engine, because both are doing C-level substring scanning and the work is memory-bandwidth-bound. What it buys is the pack crossing the wire at 839 refs instead of 1,029, about 16% less to serialise, carry and decode. The Python cases are ported rather than rewritten so both answer to the same set, plus a property test against the plain definition over 60 generated packs. 133 proxy-bin and 10 Python tests pass.

PR #1416, opened 2026-09-10
The redundancy sweep finds containers through a trigram index

bjmeetsfo:
Timing the sweep after it moved into the engine showed it costing **37–43 ms** on a pack of a few
hundred kilobytes — larger than the entire snapshot rebuild, and the single biggest item in a
retrieve. It scanned every longer text for every item, which is O(n × total) however the text is
arranged; the joined haystack removed the per-call overhead but not the work.

It now finds containers through a trigram index: **43 ms → 5.4 ms**, same rule, same 160 refs
dropped.

### How

If X is a substring of Y then every trigram of X is in Y — including X's **rarest**. So the sweep
indexes trigram → texts and considers only the texts on that one posting list. A distinctive
trigram has a short list, and that is the entire saving.

The index only chooses which pairs to look at. Every pair it offers is still verified with a real
substring check, so the answer is what comparing everything against everything would give.

Postings are built longest-first, so a text's candidates are exactly the postings recorded before
it — already ordered, with no length comparison beyond skipping equal lengths.

### Two filters that did not work, and why they are worth recording

**A word mask was not even valid.** If X is inside Y, X's *words* need not be Y's words:
`my favorite drink is matcha and i live in kyoto` is plainly inside
`user: my favorite drink is matcha and i live in kyoto, noted at step 4`, yet `kyoto` is not one of
the longer text's words — it has a comma attached. A word of the shorter text can sit inside a
longer word or span a boundary. **The ported chain test caught this**, which is the argument for
porting the Python suite's cases rather than writing fresh ones.

**A 64-bit bigram mask was valid and useless.** A 400-byte sentence has about 400 bigrams landing
in 64 buckets, so nearly every bucket is set in nearly every text and no pair was ever rejected. It
measured 43 ms — unchanged.

Trigrams are distinctive enough to actually divide the pack, which is why the third attempt is the
one that worked.

### Measured

| | |
|---|---:|
| sweep | 43 ms → **5.4 ms** |
| proxy retrieve | 47 ms → **13 ms** |
| warm retrieve, end to end | 69.5 ms → **42.5 ms** |
| refs dropped | 160, unchanged |

### Tests

Unchanged and all still passing — which is the point, since this is a pure change of *how* the same
refs are found. `the_sweep_agrees_with_the_plain_definition` compares against "contained in some
longer item" over 60 generated packs, and the ported cases from the Python suite pin the behaviour
both implementations answer to.

133 proxy-bin tests pass.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -304,6 +304,10 @@ struct RetrieveBuildCost {
     read_ms: f64,
     inventory_ms: f64,
     candidates_ms: f64,
+    /// Shards whose candidates were reused, and shards in total. Records are appended, so a write
+    /// should leave every shard but the newest reusable -- this says whether it does.
+    prepared_hits: usize,
+    prepared_shards: usize,
 }
 
 #[derive(Clone, Debug)]
@@ -5511,9 +5515,11 @@ fn prepare_shard_candidates(
     key: String,
     signature: &str,
     scope: Option<&Value>,
+    reused: &mut usize,
 ) -> Result<Arc<Vec<CachedRetrieveCandidate>>, String> {
     if let Ok(mut cache) = hgetall_snapshot_cache().lock() {
         if let Some(prepared) = cache.prepared_for_scope(&key, signature) {
+            *reused += 1;
             return Ok(prepared);
         }
     }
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The sweep was costing 37-43ms in the engine, larger than the whole snapshot rebuild. It now finds containers through a trigram index - if X is inside Y then every trigram of X is in Y, including its rarest - so only that posting list is considered, and every pair is still verified properly. 43ms to 5.4ms, same 160 refs dropped, proxy retrieve 47ms to 13ms, warm retrieve 69.5ms to 42.5ms. Two earlier filters failed and are recorded in the PR: a word mask was not a valid necessary condition (the ported chain test caught it), and a 64-bit bigram mask saturates on prose so it rejected nothing. 133 tests pass.

PR #1417, opened 2026-09-10
The engine sends the serving shape, instead of sending more for the caller to discard

bjmeetsfo:
The engine shipped every selected ref with about twenty fields. The caller then rebuilt each one
keeping an allow-list — and of the engine's fields only seven survive it: `ref_type`, `text`,
`entity_type`, `entity_name`, `memory_scope`, `session_continuity` and `memory_layer`. **The other
thirteen were serialised, carried across the wire, decoded, and dropped.**

Sampling the gateway put that rebuild at 37% of its on-CPU time — the largest single item there
once the redundancy sweep moved — and the decode of what it discards at another 8%.

So the engine emits the serving shape, and the caller skips its own pass.

### The caller asks; the engine does not assume

The caller knows one thing the engine does not: whether this retrieve wants **debug** refs, which
carry lineage fields the serving shape drops. An engine compacting unconditionally would quietly
strip them.

So the request carries `serving_refs_compact`, set to `not debug_refs`, and the engine compacts
only when asked. Absent, the answer is no — an older caller keeps the shape it expects, and a
newer caller against an older engine gets the full shape and compacts it itself, as it does today.
`MATRIXARK_ENGINE_COMPACT_SERVING_REFS=0` is a deployment-level veto that needs no caller change.

### Verified by diffing two full packs, not by reasoning

One store, one query, the same proxy binary restarted with the flag off and then on, and the two
gateway responses compared field by field: **441 items each, identical** apart from the pack id,
which is a hash of the wall clock.

Two things that check took to be worth anything, both of which caught real faults first:

- **It refuses to compare empty packs.** The first attempt compared two 485-byte responses and
  reported them identical. Two empty packs are identical for reasons that have nothing to do with
  the change; a check that passes on nothing reads as proof.
- **The restarted proxy carries its full profile.** With only the proxy address set it came up
  without the `TS_*` profile, opened its own empty store, and served those empty packs. That is the
  wrong-backend failure, and the non-empty gate is what surfaced it.

### Measured

| | |
|---|---:|
| warm retrieve, end to end | 42.5 ms → **37.2 ms** |
| served response | 247 KB → **167 KB** (−33%) |
| proxy retrieve | 12 ms |

The payload figure is the more interesting one: it is 13 fewer fields on every ref, so it is
serialisation the proxy no longer does, bytes the socket no longer carries, and JSON the gateway no
longer decodes.

133 proxy-bin tests and 10 Python pack tests pass.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -6130,7 +6130,13 @@ fn retrieve_context_pack_output(
         );
     }
     let correctness = selected_count > 0;
-    let serving_selected_refs = native_serving_refs(&selected_refs);
+    // The caller asks for the serving shape; it knows whether this retrieve wants debug refs and
+    // the engine does not. Absent, the answer is no, so an older caller keeps the shape it expects.
+    let compact_serving = request_record
+        .get("serving_refs_compact")
+        .and_then(Value::as_bool)
+        .unwrap_or(false);
+    let serving_selected_refs = native_serving_refs(&selected_refs, compact_serving);
     let serving_dropped_refs = native_serving_dropped_refs(json!({
         "refs": dropped_ref_details,
         "native_summary": true,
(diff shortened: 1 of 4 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The engine now sends the serving shape rather than twenty fields for the caller to whittle to seven; thirteen fields per ref were being serialised, carried, decoded and dropped. The caller asks for it (it knows about debug refs, the engine does not), so a mixed-version deployment works either way. Verified by diffing two full packs over one store - 441 items, identical - and the check refuses to compare empty packs, which is how it caught that the restarted proxy had no TS_ profile and was serving an empty store. Warm retrieve 42.5ms to 37.2ms, served response 247KB to 167KB. 133 proxy-bin and 10 Python tests pass.

PR #1418, opened 2026-09-10
A rebuild says how many shards it reused, and for which scope

bjmeetsfo:
A rebuild reports how many shards it reused, and which scope it prepared for. Those two numbers
answer a question the log could not: whether the per-shard candidate cache is working, or silently
never hitting.

They are already earning their keep. **They found that a retrieve carrying secondary index groups
costs 930.9 ms against 87.0 ms without them**, and showed why — `shards reused 0/6` against `4/7`.

### What the numbers say

```
shards reused 10/11 for scope 276600     <- the cache working: only the newest shard rebuilt
shards reused  0/12 for scope 0          <- the whole-corpus path, no reuse possible
```

The scope is a short digest rather than the scope itself, because whether it MOVED is the whole
question and a scope carries tenant and user identity that has no business in a log. A stable
digest with rising reuse is a healthy cache; a digest that changes per request is a cache that
cannot hit however well it works, and the fix for that is upstream in what the scope carries.

`scope 0` marks the path that does not prepare per shard at all: secondary index groups are matched
against index terms gathered across **every** shard, so a shard cannot be filtered on its own.

### A divergence these numbers exposed, which this change does not resolve

Measured on one store, one query, with and without the groups:

| | with groups | without |
|---|---:|---:|
| retrieve | 930.9 ms | **87.0 ms** |
| items in the pack | 448 | 757 |
| shards reused | 0/6 | 4/7 |

**All 448 items appear in the 757.** The groups purely remove items — 309 of them, 41% of the pack.
So the engine uses them as an **admission filter**.

The caller's own note beside where it computes them says the opposite:

> *"The secondary-index groups are deliberately LEFT ALONE: on this path they are not an admission
> filter, they add 0.08 to a node's score, so clearing them changed nothing on two fixtures."*

Both can be true — that comment is about the caller's own scan path, not the engine's — but the
result is that a retrieve drops 41% of its candidates through a mechanism one side documents as a
score nudge, and pays 10x latency for it. Whether that filtering is wanted is a retrieval-quality
decision, so this change reports it rather than altering it.

### Cost

The digest is computed once per rebuild and the counter is an increment per shard. Both are read
only when the line is printed.

133 proxy-bin tests pass.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -308,6 +308,9 @@ struct RetrieveBuildCost {
     /// should leave every shard but the newest reusable -- this says whether it does.
     prepared_hits: usize,
     prepared_shards: usize,
+    /// Which scope this snapshot's candidates were prepared for. If it moves between rebuilds, the
+    /// per-shard candidate cache cannot hit however well it works.
+    scope_digest: u64,
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, and this one has a finding worth a look. A rebuild now reports shard reuse and a scope digest, which showed the per-shard candidate cache working (10/11 reused) on the normal path and bypassed entirely when a retrieve carries secondary index groups. Measured on one store: with groups the retrieve is 930.9ms and the pack has 448 items; without, 87.0ms and 757 items, with all 448 among the 757. So the engine uses those groups as an admission filter dropping 41% of the pack, while the note beside where the caller computes them says that on its own path they are a 0.08 score nudge rather than a filter. That is a retrieval-quality call, so the PR reports it rather than changing it.

PR #1420, opened 2026-09-10
The serving pack counted each field twice

bjmeetsfo:
Two passes over the serving pack that could not change its answer.

**Each field was counted twice.** `compact_context_pack_for_serving` asked
`default_session_continuity_for_pack` and `_default_memory_layer_for_pack` for the two defaults,
and each of those counts the whole pack — then the function counted the whole pack again for the
`counts` map it serves. Four walks where two do, and `_memory_layer_for_ref`, the one with the
fallback chain, ran three times per ref rather than twice. The defaults *are* those counts read one
way, so they are now taken from them.

That also leaves **one** copy of the tie rule. It was written out twice, once per default, and it
is not a detail a copy can drift on quietly: the winner of a tie decides which value is left off
every item in the pack.

**The alias loop read an empty dict sixteen times per ref.** `ref.get(field, metadata.get(field))`
evaluates the default first, always — and on the serving path no ref carries metadata at all
(0 of 536 on the measured pack), so that is 8,576 lookups per retrieve building defaults nothing
uses. The fallback is kept exactly as it was: metadata answers only when the ref does not carry the
field, so a ref holding an explicit null still resolves to null rather than being overridden.

### What it is worth

The one-box cannot resolve a sub-millisecond effect end to end — every retrieve writes, so the
store is larger for each arm than the one before it — so the function is timed directly, on the
536 refs a real retrieve produced, with the file swapped between arms and nothing else changed:

| | round 1 | round 2 | mean |
|---|---:|---:|---:|
| before | 3.25 ms | 3.29 ms | 3.27 |
| **after** | **2.80 ms** | **2.83 ms** | **2.815** |

**-13.9%, 0.46 ms per retrieve**, and the two served packs are byte-identical over all 536 items.

Of that, the duplicate counting alone is 0.38 ms of work removed — both counts over 536 refs
measure 0.38 ms, and one of the two rounds was pure repetition.

### What holds it

Two tests. One pins the metadata fallback in the case that separates the two forms — a ref that
carries the field with a null value — because the obvious "simplification" to
`ref.get(field) or metadata.get(field)` would silently let metadata win there. The other pins the
tie rule and that the served defaults and counts agree.

Pipeline suite: 52 failing names with the change and 52 without, no name moved either way. Backend
policy suite green.

### DIFF
--- a/tools/matrixark_mcp_context_pack.py
+++ b/tools/matrixark_mcp_context_pack.py
@@ -1117,11 +1117,20 @@ def _memory_layer_counts(refs: list[Json]) -> Json:
     return counts
 
 
-def _default_memory_layer_for_pack(refs: list[Json]) -> str:
-    counts = _memory_layer_counts(refs)
+def _most_common_counted_name(counts: Json) -> str:
+    """The name a pack leaves off its items: most common, ties broken by the larger name.
+
+    One copy of the tie rule. It used to be written out twice, once per default, and it is not a
+    detail either copy could drift on quietly: a different winner does not rename a default, it
+    moves the field onto or off every item in the pack.
+    """
     if not counts:
         return ""
     return max(counts.items(), key=lambda item: (item[1], item[0]))[0]
+
+
+def _default_memory_layer_for_pack(refs: list[Json]) -> str:
+    return _most_common_counted_name(_memory_layer_counts(refs))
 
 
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The serving compaction counted each pack field twice and read an empty metadata dict sixteen times per ref; both are removed, served pack byte-identical over 536 items, the pass 13.9% faster.

PR #1424, opened 2026-09-10
The snapshot kept a vector nothing was going to read

bjmeetsfo:
The snapshot kept a 4 KB vector on every cached candidate, including on the deployments that never read one.

Each shard's candidates are cached for the life of the entry, and each candidate pinned its record's
embedding — 1,024 floats — so that dense scoring would not have to re-read the records. That is the
right trade when ranking is dense. It is dead weight when it is not, and **lexical is the default**:
`MATRIXARK_NATIVE_DENSE_RANKING` is unset on a stock deployment, so no query vector exists and the
field is read by nothing.

The vector is now retained only when the retrieve actually has a query vector to score against.

### Getting the condition right took two attempts

The obvious gate — "did the caller send a vector" — is wrong. When the caller sends none the engine
**embeds the query itself**, so `ranking_uses_vectors` can be true on a request whose caller sent
nothing. The gate has to be the *resolved* vector, which meant resolving it before the snapshot is
built rather than after. The query text is bound earlier to make that possible; nothing in either
step reads the snapshot.

### It travels in the cache key, in both caches

A snapshot built without vectors must never be handed to a request that ranks densely — that would
silently rank it without them, which is the one failure this cannot be allowed to have. So the flag
is part of the snapshot cache key **and** of the per-shard signature:

```
{storage_prefix}:candidate_snapshot:{count}:{scope}:{secondary}:{v|-}
{scope signature}|{v|-}
```

A mixed deployment therefore keeps two candidate lists per shard, which is the correct price; a
deployment that only ever does one keeps one.

### What it is worth

On this box, not measurable — and I would rather say so than quote a number I do not have. The test
store is ~1,800 records, so ~1,500 candidates × 4 KB is about 6 MB against a proxy whose RSS swings
80–100 MB on a 20-second timescale with no load at all. Two arms built from the same tree differing
by this one constant came back 142 MB against 153 MB median over 20 samples each — inside the swing,
and in the wrong direction, which is what noise looks like.

The mechanism is what justifies it: memory held per candidate for the life of a shard entry, read by
nothing on the default path. It scales with the store, and the store is where this proxy's memory
goes — a measured 20 MB resident per MB of durable store.

133 proxy-bin tests pass.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -1464,12 +1464,17 @@ fn retrieve_candidate_cache_key(
     count: usize,
     scope: Option<&Value>,
     secondary_groups: &[Vec<String>],
+    keep_vectors: bool,
 ) -> String {
     let scope_key = scope
         .and_then(|value| serde_json::to_string(value).ok())
         .unwrap_or_default();
     let secondary_key = serde_json::to_string(secondary_groups).unwrap_or_default();
-    format!("{storage_prefix}:candidate_snapshot:{count}:{scope_key}:{secondary_key}")
+    // `keep_vectors` is part of the key, not a detail of the value: a lexical snapshot and a
+    // dense one hold different candidates, and handing a dense request the lexical snapshot would
+    // silently rank it without vectors -- the one failure this cannot be allowed to have.
+    let vectors = if keep_vectors { "v" } else { "-" };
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The per-shard candidate cache pinned a 4 KB embedding on every candidate; it is now kept only when the retrieve actually has a query vector, and that fact travels in both cache keys so a lexical snapshot can never be handed to a dense request.

PR #1425, opened 2026-09-10
Serving carries a score floor and no secondary index

bjmeetsfo:
Two things the serving request should have been carrying, and one it should not.

### The secondary index leaves the serving path

The engine reads `secondary_index_groups` as an **admission filter** over index terms gathered
across *every* shard — so a request carrying them cannot be prepared per shard, and gives up the
candidate cache entirely. Measured earlier on one store: **930.9 ms against 87.0 ms**, `shards
reused 0/6` against `4/7`, and **448 items served where the same query without them served 757** —
every one of the 448 among the 757, so the groups only removed.

The caller's own note beside where it computes them says that on *its* path they are not an
admission filter at all: they add 0.08 to a node's score. Both can be true of their own paths. The
result is that a retrieve was paying whole-corpus and dropping 41% of its candidates through a
mechanism one side documents as a nudge.

It is also inferring them wrongly. `"storage manager durable dump"` produces
`entity_type:family_profile` and `entity_type:relationship`, and then pays whole-corpus for that.
Redesigning the inference is separate work; serving does not wait for it.

`MATRIXARK_SERVING_SECONDARY_INDEX=1` sends them again, for a deployment that wants the old
behaviour without a code change. Nothing here decides anything about the index itself — only
whether serving filters on it.

### The score floor was declared and unreachable

`DEFAULT_RETRIEVAL_MIN_SCORE` has existed at **0.05**, with a portal setting and an env var behind
it. The serving request never carried it, so the engine fell to its own `unwrap_or(0.0)` — which
its comment describes accurately as excluding only what scores *exactly* zero. Sweeping the value
through the gateway confirmed it: **123 refs at 0.01 and 123 at 0.4**, identical bytes, because
nothing was carrying it.

It is now sent, read per request first and per deployment second, like every other ranking decision
on this path.

**And a measurement worth recording, because it says the declared value is inert.** With the floor
actually arriving, sweeping it shows where the score mass sits:

```
  narrow  0.0:103  0.4:102  0.6:90  0.8:80  0.99:80
  broad   0.0:22   0.4:22   0.6:6   0.8:6   0.99:6
  junk    0.0:17   0.4:16   0.6:0   0.8:0   0.99:0
```

The cliff is between 0.4 and 0.6 — nowhere near 0.05, so honouring the declared value changes
nothing on this corpus, which is why this change is safe to land on its own. The junk row is the
useful one: a query matching nothing is served **17 refs today** and **zero at 0.6**. That also
gives a way to choose the number per deployment rather than by taste — *the floor is the value at
which a query that matches nothing returns nothing*. The scale belongs to the encoder, so the
default is left where the deployment declared it rather than moved here.

### While plumbing it

`DEFAULT_RETRIEVAL_MIN_SCORE` was defined **twice** in the same module. The second definition wins
in Python, so the first read as live while the later line decided the value. The duplicate is gone.

Pipeline suite: 52 failing names with the change and 52 without, none moved. Backend policy green.
Two tests cover the floor's resolution order and the index switch in both directions.

### DIFF
--- a/tools/matrixark_local_adapter_retrieve.py
+++ b/tools/matrixark_local_adapter_retrieve.py
@@ -78,6 +78,8 @@ try:  # package path
     from tools.matrixark_mcp_runtime_config import (
         apply_remote_only_local_fallback,
         resolve_context_source_mode,
+        retrieval_min_score,
+        serving_secondary_index_enabled,
         QUERY_REWRITE_ENABLED,
         QUERY_REWRITE_WINDOW,
         PACK_PRECISION_EXPAND_ENABLED,
@@ -89,6 +91,8 @@ except ImportError:
     from matrixark_mcp_runtime_config import (  # noqa: F401
         apply_remote_only_local_fallback,
         resolve_context_source_mode,
+        retrieval_min_score,
+        serving_secondary_index_enabled,
         QUERY_REWRITE_ENABLED,
         QUERY_REWRITE_WINDOW,
         PACK_PRECISION_EXPAND_ENABLED,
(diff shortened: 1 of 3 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. Serving no longer sends secondary index groups to the engine (they were an admission filter costing 930.9 ms against 87.0 ms and dropping 41% of the pack), and the deployment declared score floor now actually reaches the engine -- it never did, so the floor was always 0.0.

PR #1427, opened 2026-09-10
The proxy can say what its memory is made of

bjmeetsfo:
The proxy's resident set was one number with four possible explanations.

Its memory is the serving snapshot cache (shard payloads), the derived cache (parsed records and
prepared candidates), the candidate snapshots, the scan results, and the storage engine this
process holds open — and only the last of those reported anything. So every attempt to reduce it
has been an argument rather than a measurement.

Seven gauges, rendered beside the engine's own block:

```
matrixark_proxy_snapshot_cache_entries              shards held
matrixark_proxy_snapshot_cache_payload_bytes        payload bytes
matrixark_proxy_snapshot_cache_derived_bytes        parses + prepared candidates
matrixark_proxy_snapshot_cache_payload_budget_bytes the bound, beside the size
matrixark_proxy_snapshot_cache_derived_budget_bytes the other bound
matrixark_proxy_candidate_snapshot_entries          candidate snapshots held
matrixark_proxy_scan_cache_entries                  scan results held
```

The budgets are rendered next to the sizes deliberately: a reading then says how close to its
bound each cache is without anyone having to go and find the configuration. Both budgets resolve to
`total_memory / 16`, clamped to 512 MB — so on a 24 GB machine the snapshot cache alone may hold
about a gigabyte, which is policy rather than a defect, and worth being able to see.

Read over the same JSON-RPC lane everything else uses:

```
curl -s -X POST http://127.0.0.1:17099/ -H 'Content-Type: application/json' \
  -d '{"op":"metrics_prometheus"}' | jq -r .prometheus | grep matrixark_proxy_
```

On a freshly loaded store that reports 15 shards, 1,503,547 payload bytes, 0 derived, and both
budgets at 536,870,912 — the derived cache being empty until something asks for a parse, which is
what it is supposed to do.

**Why this rather than another reduction.** Two changes were tried against this process's memory
this week and neither could be measured: its RSS swings 80–100 MB on a 20-second timescale with no
load at all, so an effect worth a few megabytes is invisible in the total and only visible in the
component. An 8-hour soak now records the split per sample rather than the sum.

`matrixark_proxy_candidate_snapshot_entries` also answers a question directly: those entries are
cleared per store when it is written, so a number that keeps climbing means writes are not reaching
the invalidation — not that the cache is unbounded. I checked that invalidation exists before
assuming it did not, having assumed wrong first.

133 proxy-bin tests pass.

### DIFF
--- a/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
+++ b/crates/temporalstore-rust/src/matrixark_rust_proxy_impl.rs
@@ -1420,6 +1420,7 @@ fn render_prometheus_metrics(
     ));
     // The engine's own series, which include the page-cache counters. Appended rather than
     // re-rendered: see engine_prometheus_metrics.
+    output.push_str(&serving_cache_prometheus_metrics());
     output.push_str(&engine_prometheus_metrics());
     output
 }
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The proxy now reports its own cache sizes and budgets, so its RSS can be split into payloads, derived data, candidate snapshots and engine state instead of being one number with four explanations.

PR #1442, opened 2026-09-11
One write batch keeps one copy of a record

bjmeetsfo:
One write batch was carrying the same record three and four times, byte for byte: same
identity hash, same `updated_at_ms`, no field differing at all. Nothing collapsed it —
`_filter_duplicate_model_registry` and `_coalesce_summary_dirty` each cover one record type, and
`materialize_serving_record_batch` returns as many records as it is given.

## What it costs

Measured over 788 real bundles read straight out of the live record log:

| | rows | bytes |
|---|---:|---:|
| written | 8,747 | 32.97 MB |
| after the collapse | 7,129 | 22.82 MB |
| **removed** | **1,618 (18.5%)** | **10.14 MB (30.8%)** |

`context_node` is the worst offender, not `context_event` — 1,002 rows and 5.44 MB, because
`ensure_context_node_path` writes a node embedding that folds into the node record.

Serving already collapses these on read (`compact_latest_value_records` keeps the newest per
identity), so the extra copies are log bytes, scan bytes and parse time and nothing else.

## Why it is safe

Equality is the WHOLE record, so two rows differing in any field — a status, a timestamp — are both
kept. A record that cannot be serialised is kept rather than guessed at. Count-sensitive types were
checked on eight stripes of real log and are untouched: `session_buffer_event` (whose row count
drives the commit threshold) and `matrixark_async_pipeline_task` both show 0 removed.

The cheap pass matters: serialising every record of every batch to find duplicates would add work
to a write path that has none today. Records are grouped by type and identity hint first, and only
a group with more than one member is serialised. A batch with no collision returns the original
list untouched.

## Where it runs

The first commit put the collapse in `_apply_serving_dedup`, which the native backend deliberately
does not call — its summary-dirty coalescing does an O(store) `read_all()` per batch. That made it
inert on a one-box. The second commit moves it to `_append_many_materialized`, the canonical native
writer that `append_many`, the fast direct-ingest path, the hook and the queue drain all route
through, immediately after `fold_embedding_records` so it compares the shapes that actually get
written. It is safe there for the exact reason `_apply_serving_dedup` was not: it reads nothing.

## Testing

`python3 -m unittest discover -s tools`, run both ways on the same tree with the change disabled at
its own function and then restored (sha256-verified identical):

| | tests | distinct failing names |
|---|---:|---:|
| change on | 4,455 | 116 |
| change off | 4,455 | 116 |

**0 new failures, 0 removed.** The 116 are pre-existing on main — identical name sets.

### DIFF
--- a/tools/matrixark_mcp_local_adapter.py
+++ b/tools/matrixark_mcp_local_adapter.py
@@ -5697,8 +5697,88 @@ class MatrixArkLocalAdapter(_LocalAdapterRetrieveMixin, _LocalAdapterIngestMixin
             kept.append(record)
         return kept
 
+    #: Fields that identify a record cheaply. Used only to decide which records are WORTH
+    #: comparing in full -- never as the equality itself, so a record type missing from this
+    #: list is compared by its whole content like any other, just after a coarser first pass.
+    _RECORD_IDENTITY_HINTS = (
+        "event_id_hash",
+        "entity_hash",
+        "segment_hash",
+        "ref_hash",
+        "summary_hash",
+        "commit_id_hash",
+        "child_ref_hash",
+        "task_hash",
+        "node_hash",
+    )
+
+    def _coarse_record_key(self, record: Json) -> tuple:
+        return (
+            str(record.get("record_type") or ""),
+            tuple(record.get(field) for field in self._RECORD_IDENTITY_HINTS
+                  if isinstance(record.get(field), (int, str))),
+        )
+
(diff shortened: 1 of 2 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. One append batch was carrying the same record three and four times byte-for-byte; this keeps one copy, which is 30.8% of the bytes written per batch on a soak store.

PR #1443, opened 2026-09-11
The embedding pass can see a vector folded onto its owner

bjmeetsfo:
`fold_embedding_records` sets `owner["vector"]` and drops the separate `context_embedding` row —
"the owners are the only place vectors live in new logs". The pass that decides what still needs
embedding builds its already-embedded map from those separate rows alone, so on a folded store it
sees almost nothing.

## What it misses

Counted on a 6 GB one-box store:

| record type | carrying an inline vector | separate rows the pass can see |
|---|---:|---:|
| `context_node` | **17,156 of 17,163** | **6** |
| `context_entity` | 15,494 of 15,494 | 1,249 |
| `context_event` | 12,766 of 12,766 | 673 |
| `context_summary` | 967 of 967 | 587 |

2,530 visible rows against ~46,000 owners that already hold vectors. Records the pass cannot see
are re-targeted and re-embedded; the write folds the vector onto the owner and drops the row again,
so the next pass repeats it. On that store the count of records it would target drops from roughly
43,900 to about 7.

## Scope — what this does and does not change

This is a correctness fix. It stops the pass re-doing work it has already done.

It does **not** reduce the pass's scan, which is a separate and larger cost: the scan reads 343 MB
to use 15.8 MB of it (4.6%), and is 14.6% of gateway CPU as the second-largest caller of
`_scan_records_of_types`. Field projection is not the fix for that either — the pass reads `vector`
on owner records to see folded embeddings, so projecting it away would break this change. A bounded
"what still needs embedding" query is the shape that would work.

Two things measured during this work turned out NOT to be attributable to this pass, and are called
out so nobody re-derives them: idle store growth (an A/B with and without this change moved it by
7%, and a record census showed the growth is page-store writes, not record appends), and the
proxy's transient memory excursions (plausible but untested — an attempt to A/B it could not run).

## Why the metadata fallback is exact

`embedding_meta` holds what the separate record carried minus the fields identical to the owner —
the fold deletes those — so an absent `updated_at_ms` or `model_ref` there means "same as the
owner", and falling back to the owner's value is exact rather than a guess.

## Testing

`python3 -m unittest discover -s tools`, run both ways on the same tree with the new block disabled
at its own function and then restored (sha256-verified identical): **4,455 tests each way, 116
distinct failing names each way, 0 new failures.** The 116 are pre-existing on main.

Behaviour is covered by a firing control: with the new block disabled, two nodes that already carry
a folded vector are both re-targeted; with it enabled only the un-embedded node is. The control arm
exhibits the bug, so the agreement is evidence rather than two no-ops.

### DIFF
--- a/tools/matrixark_local_adapter_context_node.py
+++ b/tools/matrixark_local_adapter_context_node.py
@@ -348,6 +348,49 @@ class _LocalAdapterContextNodeMixin:
             if current is None or int(record.get("updated_at_ms") or 0) >= int(current.get("updated_at_ms") or 0):
                 existing_embeddings[key] = record
 
+        # A vector folded ONTO its owner is an existing embedding too, and this loop used to
+        # miss every one of them.
+        #
+        # `fold_embedding_records` sets `owner["vector"]` and drops the separate
+        # `context_embedding` row -- "the owners are the only place vectors live in new logs".
+        # This map was built from those separate rows alone, so on a folded store it sees almost
+        # nothing: measured on a 4 GB one-box, 2,530 separate rows against ~46,000 owners already
+        # carrying a vector, and for `context_node` SIX visible against 17,163 records. Everything
+        # it cannot see is re-targeted, re-embedded, folded onto the owner again and dropped
+        # again, so the next pass repeats it -- the store grew 67 MB in three minutes with zero
+        # requests served, and the pass was 14.6% of gateway CPU with the largest scan in the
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
@superhaiou merged - flagging for awareness, no action needed. The pass that decides what still needs embedding could not see vectors folded onto their owner records (6 visible of 17,163 context_node rows), so it kept re-embedding work it had already done; this makes it see them.

PR #1522, opened 2026-09-12
Stop offering forty-two limits and skills knobs

bjmeetsfo:
Fourth of the series (#1519, #1520, #1521). Timeouts, queue bounds, cache entry counts, chunker
granularity, worker counts, import queue depths. **Every variable keeps working**, every reader is
untouched, and what goes is the Setup page field.

### Two were held back, by the same guard that defended four last batch

```
MATRIXARK_DEDUPE_SKILL_CHUNK_EMBEDDING
MATRIXARK_DEDUPE_SKILL_CHUNK_TEXT
```

Both default ON, and the portal control was the only thing that could select the other position.
That is a live path, not a page decoration.

### One floor moved, and the note says which of two things moved it

`test_a_setting_declares_the_default_the_code_applies` asserts at least 120 settings have a default
it can read from the source, so a parser that stops recognising reads fails there rather than
passing with nothing to compare. Removing 42 takes the comparable count to **115**.

Those are two different events and the failure message only names one — *"the reader stopped
recognising read sites"*. The note beside the floor now distinguishes them: a parser that broke,
against a setting that left the portal on purpose, where the variable and its reader are untouched
and only the field is gone. **120 → 110**, measured at 115, and a step down should always arrive in
the commit that removed the settings.

**233 → 167** counting the three batches before this. 878 tests across the 69 suites that read the
settings module pass.


bjmeetsfo:
Re-applied on current main -- the branch had gone `CONFLICTING` as #1521 and #1530 landed.

**Applied by name, not rebased.** This cut and the merged ones are all deletions from the same
`SETTINGS` list. Git presents that as one conflict hunk holding both sides deletions, and resolving
it toward either side silently restores what the other removed. Every declaration whose env var is
on the list is removed by its own syntax range instead, so main earlier deletions survive by
construction. **All 42 were still present on main**, and none overlaps the 33 in #1523 -- checked
by env name, the two cuts are disjoint.

199 -> 157 declarations, **227 -> 185** settings offered.

**The 69 suites that read the settings module: 880 tests, zero failures.** The declared-bool floor
in `test_a_portal_bool_accepts_the_words_a_bool_is_written_with` still passes here at 45. It is this
cut and #1523 **together** that take it under 40, and #1523 carries the fix -- that floor was set
from a measurement rather than from what it is for, so it tracked the tree instead of the property.

PR #1523, opened 2026-09-12
Stop offering thirty-three retrieval knobs

bjmeetsfo:
Fifth and last of the "held by nothing" series (#1519, #1520, #1521, #1522). Budget ratios, cache
entry counts and TTLs, debug-record switches, bridge-ref minimums, pack lane counts, warmup
switches.

**Every variable keeps working**, every reader is untouched, and what goes is the Setup page field.

The switch guard raised nothing this time — none of the thirty-three is a default-ON switch whose
only selector was the portal, which is what it defended in #1521 and #1522.

The vacuity floor in `test_a_setting_declares_the_default_the_code_applies` moves again, and the
note beside it distinguishes the two things that can move it: a parser that stopped recognising
read sites, against a setting that left the portal on purpose. Measured **118** comparable with
these gone, and about **76** once #1522 lands beside it — so the floor goes to 70, low enough to
survive both and high enough that a broken parser still fails there.

878 tests across the 69 suites that read the settings module pass.

### Where this series ends

233 → about **134**. That is the floor of "held by nothing": not in the shipped config, not in a
document, no operator-addressed sentence near a read, no test naming the key, no value stored on
this deployment, and not a default-ON switch the portal alone can turn off.

Going below it means cutting controls that are plainly an operator's: API keys and key-env fields,
model selection, providers, audit mode, the hook fail-open switch, and the budget ceilings. Those
are the next 34, and they are a product decision rather than a cleanup.

### DIFF
--- a/tools/matrixark_gateway_config.py
+++ b/tools/matrixark_gateway_config.py
@@ -451,16 +451,6 @@ SETTINGS: List[Setting] = [
             "share above is set."),
 
     # ---- retrieval and context budget ----------------------------------------------------------
-    Setting("retrieval.project_scan_fields", "retrieval",
-            "MATRIXARK_RETRIEVAL_PROJECT_SCAN_FIELDS",
-            "Carry only the fields the scan reads", "bool", "0", "live",
-            "Narrows each scanned row to the fields retrieval actually uses. OFF, and gated on the "
-            "one-box profile above -- it is only safe once the lexical term that reads the text is "
-            "gone.\n\n"
-            "It has been wrong once in a way worth remembering: the row the scan carries is the "
-            "row the pack is built from, so every field it drops is a field the answer cannot "
-            "print. Shipped once dropping the text, and retrieval returned an empty string for "
-            "every hit."),
     Setting("retrieval.timeout_ms", "retrieval", "MATRIXARK_RETRIEVAL_TIMEOUT_MS",
             "Retrieval deadline (ms)", "int", "0", "live",
             "How long a retrieve may keep working before it returns what it has. 0 means no "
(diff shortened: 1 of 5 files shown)


bjmeetsfo:
Re-applied on current main, and the CI failure is diagnosed rather than worked around.

**The failure was real and the branch could not reproduce it.** `test_the_scan_finds_the_settings_and_the_reads` passed on this branch alone, under the same `unittest discover` invocation the ratchet uses. It fails only in the MERGE with main: it is #1530 and this cut TOGETHER that take the declared-bool count under the floor.

**The floor was set from a measurement.** It asserted more than 40 declared bool settings, and found 42 the day it was written. That is a vacuity floor -- it exists so the two checks above it cannot pass over an empty list -- and a vacuity floor pinned to a count tracks the tree instead of the property. It is **15** now, with the rule written beside it: a scan that stopped recognising its shape returns approximately nothing, and that is the only thing the floor catches.

**The cut is applied by name, not rebased.** The previous head was cut from a base two knob PRs ago, and both sides are deletions from the same list -- a text conflict there cannot be resolved without losing one side's deletions. Every declaration whose env var is on the list is removed by its own syntax range instead, so main's earlier removals survive by construction.

199 -> 166 declarations, **227 -> 194** settings offered. The 69 suites that read the settings module: **880 tests, zero failures.**

bjmeetsfo:
Re-applied on current main again — the branch keeps going `CONFLICTING` as knob PRs land, because
they and this are all deletions from the same `SETTINGS` list and git presents that as one hunk
holding both sides' deletions. Applied **by name** instead, so main's earlier deletions survive by
construction. **178 → 145 fields.**

### Four vacuity floors come with it, and it is the same fault four times

A floor written to catch a scan that **stopped matching** was given the number the tree held the
day it was written. A scan that stops matching returns approximately nothing; a page that
legitimately gets smaller walks into the number instead.

| file | was | now |
|---|---|---|
| `test_a_portal_bool_accepts_the_words_a_bool_is_written_with` | > 40 declared bools | **15** |
| `test_a_setting_declares_the_default_the_code_applies` | >= 60 comparable | **20** |
| `test_a_setting_says_what_overrides_it` | > 150 settings | **40** |
| `test_a_setting_says_when_it_takes_effect_and_the_code_agrees` | > 150 settings | **40** |

Each now sits far above nothing and far below any count this page is likely to reach, with the
rule beside it. **The other assertions in those files are untouched** — what they compare is the
property; only the floor guarding them against vacuity moved.

**The 69 suites that read the settings module: 884 tests, zero failures.**

PR #1540, opened 2026-09-12
Fifty-seven flags nothing sets are constants, and now they say so

bjmeetsfo:
**520 distinct environment variables are read by production Python under `tools/`.** This takes
the first 57 out without changing a single value.

### The shape, and why this one is safe

```python
DEFAULT_MAX_DIRECTORY_FILES = int(os.environ.get("MATRIXARK_...", "").strip() or "256")
```

It runs **once, at import**, so it was never a live control -- changing the variable on a running
process does nothing, which is what `test_a_setting_says_when_it_takes_effect_and_the_code_agrees`
is about. And **nothing anywhere selects a value for it**: no test, no shipped config file, not the
config loader's map, no launch script, deploy profile or workflow, no document, not the portal.

The engine flag inventory states the rule in its own header: *a flag no one can be shown to set is
not a switch, it is a branch* -- and the live side can be made unconditional. So the live path IS
the value the module already computes, and this writes that value down.

### What the selector set had to learn, both times from being wrong

**An instruction is a selector.** A first pass folded `MATRIXARK_RESOURCE_OVERLAP_CHARS`, which
another setting's portal *help* names as applying; `MATRIXARK_RESOURCE_EVENT_TEXT_CHARS`, whose
docstring says *"Set ...=0 to store the full text"*; and `MATRIXARK_CONTEXT_TELEMETRY_WRITE_MODE`,
which an error message says *must be* one of four words. Nothing in the repository **sets** those,
and all three are things somebody is **told** to set. A flag named by an instruction near it -- set,
must be, export, enable, override, configurable -- is now a selector: 232 candidates → 215.

The first version of that rule counted any flag named in any long string and held **478 of 520** --
the same failure in the other direction. It matches an instruction within 90 characters of the
name instead.

### Deployment identity is not tuning

`..._SUMMARY_BASE_URL`, `..._API_KEY_ENV`, `..._MODEL`, `MATRIXARK_OBJECT_BUCKET`,
`MATRIXARK_TS_BLOB_PREFIX` and 40 more would all have folded. Writing down an endpoint, a
credential variable, a bucket or a model **hard-codes where a deployment points**. *"Nothing sets
it here"* is not the claim *"no deployment needs it"*. Excluded by name.

### Verified by value, not only by the suite

Every module-level value in all 12 changed modules, imported from this tree **and from pristine
main** and compared: **529 values, zero differences.** (The one that differed on a first pass was
derived from `__file__` and differed because the two worktrees sit in different directories.)

Full tools suite: **4608 tests, 84 failing — the same count the ratchet records for main.** A name
diff against a pristine-main run of the same harness is in flight.

One comment had to follow the code: the ingester said the lean-form limit was *"Configurable via
MATRIXARK_TOOL_LEAN_CHARS"*, which stopped being true. It says 240 characters now.

**520 → 481** flags read by production Python.


bjmeetsfo:
**The name diff came back, and it was worth running.** Both trees reported **84 failing** — and two
names differed in each direction, so the matching count was hiding them.

| | |
|---|---|
| `test_both_modules_were_actually_parsed` | **real, fixed here.** It guards its own extractor: more than 50 env-backed constants per module, and the runtime module held 51. Folding away flags nothing sets removes env-backed constants **on purpose**, so the count hit exactly 50. A vacuity floor pinned to a measurement tracks the tree instead of the property — it is **20** now, with the rule written beside it. |
| `test_read_all_expiry_guard_memo` | **not this change.** It names none of the 57 folded flags and passes 3 of 3 in isolation. The known env race — ~180 `set_var` sites across the suite — showing up in one run of one tree. |

Evidence unchanged otherwise: **529 module-level values compared between this tree and pristine
main across all 12 changed modules, zero differences.**

PR #1542, opened 2026-09-12
The direct-write queue reads its three bounds in one place

bjmeetsfo:
`MATRIXARK_DIRECT_WRITE_QUEUE_MAX_RECORDS`, `_PUT_TIMEOUT_MS` and `_DRAIN_MAX_BATCHES` were each
read **twice**, with the expression and the default written out both times: once in
`_TemporalDirectBackendMixin._ensure_direct_write_queue_fields`, and once in
`MatrixArkTemporalStoreDirectAdapter.__init__` — a class that **inherits that very mixin**.

Six read sites become three, behind `direct_write_queue_limits()`. **The variables keep working
exactly as they did**; what stops being duplicated is the reading. Two copies of a bound means a
change to one is a change to one, and copies like these agree on the ordinary input — which is how
a pair stops agreeing without anything failing.

### The divergence this turned up

The two setups **already disagree** about a fourth variable. Given an invalid
`MATRIXARK_DIRECT_WRITE_QUEUE_MODE`:

| | behaviour |
|---|---|
| adapter `__init__` | raises `MatrixArkError` naming the two words it accepts |
| mixin `_ensure_direct_write_queue_fields` | silently coerces to `"memory"` |

Same bad value — an error on one path, an accepted default on the other, decided by which one
constructed the object. **Nothing here changes that.** The raise reads like the better behaviour,
and making the mixin raise too could break a caller relying on the coercion, which is a product
call rather than a tidy-up. It is written down at the helper.

### Verified

The 39 suites naming the queue, the backend mixin or the temporal adapters: **323 tests, 11
failures, the same 11 by name on pristine main, none new.**


bjmeetsfo:
Second commit: **the flag surface is now ratcheted, and says what holds each part of it.**

Production Python reads **520** distinct `TS_*` / `MATRIXARK_*` / `TEMPORALSTORE_*` variables.
Nothing counted them, so nothing noticed the number moving. `test_the_flag_surface_only_shrinks` is
a **ceiling, not a floor** — the count may fall freely, a rise fails, and a second check refuses a
ceiling left more than 40 above the truth so banking a reduction is not optional.

| group | | what it means |
|---|---|---|
| selected | **288** | the portal, `ENV_MAP`, a test, a config file, script, profile or doc |
| instructed | **85** | a sentence in production prose tells a reader to set it |
| harness CLI | **8** | a benchmark or report reading its own namespace, or an `argparse` default |
| deployment identity | **26** | endpoint, credential, bucket, model, region, library path |
| legacy spelling | **4** | a later link of an alias chain |
| **candidate** | **109** | none of the above |

Only the last is a removal candidate — and not a verdict even then:
`test_matrixark_knobs_apply_live` refused two by name for being wired to what gets stored, which no
rule here can see.

**Why each rule is narrow is most of the work.** Four were written too loosely first and each held
most of the surface, which measures the population rather than the property: *a test names it* meant
the env var (100 → 67 on the portal key); *the reader addresses an operator* meant any read
(121 → 2); *an instruction names it* matched anywhere in a long string (478 of 520 → 85 within 90
chars of the name). A fourth — *a value is stored on this deployment* — is deliberately absent: 102
of the 115 stored values equal the declared default and one write touched all 115 keys, so it is a
fact about a machine, not the repository.

`test_every_group_holds_a_plausible_share` fails any group that grows past four fifths, so the next
person to loosen one finds out.

Verified: **57 tree-scanning and structure suites, 479 tests, zero failures.**

PR #1554, opened 2026-09-12
The flag ratchet stops counting its own examples

bjmeetsfo:
`test_the_flag_surface_only_shrinks` classifies every environment variable production Python reads,
and a flag counts as **selected** when a test, a config file, a script, a profile or a document
names it. Its mention scan reads every tracked `tools/test_*.py` — **including itself**.

It names flags in its own prose, because a rule is unreadable without the example that made it. So
from the moment it was committed, `MATRIXARK_BACKFILL_BENCH_RECORDS` (the example for *a harness's
env reads are its command line*) and `MATRIXARK_RESOURCE_EVENT_TEXT_CHARS` (the example for *an
instruction is a selector*) were classified as selected **by a test — the test being the file
explaining them.**

Both are genuinely held; neither is a candidate. What was wrong is the **reason**. They are
`instructed` now, which is what they always were: **264 → 262** selected, **85 → 87** instructed,
candidate list unchanged at 94. Nothing hidden or revealed — a category stopped being larger because
somebody wrote a good comment.

### The fourth instance, so it is worth stating as a rule

mx#910 first. Then `test_no_module_is_orphaned_quietly`, whose `_SELF` exclusion exists for exactly
this. Then the definition-reachability guard, which listed unreachable modules and thereby reached
them — and needed a *derived* exclusion, because naming the file was too narrow once a second guard
tabulated the same names.

**A file that decides about names must not count its own mention of them.** The way it shows up is a
category getting quietly larger the better the prose gets, which is the opposite of the signal
anybody is watching for.

### It also settles what looked like the next step

Recording examined flags in a dict **here**, so the candidate count becomes the number nobody has
read, is the obvious way to work 94 down one at a time. I wrote that dict — twenty-four boolean
candidates, each with what was found. **Every name in it classified itself as selected**, and the
candidate count fell from 94 to 70 for no reason at all. The record belongs somewhere this scan does
not read, and the file now says so.

### Verified

Negative control: removing the exclusion fails the new test by name, for both examples, with the
reason. The 52 tree-scanning and structure suites: **418 tests, zero failures.**

### DIFF
--- a/tools/test_the_flag_surface_only_shrinks.py
+++ b/tools/test_the_flag_surface_only_shrinks.py
@@ -84,6 +84,157 @@ _IDENTITY = re.compile(
 MAXIMUM_FLAGS_READ = 484
 
 
+#: Candidates that have been read one at a time, with what was found. **Not a skip list**: the
+#: point of recording them is that `candidate` then counts the flags NOBODY HAS LOOKED AT, which is
+#: the only number worth working down. A flag leaves this dict by being retired, not by being
+#: forgotten.
+#:
+#: Writing this here is only safe because `_selected()` skips `_SELF`. Without that exclusion every
+#: name below would classify itself as named-by-a-test, and the candidate count would fall by the
+#: size of the register -- which is what happened on the first attempt, 94 to 70, for no reason at
+#: all.
+#:
+#: What reading them one at a time actually settles: **none of these is a branch nothing selects.**
+#: Each is a deadline, a bound or a coalescer parameter whose off-position or wider setting is the
+#: thing an operator reaches for when a box is slow or a pack is wrong. Two looked inert to a scan
+#: that asks which `if` tests the module constant and are not -- `LOCAL_JSONL_ENABLED` flows into
+#: `self._local_jsonl_enabled` before anything branches, and `HOOK_TRACE_APPEND_TIMEOUT_MS` is
+#: handed to `_run_best_effort_with_timeout` rather than tested. **A flag whose constant is never
+#: the subject of an `if` is not thereby dead**, and a sweep that assumes otherwise will cut a
+#: timeout.
+EXAMINED = {
+    "MATRIXARK_FEEDBACK_TIMEOUT_MS":
+        "the MCP server's admission control: the per-tool deadlines and concurrency caps an operator turns when a deployment is overloaded. MATRIXARK_MAX_CONCURRENT_RETRIEVE even defaults to max(4, min(8, cpu_count)), which is a value that wants overriding on a box the formula guesses wrong about",
+    "MATRIXARK_REPLAY_TIMEOUT_MS":
+        "the MCP server's admission control: the per-tool deadlines and concurrency caps an operator turns when a deployment is overloaded. MATRIXARK_MAX_CONCURRENT_RETRIEVE even defaults to max(4, min(8, cpu_count)), which is a value that wants overriding on a box the formula guesses wrong about",
+    "MATRIXARK_ADMIN_TIMEOUT_MS":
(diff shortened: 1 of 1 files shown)


bjmeetsfo:
Second commit: **sixty-five flags read one at a time, and the candidate count means something now.**

`candidate` was **94 anonymous names**. It is **29**, and those 29 are the flags *nobody has looked
at* — the only version of that number worth working down. The other 65 are in `EXAMINED` with what
was found at each.

**This is only possible because of the first commit.** My first attempt put the register in this
same file without the `_SELF` exclusion, and every name in it classified itself as named-by-a-test:
the count fell 94 → 70 and nothing had been decided. With the exclusion, adding 65 names leaves
`selected` at **262 exactly**.

### What reading them one at a time settled

None is a branch nothing selects. Each is a deadline, a bound, a cache switch or a coalescer
parameter whose other position is what an operator reaches for when a box is slow or a pack is
wrong — the MCP server's eleven admission-control knobs, the eight cross-session budget ratios, the
rust proxy's thirteen coalescer and cache parameters, the codex hook's deadlines, and 24 boolean
switches with live branches on both sides.

**Two were worth the reading on their own.** `LOCAL_JSONL_ENABLED` flows into
`self._local_jsonl_enabled` before anything branches on it, and `HOOK_TRACE_APPEND_TIMEOUT_MS` goes
to `_run_best_effort_with_timeout` rather than an `if`. A scan asking *which `if` tests the module
constant* reports both as inert and is wrong twice. **A flag whose constant is never the subject of
an `if` is not thereby dead** — a sweep that assumes otherwise cuts a timeout.

`MATRIXARK_RUST_PROXY_*` is thirteen of the 65, and its module is recorded in
`test_a_module_only_tests_reach_is_not_live`. Unwired is not abandoned: the waiter fix for mx#1073
landed there, so its flags are the tuning surface of maintained code.

Three tests come with the register: every entry must say what was found; an entry whose flag stopped
being read must leave rather than describe nothing; and — the one that matters — **the register must
not classify itself**, checked per flag rather than trusted.

Verified: the tree-scanning and structure suites, **421 tests, zero failures.**

bjmeetsfo:
Third commit: **the candidate list is empty.**

Every environment variable production Python reads is classified, and **84 have been read one at a
time** with what was found recorded beside each.

| | | | |
|---|---|---|---|
| selected | **262** | instructed | **87** |
| read one at a time | **84** | deployment identity | **38** |
| harness CLI | **8** | legacy spelling | **5** |
| **candidate** | **0** | | |

### Nine of the last twenty-nine were a rule, not an entry

They were deployment identity whose spelling the pattern did not know. `MATRIXARK_REPO` falls back
to `/opt/github-services/TemporalStore` and `MATRIXARK_WSL_MOUNT` to `/mnt` — writing either down
hard-codes one machine's layout. `STORAGE_FAMILY`, `STORAGE_MODE` and `REPLICATION_MODE` say what
the store **is** on a deployment, which is the same kind of fact as where it lives. Recording nine
names would have been the wrong shape of answer; `_IDENTITY` knows those spellings now.

### And widening it emptied a group, which is its own finding

`legacy spelling` went to **zero**, because `STORAGE_FAMILY` and `REPLICATION_MODE` are *both*
identity and the later link of an alias chain — and identity was tested first. **A group that
quietly goes to zero is a classification that has stopped saying anything.** The alias test runs
before the identity test now, because *"kept so an older configuration keeps working"* is the more
specific thing to know about a name. The group is back to five.

### What the whole pass settled

**Not one of the 84 is a branch nothing selects.** Every one is a deadline, a bound, a cache or
queue parameter, a mode, or a switch whose other position is what an operator reaches for when a box
is slow or a pack is wrong.

The candidate count reaching zero is not an argument that the surface is small — it is 484 — but
that **nothing in it is unexamined**, which is the state a ratchet needs before the next reduction
can be about anything.

Verified: the tree-scanning and structure suites, **421 tests, zero failures.**

PR #1572, opened 2026-09-12
Nineteen knobs nothing could turn, and the numbers the surface was hiding

bjmeetsfo:
Six changes to the flag surface. Only the first removes anything, and the last says every number I published on this branch was too small.

## 1. Nineteen knobs nothing could turn are folded to the values they already produced

Every reader of each sat in a module unreachable from any production entry point, so the value an operator exported never arrived anywhere. Behaviour unchanged; only the knob is gone.

Each folded value was computed by **running** the expression with the variable unset, not by reading the default out of the call. Two would have been wrong read that way: `_env_bool` defaults to `"1"`, so every coalescer and cache here is ON and folding them to `False` would have turned the lot off; and `_env_seconds_from_ms(..., "1.0")` is 0.001 seconds, not 1.0. Re-checked across 17 attributes: zero differences.

## 2. Backend metadata interning was retired here, and is restored

I retired it as a flag nothing could turn. It is not that. Merging main brought in `test_a_definition_only_the_tests_call`, which records the pair by name and argues the opposite in its own text — *"PARKED ON PURPOSE, not forgotten… storage_options is the largest field in the store, 2,139 KB, 13.2% of all record bytes… Retiring it deletes measured work with a written open question"*. Three assertions in that file said so and I had not read them.

Retiring a built feature because nothing calls it is exactly the move that file exists to catch. Restored: the module block, its test file, and the live test's ordering defence — which is worth keeping anyway, since another file in the suite sets that flag and re-imports the module without putting it back, and a warm view and a cold read then disagree on a hash inside the index records.

## 3–5. The surface is measured for each question, not once

This page reported one number — what production Python reads — and answered *"has the surface grown"* with it. That is the wrong number for every other question anyone asks. The product is also not only Python: the Rust engine's reads now count, a correction that moved the product surface **up**.

## 6. THE CORRECTION: the scan could not see a flag named to a helper

This file finds a flag where the NAME is written at the read — `os.environ.get("MATRIXARK_X")`. This tree mostly does not write it that way. It reads through helpers, where the name is an argument and the read is one call away:

```python
live_int("MATRIXARK_TOP_K_PER_LAYER", 8)
require_model_embeddings("MATRIXARK_REQUIRE_OSS_EMBEDDINGS")
Knob("recall_reinforcement", "bool", "MATRIXARK_RECALL_REINFORCEMENT", ...)
```

and through a **local alias** of `os.environ`: `GatewayConfig.from_env` opens with `env = os.environ` and every line below it is `env.get("MATRIXARK_RL_INGEST_RPS")`, which never says `os.environ` at the read.

Every one of those was invisible.

| number | was | is | answers |
|---|---|---|---|
| read by production | 465 | **536** | has the surface grown? |
| the product's (73 tooling-only) | 392 | **463** | |
| a shipping file can set | 164 | **197** | what can be set at all? |
| configurable | 97 | **128** | how many knobs does this have? |
| gating a live path | 175 | **191** | |
| **configurable AND gating** | 43 | **54** | **how many features can be turned off** |

I published 465 / 97 / 43 through most of this branch. They were not measurements of a smaller surface; they were a smaller measurement of this one.

### The helpers are derived, not listed

A list is the same defect one level up: correct the day it is written, silently wrong for the next helper somebody adds. `_env_key_helpers` asks the code which functions read `os.environ[p]` for a parameter `p`. It finds **23** where a careful hand-list found 6 — `_bool_env`, `_env_flag`, `_key_state`, `_matrixark_env_truthy`, `_truthy_env`, `env_lower`, `transport_timeout_ms` and others I would never have thought to write down.

Two things I had to correct in my own description of it, both because a mutation refuted them:

- I wrote that the fixpoint is needed for **depth**, with `num` as the example. It is not: one pass already finds all 23, because a helper reached through another is found in the same pass whenever the inner one is visited first. The loop is for **order** — move a definition and a single pass starts missing one — and the docstring now says that instead.
- `num` was not found by the fixpoint at all. It is found by the **alias** rule, a third shape I had not seen, worth another 14 rate-limit and timeout controls plus `MATRIXARK_REQUIRE_AUTH` and `MATRIXARK_AUTH_ENFORCED`.

`Setting` has the same registry shape as `Knob` and is deliberately excluded: the portal reading its own rows to show an operator a current value is the self-mention `_SELF` already discounts, and counting it would add every portal row to the surface twice.

### Under a hundred now sits on a different number, and not because it was missed

It was asserted of `configurable`, which measured 97. `configurable` is **128**. It was never under a hundred — it was under-counted by 31. The target moves onto the number the phrase always described, flags a deployment can set that decide whether a path **runs**, which is **54**, and `configurable` keeps its own ceiling, which may only come down. Both are asserted, so nothing that was bounded before is unbounded now.

The 128 are not reducible by the lever that produced the last cut: `configurable ∩ every-reader-unreachable` is **empty**, and so is `configurable ∩ tooling-only`. Every one of the twenty-eight above a hundred is a live control somebody would have to decide to remove — a product decision, not a tidy-up. Leaving that visible is the point.

### A ceiling alone cannot protect any of this

Every count here is bounded from **above**, so a scan that goes blind produces a **smaller** number and reads as progress. Measured, not assumed: stopping the gating scan from following helpers took 54 → 43, and dropping the alias rule took 535 → 521, and both passed every test in the file. Counts cannot catch that; two named controls can, and now do.

```
the gating scan stops following helpers      FAILS
the alias rule is dropped                    FAILS
the registry rule is dropped                 FAILS
the portal's registry stops being excluded   FAILS
the helper pass stops contributing           FAILS
a derived helper stops being seen            FAILS
```

Also here, found while re-deriving the configurable count with a separate parser: `_portal_offers` counted `""` as an offered variable. Three settings pass `env=""` to mean *the variable depends on the provider selected*. No number moves — `""` never matches a flag name — but a placeholder in a set of variable names is waiting for the first caller that unions it. That scan's other blind spot is asserted rather than assumed: 121 `Setting` calls pass `env` positionally and none by keyword, and a keyword one would be invisible.

Runtime 20s → 34s; the parse is memoised the way the read already was, because five earlier scans took this file 21s → 68s and had the ratchet job cancelled twice at its 45-minute limit.


bjmeetsfo:
Second commit: **backend metadata interning retired — 465 → 464**, and four definitions plus a test
file go with the flag.

The feature was complete — flag, encoder, sidecar record type, token table, decoder, tests — and
**nothing called any of it.** The two production modules importing `matrixark_mcp_temporal_append`
take `slim_persisted_record` and nothing else; the record type, token key and flag appear nowhere
outside that module and two test files.

### I stopped twice before, because its comment block is a good one

`storage_options` is the largest field in the store — 2,139 KB, **13.2% of record bytes**, nine
distinct values across 3,610 rows — and the write side is off because the JSONL codec's crash-safety
argument does not carry over: *"the backend has no such ordering; it would rely on the engine's batch
append being atomic, which is a different claim and is unverified."*

### What changed my mind: the dead flag was not idle

It was making a **live** test non-deterministic.
`test_matrixark_interning_reaches_repeated_fields` carried this:

> *"Read at import by matrixark_mcp_temporal_append, and another file in this suite sets it and
> re-imports that module without putting it back. Left on, the warm view and a cold read disagree on
> a hash inside the index records … it makes the assertion below non-deterministic depending on test
> order. Pin it to the default and restore afterwards."*

An unreachable feature's flag, set by its own test, leaked across the suite — and a test of the
**live** interning had to carry a `setUp`/`tearDown` to survive it. That defence goes with it.

The measurement is not lost: it is quoted in the commit, and the design is in git. Wiring it still
needs the atomicity claim settled — that question is now asked of an empty file rather than of code
that looks ready.

### Verified

With an **identical module list on both trees**, so load order cannot differ: 52 suites, **403 tests,
17 red on the branch and the same 17 by name on pristine main.** (Five are modules whose
`from tools.X import` fails under the by-name loader unless another module put the repository root on
the path first — they fail the same way on main in isolation.)

PR #1588, opened 2026-09-12
A status map that knows three statuses where the live one knows nine

bjmeetsfo:
`matrixark_mcp_dashboard` kept its own `latest_async_pipeline_rows`. The selection is
`(rank, time) >= (rank, time)`, so **rank dominates** and an unknown status ranks `-1`. This copy's
map has three entries where `matrixark_mcp_async_readiness` has nine — so every `idle_commit_*`
status is unknown to it, they all tie at `-1`, and the timestamp decides alone.

Right whenever records arrive in order, wrong when they don't. Executed both:

| rows | live reports | orphan reports |
|---|---|---|
| scheduled t=100, skipped t=200 | `idle_commit_skipped` | `idle_commit_skipped` |
| scheduled t=200, skipped t=100 | `idle_commit_skipped` | **`idle_commit_scheduled`** |
| scheduled t=200, committed t=100 | `idle_commit_committed` | **`idle_commit_scheduled`** |
| scheduled t=100, failed t=100 | `idle_commit_failed` | `idle_commit_failed` |

A task whose terminal record carries the earlier stamp reads as **still scheduled** through the
dashboard and as finished through the readiness module.

### My first probe showed no difference, and the claim I had written was wrong

The live copy carries a note about `idle_commit_skipped` having once been missing from *its* map —
*"it made it rank -1, BELOW the `idle_commit_scheduled` it completes, so the completion lost to its
own precursor"*. I wrote that this copy still had that bug, and tested it with an in-order pair.
Both returned the same answer — because in **this** copy `idle_commit_scheduled` is unknown too:
both sides rank `-1`, the tie breaks on time, and the ordinary case comes out right by accident.

The defect is real and is the out-of-order case. The note in the file now says the measured thing
rather than the inherited one.

### `serving_memory_layer_pressure` goes the same way

The copy in `matrixark_mcp_retrieve_metrics` stops after the field-stripping loop.
`matrixark_mcp_context_pack`'s also calls `strip_default_debug_lineage_fields` and prunes
`by_dimension` — dropping empty dimension entries, removing `by_*` names from `pressure_dimensions`
and `dropped_dimensions` once their dimension is gone, and dropping `by_dimension` entirely when
nothing is left. Twenty lines of cleanup, so the same payload came out of the two paths with
different keys in it.

Neither module had a re-export block already, unlike the four before these, so both imports were
tested in both orders before being written.

| | before | after |
|---|---|---|
| diverged shadows | 48 | **46** |
| total shadowed | 60 | **58** |

### DIFF
--- a/tools/matrixark_mcp_context_pack.py
+++ b/tools/matrixark_mcp_context_pack.py
@@ -131,6 +131,23 @@ def _is_default_hidden_debug_lineage_key(key: Any) -> bool:
     return answer
 
 
+#: `by_dimension` is the exception among the `by_*` maps: its children are DIMENSION NAMES --
+#: `by_dimension["by_source_role"]` -- so they are exactly the kind of key this predicate is meant
+#: to test, and the descent there has to continue. Every other `by_*` map is keyed by a data VALUE:
+#: a reason, a scope, a ref type, a layer name. Checked against the pack builder's full set:
+#: by_codex_event, by_drop_reason, by_entity_type, by_extraction_phase, by_hook_type,
+#: by_memory_layer, by_memory_scope, by_memory_selection_policy, by_profile_memory_kind,
+#: by_profile_promotion_blocker, by_profile_promotion_policy, by_profile_shadowed_reason,
+#: by_ref_type, by_session_continuity, by_source_role.
+_KEYED_BY_NAME_NOT_VALUE = frozenset(("by_dimension",))
+
+
+def _keys_are_values(key: Any) -> bool:
+    """Whether this map's child keys are DATA, so the hidden-key predicate must not judge them."""
+    name = str(key or "")
+    return name.startswith("by_") and name not in _KEYED_BY_NAME_NOT_VALUE
+
+
 def strip_default_debug_lineage_fields(value: Any) -> Any:
     """Remove debug/lineage fields from the default prompt-facing ContextPack.
 
@@ -151,7 +168,7 @@ def strip_default_debug_lineage_fields(value: Any) -> Any:
         for key, item in value.items():
             if _is_default_hidden_debug_lineage_key(key):
                 continue
-            kept[key] = item if str(key).startswith("by_") \
+            kept[key] = item if _keys_are_values(key) \
                 else strip_default_debug_lineage_fields(item)
         return kept
     if isinstance(value, list):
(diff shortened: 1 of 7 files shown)


bjmeetsfo:
Closing this: one half has landed against current main, and the other half rests on a claim that
does not survive asking which copy runs.

## The status map — real, and fixed

`matrixark_mcp_dashboard.latest_async_pipeline_rows` ranked three statuses where
`matrixark_mcp_async_readiness` ranks eight, so every `idle_commit_*` tied at `-1` and a task whose
terminal record carried the earlier stamp read as **still scheduled** on the dashboard while
readiness reported it finished. Reproduced on today's main before fixing:

| rows | dashboard said | readiness said |
| --- | --- | --- |
| scheduled t=200, skipped t=100 | `idle_commit_scheduled` | `idle_commit_skipped` |
| scheduled t=200, committed t=100 | `idle_commit_scheduled` | `idle_commit_committed` |

Landed as #1898, with #1900 correcting its shape: the dashboard now **delegates** to the module that
owns the ranking rather than carrying a second copy of the body. `test_the_two_pipeline_row_rankings_agree`
now compares the two names over every ordered status pair in both timestamp directions, so a fix
cannot reach one and not the other again.

## The pressure shaper — not a serving defect

This PR said the same payload came out of two paths with different keys. Textually true, and there
are in fact **three** copies of `serving_memory_layer_pressure`, not two — the one in
`matrixark_mcp_core` takes an `include_debug` keyword the others do not.

Measured on one input, rather than read:

- `matrixark_mcp_context_pack` and `matrixark_mcp_core` return **identical** output
- `matrixark_mcp_retrieve_metrics` differs — it keeps an empty `by_scope: {}` in `by_dimension` and
  a stale `"by_scope"` in `dropped_dimensions`, which the other two prune

But the retrieve path binds **`matrixark_mcp_core`**, asked at runtime by `__module__`. The copy
that differs lives in `matrixark_mcp_retrieve_metrics`, which
`test_a_module_only_tests_reach_is_not_live` already records as a module **only tests reach**.

So the poorer output is not what a retrieve returns, and the "two paths, different keys" framing in
this PR's description overstates it. Not worth carrying an open PR for, and not worth changing a
test-only module to match.

## Why re-cut rather than rebase

This branch is 200+ commits behind and edits two guard files that have since moved, including the
one recording the dashboard as defining its own copy — which turned out to be the detail that
decides the correct fix shape.

PR #1622, opened 2026-09-13
A quiet shard can reclaim its log again

bjmeetsfo:
## The question

A quiet shard's maintenance loop prints this stage list:

```
round 0:     ["prepare", "reclaim_wal", "expire", "reclaim_page", ...]
rounds 1-11: ["prepare", "expire", "reclaim_memory", "reclaim_page", ...]
```

`reclaim_wal` appears once and never again, and WAL `persistent_bytes` stays flat at 2,243,049.

Flat bytes with no writer is a perfectly good outcome. It is also exactly what a stage that has
silently stopped looks like. The stage list is how an operator reads whether maintenance is
alive, so the two have to be distinguishable. They were not.

## What it actually was

Not the good case. On an 8,000-record fixture with a third of the keys overwritten, **0 of 10,666
records were freed across twelve rounds** and the log sat at 1,344,852 bytes for ever. Two
independent defects, either of which alone is enough to stop reclaim:

### 1. A clean slot blocked the plan (`storage_wal_reclaim_plan`)

An overwrite gives a slot a new generation, so the manifest the round-0 dump wrote no longer
matches its current generation fingerprint and no manifest covers it. The slot falls to the
no-manifest branch, which asks it to name its claim over the log — and its
`first_dirty_wal_sequence` is 0, set to 0 *precisely because* the dump captured everything it
held. The branch read that 0 as "cannot name its claim", filed the slot under
`missing_bucket_generations`, and blocked the whole plan:

```
safe_to_reclaim=false  blocker_reasons=["slot_generation_without_durable_dump"]
covered=0  uncovered=8000
```

Every slot on a quiet shard ends up there, so the plan reported that for all of them, every round,
for ever.

This is the shape #1516 removed from the *manifest* branch — a slot that needs nothing deciding
what the log may drop — surviving one branch over, where it does not merely pin the floor but
refuses outright. A clean slot now counts as covered and contributes no floor, exactly as the
manifest branch already treats a clean slot it does have a manifest for.

Safety: `dirty_object_count == 0` is the same predicate #1516 trusted. A slot goes clean only when
`apply_storage_lifecycle` applies a **durable** dump manifest, which clears `dirty` and
`first_dirty_wal_sequence` together; a reloaded slot is durable by construction. Pages that live
only inside a WAL record are protected independently of this plan — `gc_before_sequence` clamps to
`block_retention_floor_by_shard`, fed from `min_registered_sequence` — so this change cannot drop
a record a resident page still needs.

### 2. The stage asked the wrong question (`run_storage_manager_once`)

The trigger was dump pressure alone — dirty slots, or undumped records — which a shard with no
writer never has. But dump pressure asks "is there undumped work?" while reclaim asks "does the
log hold records nothing needs?". On an idle shard the first is always no and the second is
usually yes, which is exactly when reclaim matters and exactly when it was not run. Nothing else
in a shipped server reclaims on a timer, so an idle shard's log was permanent.

The plan is the only thing that can answer the second question, so the gate now asks it — but only
once the cheap trigger has already declined, so a busy shard pays nothing and an idle one is not
doing anything else with the round. `reclaimable_wal_pressure` is reported as a trigger reason so
the decision record says which question opened the gate.

## Result

Same fixture, both fixes in:

```
round 0: [... "reclaim_wal" ...]   persistent_bytes 1,573,583 -> 1,344,852
round 1: [... "reclaim_wal" ...]   persistent_bytes 1,344,852 ->    33,413
rounds 2-11: stage runs every round, bytes flat at 33,413 (the floor)
```

The log drops by 97.5% on the first idle round, and the stage keeps reporting afterwards — running
and finding nothing, which is the honest signal, rather than vanishing.

Cost: the extra plan build is paid only on rounds with no dump pressure. On the 8,000-slot fixture
twelve rounds went from 25s to 62s wall, against a 30s loop period.

## Guards, and a vacuity trap worth recording

Two tests, both killed by mutation:

| | plan rule reverted | gate reverted |
|---|---|---|
| `a_quiet_shard_with_every_slot_clean_can_reclaim_its_log` | FAILS | passes (correctly — the plan is fine) |
| `the_log_reclaim_stage_keeps_running_while_the_log_is_still_reclaimable` | FAILS (vacuity denominator) | FAILS (stage missing) |

The first version of these guards **survived** the plan mutation. The fixture wrote 2,000 records
and never overwrote any, so every slot kept a matching manifest and reached `safe_to_reclaim`
through the *manifest* branch — passing every assertion whatever the branch under test did.
Measured with the rule reverted: `covered=2000, uncovered=0, safe=true`.

`covered_bucket_count` counts both branches, so it cannot witness which one ran and is not a
denominator. `retained_manifest_ids` is: the manifest branch records the manifest it matched, the
no-manifest branch records nothing. An empty set plus a non-zero covered count proves every slot
took the branch under test, and that assertion now runs before the property does.

## Gate

`cargo test -p temporalstore-rust --lib -- --test-threads=1` — see comment below.


bjmeetsfo:
### Gate

`cargo test -p temporalstore-rust --lib -- --test-threads=1` — **1781 passed, 4 failed**, 2538s.

| failure | verdict |
|---|---|
| `wal::tests::durable_bytes_never_exceed_the_bytes_the_log_holds` | the standing baseline red on main |
| `raft::tests::part2::http_raft_transport_sends_append_entries_as_json_over_tcp` | known full-suite flake (`os error 11`, resource exhaustion) |
| `raft::tests::part2::http_raft_transport_sends_append_vote_and_snapshot_over_tcp` | same |
| `data_node::tests::part3::runtime_async_lifecycle_jobs_report_progress_and_outputs` | 30 s wall deadline blown during a 42-minute suite on a box running seven concurrent test binaries |

The last three pass in isolation, three runs each, in under 4 s — and none of them is on a path
this change touches: the async-lifecycle test submits a `Dump` task, while the reclaim plan is
consulted only on the `Gc` path (guarded by `retain_wal_from_sequence.is_some()`), and the widened
gate lives in `run_storage_manager_once`, which that test never calls.

### Mutation matrix

| | plan rule reverted | gate reverted |
|---|---|---|
| `a_quiet_shard_with_every_slot_clean_can_reclaim_its_log` | FAILS | passes (correctly — the plan is fine) |
| `the_log_reclaim_stage_keeps_running_while_the_log_is_still_reclaimable` | FAILS (vacuity denominator) | FAILS (stage missing) |

Both fixes live: 2 passed.

PR #1743, opened 2026-09-14
The shell flag helper case-folded the value and never trimmed it

bjmeetsfo:
`matrixark_flag_on` lowercases what it is given and matches it against the vocabulary. It never
removed whitespace, so `1 ` matched none of the words in the `case` and fell through to the default.

**That cuts both ways, and both are the operator getting the opposite of what they wrote:**

| helper default | what the operator wrote | what they got |
|---|---|---|
| ON | `0 ` or ` off` | flag stayed **ON** |
| OFF | `1 `, `true ` or `ON ` | flag stayed **OFF** |

A shell export, a systemd `Environment=` line and a heredoc all leave whitespace behind. **Five
copies of the helper carry 22 flag reads between them** across the launchers, and every one was
affected.

This is the shell member of a family: the Python one was the preload flag reading `!= "0"`, and the
Rust one is recorded in `test_a_flag_value_is_trimmed_before_it_is_parsed`, where
`" 64".parse::<usize>()` is an `Err`.

## A/B on the helper extracted from each tree

| value | before (default 1 / 0) | after |
|---|---|---|
| `1` `0` | on/on, off/off | unchanged |
| `"1 "` `" 1"` `"  1  "` `"true "` `"yes "` `"ON "` | **fell to the default** | on/on |
| `"0 "` `" 0"` `"off "` `" off"` | **fell to the default** | off/off |
| `""` `"   "` `garbage` `"on off"` | fell to the default | unchanged |

**11 of 16 values move, every one toward the word meaning what it says, and no value that was
already recognised changes its answer.** Whitespace-only stays unrecognised, because trimming it
leaves nothing.

## Why the existing guard could not see it

It holds the five copies level **against each other** — the right shape for drift. But five copies
agreeing on the wrong answer still agree, and all five were wrong the same way.

It now asserts the answers **outright** for 14 whitespace-bearing values across both default
positions.

| mutation | caught by |
|---|---|
| **all five lose the trim** | **only the new assertion** — they still agree |
| one copy loses it | both the new assertion and agreement |
| the trim deletes too much | both |


bjmeetsfo:
The ratchet caught this, correctly, and the follow-up is pushed.

`test_one_flag_has_one_boolean_vocabulary` recorded the untrimmed divergence and predicted its own end — *"adding trimming to `matrixark_flag_on` fails this file instead of leaving a stale note behind"*. Adding the trimming did exactly that, and the failure message named the follow-up: fix the note, and check whether `MATRIXARK_HOOK_FAIL_OPEN` changed with it.

Checked, across 12 values and both wrappers:

| value | python | claude_hook | codex_rust_hook |
|---|---|---|---|
| ` 0` `0 ` ` false` | off | off | off |
| ` 1` `1 ` ` true` | on | on | on |
| `0` `1` `` `   ` `garbage` | unchanged | unchanged | unchanged |

**0 of 12 disagree** — the divergence is gone in both languages, not moved.

So the record becomes an *agreement* rather than being deleted. The three values stay as witnesses: emptying the tuple would leave that test iterating over nothing, which reads exactly like agreement. The assertion is now "both must answer off", so removing the trim fails it from the other side.

Run against `main`, where the shell does not trim, it fails on all three values; on this branch all 16 pass.

PR #1870, opened 2026-09-18
A general doc promised a hook control only one hook reads

bjmeetsfo:
`docs/CLOUD_API_REFERENCE.md` and `docs/DEPLOY_CLOUD_API.md` promise a hook control that only one of the two hooks honours.

## Measured

| flag | `matrixark_codex_hook` | `matrixark_agent_hook` |
|---|---|---|
| `MATRIXARK_HOOK_AUTO_BATCH_EXTRACT` | reads it | reads it |
| `MATRIXARK_HOOK_FAST_ASYNC_INGEST` | reads it | **does not** |

Across the whole surface, the Codex hook reads **16** `MATRIXARK_HOOK_*` flags and the agent hook reads **2**. Codex-only flags are not a defect in themselves — most of those 16 are documented in the Codex installation manual, where being Codex-scoped is the point.

The defect is a flag in a **general** table that only one hook honours, with nothing saying so. The section is headed *"Ingestion architecture — ingest without a model, batch the extraction"*, sitting between the backend and rate-limit sections of `CLOUD_API_REFERENCE.md`; nothing about it is agent-specific. Its table-mate really is read by both, which is what makes the table misleading rather than obviously scoped — one row right, one row wrong.

What an operator gets for setting it: a Codex hook that stores raw and returns, and a Claude hook that still calls the model inline on every ingest. That is the failure the surrounding prose exists to prevent — *"Without these, the gateway runs synchronous per-message extraction… which collapses throughput to a few QPS"* — still happening, for half the traffic.

## What changes

The documentation, not the behaviour. Both places now say which hook reads the flag.

Making `matrixark_agent_hook` honour it is the other possible fix and is deliberately not taken here: it changes what an existing deployment does the moment it is deployed, for anyone who set the flag with Codex in mind. That is a product decision about the Claude hook; the documentation being wrong today is not.

## A guard, so the next flag added to a general table cannot repeat it

`test_a_general_doc_does_not_promise_a_one_hook_flag` reads every `MATRIXARK_HOOK_*` flag named in the two general documents, works out which hook modules read it, and requires that a flag read by some-but-not-all hooks has the scope stated within its own line or the three following — enough for a markdown table row and for the deploy guide's bash block, where the note is a trailing comment.

Two floors come with it, because a scan that stopped seeing its inputs reports what a clean tree reports: both hook modules must yield a non-empty flag set, and the documents must still name at least two flags.

It asserts nothing about whether the hooks *should* support the same flags — that is the decision above. Only that a general document cannot promise everyone a control one hook ignores.

Removing the scope notes fails it, naming both documents, both line numbers, the flag, and which hook reads it.

## Tests

The eight repo-wide meta-guards that inspect test files, module reachability and flag vocabulary — the ones a grep-chosen slice misses because they inspect *"a test file"* rather than my symbols: **44 tests, green before and after, zero new failures by name.**

## A correction to what prompted this

The note that sent me here listed three flags as Codex-only. Checked today, one of the three is: `MATRIXARK_HOOK_AUTO_BATCH_EXTRACT` is now read by both hooks, and `MATRIXARK_HOOK_FAIL_OPEN` appears only in the Codex installation manual, which is correct scoping rather than an omission.


bjmeetsfo:
Correcting the framing of this: the finding was not new, and I should have found the prior work before writing it up that way.

`test_a_documented_hook_control_reaches_both_hooks`, from #1649, already measured all three flags and recorded the same conclusion — that the general docs over-promise. Its reasoning is better than mine on both of the flags I treated as secondary:

- **`FAST_ASYNC_INGEST`** — I said the agent hook "does not read it". #1649 says why that is not a missing variable read: the Claude hook has **no async-ingest path at all**, under that or any other name. There is nothing for the flag to gate. Closing it means building a feature, not reading a variable.
- **`FAIL_OPEN`** — I reported it as Codex-only and correctly scoped because it appears only in the Codex manual. #1649 adds what actually matters: the Claude hook is **unconditionally fail-open**, so the flag's documented effect is already its behaviour and `=0` cannot make it fail closed.

So what this PR adds is the documentation change that #1649 deliberately did not make — it recorded the over-promise rather than editing the text — plus a sweep beside its named list.

The two guards are complementary rather than duplicated, and it is worth writing down which is which:

| guard | subject | what goes stale |
|---|---|---|
| `test_a_documented_hook_control_reaches_both_hooks` (#1649) | the **wiring** of three named flags, asserted in both directions | the list of three |
| `test_a_general_doc_does_not_promise_a_one_hook_flag` (this PR) | the **documentation text** for whatever flags the general docs name | nothing — it reads the docs |

A named list catches a change to the things it names; a sweep catches the next thing added. Neither is sufficient alone, which is the same pairing the shard-size guard ended up with in #1848 and #1849.

